PR #80, opened 2015-05-08
Added Node::GetChildrenWithName

mattbenic:


Source/Atomic/Scene/Node.cpp
```diff
@@ -958,6 +958,32 @@ void Node::GetChildren(PODVector<Node*>& dest, bool recursive) const
         GetChildrenRecursive(dest);
 }
 
+void Node::GetChildrenWithName(PODVector<Node*>& dest, const String& name, bool recursive) const
+{
+	GetChildrenWithName(dest, StringHash(name), recursive);
+}
+
+void Node::GetChildrenWithName(PODVector<Node*>& dest, const char* name, bool recursive) const
```
JoshEngebretson: Just a note, I believe this overload exists to resolve ambiguity between the String& and StringHash overloads when passing a quoted string, as it chooses String, I don't think we need the StringHash nameHash overload at all as this is searching by name which shouldn't be hashed (like a type's name)

Source/Atomic/Scene/Node.cpp
```diff
@@ -958,6 +958,32 @@ void Node::GetChildren(PODVector<Node*>& dest, bool recursive) const
...
+
+void Node::GetChildrenWithName(PODVector<Node*>& dest, StringHash nameHash, bool recursive) const
+{
+	dest.Clear();
+
+	if (!recursive)
+	{
+		for (Vector<SharedPtr<Node> >::ConstIterator i = children_.Begin(); i != children_.End(); ++i)
+		{
+			if ((*i)->GetNameHash() == nameHash)
```
JoshEngebretson: Fast string compare, so the StringHash overload is indeed needed.

JoshEngebretson:
Thanks!  This is now on master, https://github.com/AtomicGameEngine/AtomicGameEngine/pull/85

There are actually quite few cases of methods taking arrays like this so questionable adding to JSBind, will keep thinking about passing arrays.

PR #82, opened 2015-05-10
Fixed binding of android and web platform on windows

marynate:
- JSBind should use OpenGL Graphics module when building Android and Web platform on windows


JoshEngebretson:
Great, thanks! :+1:

PR #86, opened 2015-05-11
Fix android build on windows

marynate:
- Fixed compile errors when building android player on windows
- Build is successful, but haven't been tested on android device or emulator


JoshEngebretson:
Hello, we're you able to build Android on Windows using master?  In theory this should work as applied similar fixes while getting CI working.  However, I have not tried building Android using NDK on Windows.

Thanks :)

marynate:
The master branch build just fine now on Windows. Great job!

PR #109, opened 2015-06-11
Node.GetChildrenWithName, Node.getJSComponent, error reason for failed plugin loading, Bug fix for collision type check.

mattbenic:


Source/Atomic/Core/ProcessUtils.cpp
```diff
@@ -175,6 +175,10 @@ void PrintUnicode(const String& str, bool error)
         WString strW(str);
         DWORD charsWritten;
         WriteConsoleW(stream, strW.CString(), strW.Length(), &charsWritten, 0);
+        if (IsDebuggerPresent())
+        {
+            OutputDebugString(str.CString());
+        }
```
JoshEngebretson: Great!

Source/Atomic/Scene/Node.h
```diff
@@ -376,7 +376,7 @@ class ATOMIC_API Node : public Animatable
     const Vector<SharedPtr<Node> >& GetChildren() const { return children_; }
     /// Return child scene nodes, optionally recursive.
     void GetChildren(PODVector<Node*>& dest, bool recursive = false) const;
-    /// Return child scene nodes with a specific component.
+	/// Return child scene nodes with a specific component.
     void GetChildrenWithComponent(PODVector<Node*>& dest, StringHash type, bool recursive = false) const;
```
JoshEngebretson: It looks like a tab may be lurking....

Source/Atomic/Scene/Node.h
```diff
@@ -495,7 +495,7 @@ class ATOMIC_API Node : public Animatable
     void RemoveChild(Vector<SharedPtr<Node> >::Iterator i);
     /// Return child nodes recursively.
     void GetChildrenRecursive(PODVector<Node*>& dest) const;
-    /// Return child nodes with a specific component recursively.
+	/// Return child nodes with a specific component recursively.
     void GetChildrenWithComponentRecursive(PODVector<Node*>& dest, StringHash type) const;
```
JoshEngebretson: Here too...

Source/AtomicJS/Javascript/JSComponent.cpp
```diff
@@ -422,7 +422,7 @@ void JSComponent::HandleScriptEvent(StringHash eventType, VariantMap& eventData)
         duk_context* ctx = vm_->GetJSContext();
         JS_HEAP_PTR function = scriptEventFunctions_[eventType];
 
-        if (eventType == E_PHYSICSBEGINCONTACT2D || E_PHYSICSENDCONTACT2D)
+        if ((eventType == E_PHYSICSBEGINCONTACT2D) || (eventType == E_PHYSICSENDCONTACT2D))
         {
```
JoshEngebretson: Nice catch

Source/AtomicJS/Javascript/JSPlugin.cpp
```diff
@@ -86,7 +88,9 @@ namespace Atomic
         if (duk_pcall(ctx, 1) != DUK_EXEC_SUCCESS)
         {
             success = false;
-            LOGERRORF("Native Plugin: error calling atomic_plugin_init %s", pluginLibrary.CString());
+            LOGERRORF("Native Plugin: error calling atomic_plugin_init %s with error %s", 
+                pluginLibrary.CString(), 
+                duk_safe_to_string(ctx, -1));
```
JoshEngebretson: cool, eventually want these to probably be an error event, or maybe throw

Source/AtomicJS/Javascript/JSScene.cpp
```diff
@@ -24,6 +24,27 @@ static int Node_CreateJSComponent(duk_context* ctx)
...
+		if (StringHash(jsComponents[i]->GetClassName()) == classNameHash)
+		{
+			js_push_class_object_instance(ctx, jsComponents[i]);
+			break;
+		}
+	}
+
+	return 1;
+}
+
```
JoshEngebretson: Ok, so a couple things here:

1) There may be more than one JS component on a node, this only gets the 1st

2)  This function should be unnecessary as the node object is populated with a property of the component class:

https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/AtomicJS/Javascript/JSVM.cpp#L177

.. and there is a lookup to all components of that class on  the node in the components lookup on the node:

https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/AtomicJS/Javascript/JSVM.cpp#L180

This should all be documented, sorry.  

I also think that the JSComponents will be able to be improved with some of the new scripting features, like being able in use inheritance on a prototype chain from script.

Source/Atomic/Scene/Node.h
mattbenic: Must have been prior to our switching to spaces.Do I need to correct my side and re-create the PR?

Source/AtomicJS/Javascript/JSScene.cpp
mattbenic: 1) I do realize it would only return the first, I'd think a get JSComponents() that returns all of them would be helpful there (particularly since the named fields will only cover the most recently added component of that type).

2) Yeah you mentioned when we first added this that they could be reached that way, still seemed like something worth keeping though.

JoshEngebretson: If I have a component calls MyComponent, the last one added on  the node would be available as:

self.node.myComponent 

This would also work if there was only one.

if I need access to more than one of the JSComponents on the node:

self.node.components["MyComponent"]; // list of MyComponents on the node

mattbenic: Ah ok, yes that'll cover that case as well. So just remove this entirely?

JoshEngebretson: It would seem this should instead be a general GetComponents which takes a typehash, instead of specifically comparing the JSComponent typehash in  the function.  This would also work for JSComponents.

I filed an issue:
https://github.com/AtomicGameEngine/AtomicGameEngine/issues/110

Source/Atomic/Scene/Node.h
JoshEngebretson: Yeah, as actually doing a code review here should probably flag it :)   It is pretty easy to spot these when submitting a PR.  

You don't have to recreate the PR, if you commit to the PR branch the changes will show up in the PR

mattbenic: Ok, yeah I changed those directly on GH (oddly the tab shows as spaces in
VS, and deleting the whitespace and re-adding it didn't show in git either)

Regards,
Matt Benic
mattbenic@gmail.com | +27 (72) 1826161 | za.linkedin.com/in/mattbenic

Source/AtomicJS/Javascript/JSScene.cpp
mattbenic: Ok, if that's something you'd like me to sort out I'll get to it on Monday.

JoshEngebretson: I might get to it before that, pretty nice to be able to get the components on a node :rabbit: 

Otherwise, the PR looks great, thanks!

JoshEngebretson:
For fun, let's run this through the CI machinery :)

JoshEngebretson:
ok to test

JoshEngebretson:
Woohoo, builds on Windows, Mac, Linux, iOS, Android, and WebGL :)

I'll add you to the whitelist, so any PR will automatically go to CI

JoshEngebretson:
add to whitelist

mattbenic:
Cool :)

JoshEngebretson:
Ok, Node.getComponents is now available to script as of https://github.com/AtomicGameEngine/AtomicGameEngine/commit/1014851cb60e57a1d756f2001aaae76772f3a676

// Get all components
var components = node.getComponents();

// Get all components recursive
var components = node.getComponents("", true);

// Get only JSComponents on node
var components = node.getComponents("JSComponent");

// Get all static models, recursively
var components = node.getComponents("StaticModel", true);

Source/AtomicJS/Javascript/JSAPI.cpp
```diff
@@ -69,13 +69,17 @@ void js_class_declare_internal(JSVM* vm, void* uniqueClassID, const char* packag
         duk_put_prop(ctx, -3);
         duk_pop(ctx);
     }
-    else
+    
+    if (!duk_get_global_string(ctx, package))
     {
-        // RefCounted only supported in Atomic package
-        assert(String("Atomic") == package );
+        // Failure leaves undefined on the stack
```
JoshEngebretson: Does this code path ever get hit?  The package object should exist before declaring any classes in it.  The assert that was in here when uniqueID == NULL is due to how the class -> package lookup is done when pushing a variant (to only push bound classes).  

I made a ticket to improve this:  https://github.com/AtomicGameEngine/AtomicGameEngine/issues/124

Were you hitting the assert binding something that was derived directly from RefCounted?

Source/AtomicJS/Javascript/JSPlugin.cpp
```diff
@@ -26,7 +26,9 @@ namespace Atomic
 
         if (handle == NULL)
         {
-            errorMsg = ToString("Native Plugin: Unable to load %s", pluginLibrary.CString());
+            errorMsg = ToString("Native Plugin: Unable to load %s with error %s", 
+                pluginLibrary.CString(),
+                SDL_GetError());
             return false;
```
JoshEngebretson: Nice add

Source/ToolCore/JSBind/JSBTypeScript.cpp
```diff
@@ -32,10 +32,18 @@ String JSBTypeScript::GetScriptType(JSBFunctionType* ftype)
...
-        scriptType = ftype->type_->asEnumType()->enum_->GetName();
+        scriptType = enumType->GetName();
+
+        if (enumType->GetPackage()->GetName() != package_->GetName())
+        {
+
+            scriptType = enumType->GetPackage()->GetName() + "." + enumType->GetName();
+
+        }
+    }
```
JoshEngebretson: Properly scoped, enums, nice!

Source/ToolCore/JSBind/JSBTypeScript.cpp
```diff
@@ -57,6 +65,9 @@ String JSBTypeScript::GetScriptType(JSBFunctionType* ftype)
         scriptType = "string[]";
     }
 
+    if (ftype->type_->asHeapPtrType())
+        scriptType = "any";
+
```
JoshEngebretson: Do you recall when this happens?  It again with something derived from RefCounted?

Source/AtomicJS/Javascript/JSAPI.cpp
mattbenic: It does. I'm using JSBind on some non-Atomic namespaced code, but the class does extend from RefCounted. When this is loaded, the Atomic namespace object exists, but our namespace object does not.

Source/ToolCore/JSBind/JSBTypeScript.cpp
mattbenic: I was with one of the callback registering functions on components I think. Since the callback parameter wasn't a recognized type it resolved to a Number, which TypeScript wouldn't accept. First prize would of course be to have the callback types be properly resolved to proper TS callback syntax, but that was a much bigger add.

Source/AtomicJS/Javascript/JSAPI.cpp
JoshEngebretson: Ok, I filed an issue to define when the the package object is created.  It has been in "user code", though I will be moving it the package init as generated by JSBind, there are some details in the issue: https://github.com/AtomicGameEngine/AtomicGameEngine/issues/133

mattbenic: Thanks, following it

Source/ToolCore/JSBind/JSBTypeScript.cpp
mattbenic: On second thought, I would suggest that the default for non-specific types should probably be any, to at least allow user TS to compile without changes to engine code. I realize it makes picking up new types that should have specific outputs here a bit tougher, but it's the better approach for user friendliness IMO

JoshEngebretson: JSBind should be filtering methods it doesn't have proper typing on, an example or two would help here

I would say "any" is definitely better then a superfluous "number" typing!

mattbenic: Ok, when I get a chance I'll run with that fix removed, it'll pop the example up immediately :)

JoshEngebretson:
Ok, as of https://github.com/AtomicGameEngine/AtomicGameEngine/pull/136

The package objects are being created by JSBind and will not have to be manually created or lazily created when registering a class

... and as of https://github.com/AtomicGameEngine/AtomicGameEngine/pull/137

RefCounted classes now have efficient RTTI class identifiers and should be able to be used outside the "Atomic" package.  This requires a REFCOUNTED macro in the class declaration much like OBJECT, though you don't need to include both for Objects, for example:

https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/Atomic/Core/Context.h#L49

Let me know if this works for you!

:star2: 

- Josh

JoshEngebretson:
Hey Matt, were you going to open a new PR with some of these updates?  

If not, I can cherry pick the relevant changes, though none of the core stuff trips the JSHeapPtrType() => "any" conversion here. I'd like to get a repro case of that, thanks!

mattbenic:
Hi Josh,
We had a holiday here yesterday, so I'm getting back to the artist build
today (hopefully more successfully now that I'll be raking it). If that
works out then yes I'll PR for you today. I'm going to try that repro first
(doing that right now) and will pass on the details. I can't respond on
gitter today unfortunately, it's one of a handful of sites our ganky
hardware firewall at the office mangles due to a firmware issue.

Regards,
Matt Benic
mattbenic@gmail.com | +27 (72) 1826161 | za.linkedin.com/in/mattbenic

mattbenic:
I tracked down the case we had, but it's no longer valid since the event
system has changed. We were using the following in a component:

self.postRenderUpdate = function () {
world.drawDebugGeometry(debugRenderer, true);
}

function update(_timeStep: number) {
if (drawDebug)
self.listenToEvent(null, "PostRenderUpdate", self.postRenderUpdate);
}

That listenToEvent would fail typescript compilation since at the time
listenToEvent's JS_HEAP_PTR parameter was generated as being of type number
instead of type any when the TypeScript was output. This script isn't
really used anymore our side, so it's likely not an issue anymore (and why
we hadn't noticed the event change yet).

Regards,
Matt Benic
mattbenic@gmail.com | +27 (72) 1826161 | za.linkedin.com/in/mattbenic

JoshEngebretson:
Ok, sounds good.  

The script event model now mirrors the native one, so all engine events (and event data) are available to script.  Let me know if there are any gaps in functionality.

- J

mattbenic:
Hi Josh,
Sorry I didn't get the build setup completed yesterday, so no PR just yet.

Regards,
Matt Benic
mattbenic@gmail.com | +27 (72) 1826161 | za.linkedin.com/in/mattbenic

JoshEngebretson:
There's no rush, I just don't want to lose this stuff in the churn :)

- Josh

PR #146, opened 2015-08-05
Editor will give a real message error through using a source maps

rsredsq:
There's a still a bug that duktape couldn't execute a file with no EOF, unlikely TS compiler doesn't append an EOF after generating source maps https://github.com/Microsoft/TypeScript/issues/3287
Editor now compiles into /out/ directory, so changed .gitignore to ignore only that folder


Source/AtomicJS/Javascript/JSVM.cpp
```diff
@@ -247,6 +247,42 @@ void JSVM::SendJSErrorEvent(const String& filename)
 
 }
 
+int JSVM::GetRealLineNumber(const String& fileName, const int lineNumber) {
+    int realLineNumber = -1;
+    String map;
+    String path = fileName;
+    if (!path.EndsWith(".js.map"))
+        path += ".js.map";
+    SharedPtr<File> mapFile(GetSubsystem<ResourceCache>()->GetFile(path));
```
JoshEngebretson: Should the be some error checking that the file is indeed open?

Source/AtomicEditor/Application/AEEditorApp.cpp
```diff
@@ -158,9 +158,10 @@ void AEEditorApp::HandleJSError(StringHash eventType, VariantMap& eventData)
     String errFilename = eventData[P_ERRORFILENAME].GetString();
     //String errStack = eventData[P_ERRORSTACK].GetString();
     int errLineNumber = eventData[P_ERRORLINENUMBER].GetInt();
+    int errRealLineNumber = vm_->GetRealLineNumber("AtomicEditor/out/" + errFilename, errLineNumber);
```
JoshEngebretson: Hm, check that the error happened in a source mapped file, somehow?  How does this handle normal JS errors in non-sourcemapped files?

Resources/EditorData/AtomicEditor/Script/jsutils.js
```diff
@@ -1,5 +1,6 @@
 __atomic_acorn = require('./acorn');
 __atomic_beautify = require('./beautify');
+__atomic_sourcemap = require('./source-map');
```
JoshEngebretson: Hah, at least it is consistent in its horribleness :)

Resources/EditorData/AtomicEditor/Script/jsutils.js
```diff
@@ -65,3 +65,10 @@ exports.jsBeautify = function (source) {
 	return __atomic_beautify.js_beautify(source);
 
 }
+
+exports.getRealLineNumber = function (map, line) {
+    var jsonMap = JSON.parse(map);
```
JoshEngebretson: This should speed up with newer Duktape in a bit

Script/AtomicEditor/tsconfig.json
```diff
@@ -5,7 +5,9 @@
         "declaration": false,
         "noImplicitAny": false,
         "removeComments": true,
-        "noLib": false
+        "noLib": false,
+        "outDir": "out/",
+        "sourceMap": true
```
JoshEngebretson: Nice add

Source/AtomicJS/Javascript/JSVM.cpp
rsredsq: https://github.com/rsredsq/AtomicGameEngine/commit/906d84ad1c56991c7b18a3d7795eef6f12045c13 should fix that

Source/AtomicEditor/Application/AEEditorApp.cpp
rsredsq: https://github.com/rsredsq/AtomicGameEngine/commit/906d84ad1c56991c7b18a3d7795eef6f12045c13 should also fix that too

JoshEngebretson:
I am almost hesitant to mention this, but, here is where the source code is loaded on a require call:

https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/AtomicJS/Javascript/JSRequire.cpp#L135

Would it make sense to append to the EOF?

rsredsq:
@JoshEngebretson 
After a few tests with that bug, I found that it happens because of https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/Atomic/IO/File.cpp#L542
It adds a '\0' at the end of a string, and duktape gives an error.
Check this please: https://github.com/rsredsq/AtomicGameEngine/commit/89c8584cd18bf8c09c59c42e57b38830ebbccebd
It fixes that, if I change '\0' to just ' ' and add a new line

JoshEngebretson:
Excellent sleuthing!  I am really glad you looked into it and found the culprit!  

Is adding the space strictly necessary or can you just replace the terminating '\0' with '\n'?  I like that it is localized to the JS source loader, there are other places ReadText shouldn't do this, though should return a (unmodified) terminated C string.

JoshEngebretson:
ok to test

PR #190, opened 2015-08-12
Require out/main.js instead of raw evaling it

rsredsq:


Script/AtomicEditor/main.ts
```diff
@@ -3,6 +3,16 @@
...
+    static Editor:Editor;
+
+    static main() {
+        Main.Editor = new Editor();
+    }
+}
+
+Main.main();
+
+export = Main;
```
JoshEngebretson: This export makes it seem like we're requiring this, however we are evaling it.  Let's remove the export = Main, eventually I would like to get main.ts required as this is still causing a Main in the global namespace as we're requiring

JoshEngebretson:
Closing in favor of https://github.com/AtomicGameEngine/AtomicGameEngine/pull/192

PR #212, opened 2015-08-14
Support constructor arguments to JSComponent derived components, whil…

JoshEngebretson:
…e still supporting flat JS components


JoshEngebretson:
@shaddockh This PR may affect you as there is a new way to export "flat JS" components, please see for examples: https://github.com/AtomicGameEngine/AtomicExamples/commit/a85dd69120dfca6617d587790350317035720a5d

Instead of using module.exports to export the component function which takes self, you export the function to the component property of the exports object

Issue #226, opened 2015-08-15
Playerspawn from Tiled imports as Object


LaraEngebretson:
All Entities, Objects and Physics are imported with the name "Object".
<img width="193" alt="tiled_object" src="https://cloud.githubusercontent.com/assets/722928/9290326/c017c924-4342-11e5-8688-ce0fcea78f70.png">

Issue originally reported here: #159

PR #255, opened 2015-08-22
Added custom name for objects in tmx map

rsredsq:


rsredsq:
then #226 could be closed

PR #259, opened 2015-08-23
TSH-EVENTLOOP

shaddockh:
Implemented AtomicEventLoop based upon the DukTape Ecmascript event loop example.  Modified this by removing the reliance on sockets and instead subscribing to the Atomic.engine update and postUpdate events.

Also implemented a version of setImmediate/clearImmediate

### DIFF
--- a/Resources/CoreData/AtomicModules/AtomicEventLoop.js
+++ b/Resources/CoreData/AtomicModules/AtomicEventLoop.js
@@ -485,8 +485,8 @@ Atomic.engine.subscribeToEvent('Update', throttleProcessTimers(100));
 
 // Hook into the postUpdate event of the engine and process all the setImmediate calls.  These should
 // really process once the current update loop is completed.
-Atomic.engine.subscribeToEvent('PostUpdate', function () {
-    EventLoop.processImmediates;
+Atomic.engine.subscribeToEvent('PostUpdate', function (eventData) {
+    EventLoop.processImmediates();
 });
 
 // Load up the global methods .  This module doesn't export anything, it just sets up the global methods.
--- a/Resources/CoreData/AtomicModules/AtomicEventLoop.js
+++ b/Resources/CoreData/AtomicModules/AtomicEventLoop.js
@@ -35,8 +35,6 @@ EventLoop = {
     expiring: null, // set to timer being expired (needs special handling in clearTimeout/clearInterval/clearImmediate)
     nextTimerId: 1,
     minimumDelay: 1,
-    minimumWait: 1,
-    maximumWait: 60000,
     maxExpirys: 10,
 
     // misc
@@ -209,6 +207,18 @@ EventLoop.processTimers = function () {
             // 'removed' above.
             this.insertTimer(t);
         }
+    }
+    // see if there are any events left over for this cycle
+    n = timers.length;
+    if (timers.length <= 0) {
+        return false;
+    }
+    t = timers[n - 1];
+    if (now > t.target) {
+        // we have more timers that need to be handled.  Let the driver know that we should pick them up in the next cycle
+        return true;
+    } else {
+        return false;
     }
 }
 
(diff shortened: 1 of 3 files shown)


JoshEngebretson:
add to whitelist

JoshEngebretson:
This looks great, nice job hooking it up to the event model.  Once in use, might be some issues brought up, but good initial cut... and awesome that there is also an example :+1: 

One thought, what do you think about running the update and postupdate hooks at 10hz, I am concerned that hitting the event loop processing every frame could have detrimental performance implications?

shaddockh:
The timers can really only be scheduled to the millisecond level, so anything that runs faster than that would just be a waste of cycles.  Is there a way to throttle the update and postupdate so that they only fire at most once a millisecond?  Or are you suggesting that they only fire at most once per 10 milliseconds?  If that's the case, I think that might also be acceptable.

JoshEngebretson:
Yeah, every 10ms at most might be good can be tweaked.  This would also reduce the chances someone will try and do high performance timing with the event system, for which it really isn't intended.

the Update and PostUpdate events both have a timeStep parameter passed, which could be used as a deltaTime for controlling update frequency.

JoshEngebretson:
Actually, 10ms is still 100fps, which seems like overkill as there are better timing options.  10hz (100ms) might be a good rate?

shaddockh:
That sounds good.  Running the example at setInterval(fn, 100) does introduce a stutter in the animation, but I agree, using the event system for animation is the wrong way to do it.  So would the throttle be something that is implemented in-engine? Or would I need to throttle the event loop javascript?

Resources/CoreData/AtomicModules/AtomicEventLoop.js
```diff
@@ -0,0 +1,475 @@
...
+/**
+ * custom call to exit out of the current event loop.
+ */
+function requestEventLoopExit() {
+    EventLoop.requestExit();
+}
+
+// Hook into the update event of the engine and process all the timers
+Atomic.engine.subscribeToEvent('Update', function updateTimer() {
+    EventLoop.processTimers();
```
JoshEngebretson: updateTimer will be called with a timeStep, so could be updateTimer(timeStep)

The timeStep could be used as a delta to control whether to call EventLoop.processTimers or early return

Resources/CoreData/AtomicModules/AtomicEventLoop.js
```diff
@@ -0,0 +1,475 @@
...
+});
+
+// Hook into the postUpdate event of the engine and process all the setImmediate calls
+Atomic.engine.subscribeToEvent('PostUpdate', function updateImmediates() {
+    EventLoop.processImmediates();
```
JoshEngebretson: see note about updateTimer, the PostUpdate event also provides a timeStep argument

JoshEngebretson:
Let's do it in AtomicEventLoop for now, if problematic maybe can introduce something native side.

Thanks!

shaddockh:
I was thinking about this today and this is what I have now.
- the setTimer/setInterval checks are throttled to check every 100ms
- if there are less than 10 timers that expire in any given update, they all execute
- if there are more than 10 timers that expire, then 10 will be executed, and the remaining will be executed in the next update and won't wait 100ms before updating.  This gives the engine time to process background events and not hog the CPU for a prolonged series of update handlers.
- for setImmediate, the entire point behind them is that they should execute immediately after the current update cycle, so these are not throttled.

Resources/CoreData/AtomicModules/AtomicEventLoop.js
```diff
@@ -0,0 +1,502 @@
...
+}
+
+// Hook into the update event of the engine and process all the timers.  These
+// will be throttled to only run every 100ms
+Atomic.engine.subscribeToEvent('Update', throttleProcessTimers(100));
+
+// Hook into the postUpdate event of the engine and process all the setImmediate calls.  These should
+// really process once the current update loop is completed.
+Atomic.engine.subscribeToEvent('PostUpdate', function () {
+    EventLoop.processImmediates;
```
JoshEngebretson: Is this intended to be a call?

JoshEngebretson:
Ok, I like the updates :)  One question and then can land this and the example.

Resources/CoreData/AtomicModules/AtomicEventLoop.js
shaddockh: If you are asking about line 484, then yes, it's correct.  What's happening is that the call to throttleProcessTimers returns a function that's wrapped inside a closure that accumulates the delta time between calls and once it gets above 100 (or whichever amount is specified here), then it actually calls EventLoop.processTimes, otherwise it just pops back out and waits for the next update.

shaddockh: Actually, hold off on accepting this PR.  I did a quick search to see how DukTape handles closures and apparently they are less performant than just creating a prototype object.  I'd like to modify this code to track the state inside a small prototype class instead.  

from a duktape issue:
>Prototype functions are more efficient: closures involve tracking scope objects internally to deal with potentially unknown variable lookups in the inner function (which may, for instance, eval an argument!). Duktape does some primitive detection to avoid tracking scopes for inner functions when possible - e.g. if a function makes direct variable accesses only to its own identifiers, makes no eval() calls, and has no inner functions, its outer scope doesn't need to be tracked. This is not very sophisticated, but covers some basic cases.

JoshEngebretson: I am referring to 489, sorry for not being clear that looks like a NOP to me.

JoshEngebretson: I also think we can stick with the closure as it is clean and very common in JS.  This looks great for handling some events, if we're looking at a lot of them might want to move some of it native (though that isn't a panacea).

shaddockh: good catch.  Ok, fixed and committed.

JoshEngebretson:
Ok, landing!  Thanks and welcome to the authors list :)  

One thing that is going to be needed is a unit test framework for core stuff.  Probably once the CLI is back online.

PR #278, opened 2015-08-29
Added audio, ui variables to haxe api.

rsredsq:
Added setMaterial function to jsbind.

### DIFF
--- a/Source/AtomicJS/Javascript/JSAtomic3D.cpp
+++ b/Source/AtomicJS/Javascript/JSAtomic3D.cpp
@@ -1,5 +1,6 @@
 
 #include <Atomic/Atomic3D/StaticModel.h>
+#include <Atomic/Atomic3D/CustomGeometry.h>
 
 #include "JSAtomic3D.h"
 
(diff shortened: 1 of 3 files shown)


Source/AtomicJS/Javascript/JSAtomic3D.cpp
```diff
@@ -24,6 +25,23 @@ static int StaticModel_SetMaterialIndex(duk_context* ctx) {
...
+
+    // event receiver
+    CustomGeometry* geometry = js_to_class_instance<CustomGeometry>(ctx, -1, 0);
+
+
+    geometry->SetMaterial(index, material);
+
+
+    return 0;
+}
```
JoshEngebretson: Instead of a custom JS API implementation, is it possible to use overloads capabilities of binding tool to select which to use:

  /// Set material on all geometries.
    void SetMaterial(Material* material);
    /// Set material on one geometry. Return true if successful.
    bool SetMaterial(unsigned index, Material* material);

As with: https://github.com/AtomicGameEngine/AtomicGameEngine/blob/30b5ea42ecd21443d39792a273770c6ff8560de9/Script/Packages/Atomic/Atomic3D.json#L13

If not, any custom API should be also added to  the typescript decl:

https://github.com/AtomicGameEngine/AtomicGameEngine/blob/30b5ea42ecd21443d39792a273770c6ff8560de9/Script/Packages/Atomic/Scene.json#L28

JoshEngebretson:
Great job, landing!

PR #308, opened 2015-09-18
Billboards exposed to the JSAPI.

rsredsq:


Source/Atomic/Atomic3D/BillboardSet.cpp
```diff
@@ -310,15 +322,15 @@ void BillboardSet::SetBillboardsAttr(const VariantVector& value)
     unsigned numBillboards = index < value.Size() ? value[index++].GetUInt() : 0;
     SetNumBillboards(numBillboards);
 
-    for (PODVector<Billboard>::Iterator i = billboards_.Begin(); i != billboards_.End() && index < value.Size(); ++i)
+    for (Vector<SharedPtr<Billboard>>::Iterator i = billboards_.Begin(); i != billboards_.End() && index < value.Size(); ++i)
     {
-        i->position_ = value[index++].GetVector3();
-        i->size_ = value[index++].GetVector2();
+        i->Get()->position_ = value[index++].GetVector3();
```
JoshEngebretson: Store to local ptr instead of i->Get() repeatedly?

JoshEngebretson:
Very nice!  It is awesome to have these available to JSAPI  :1::

It would be good to cache and operate on a local billboard pointer instead of having all  the iterator Get calls during iteration.  There's a number places, do you mind updating with this change?

Issue #237, opened 2015-08-15
Check import scale on .fbx


LaraEngebretson:
If set to .01, it gets changed to .00999999...

PR #309, opened 2015-09-19
Fix for #237, using scale as double number to avoid float "noise"

rsredsq:


JoshEngebretson:
Does this have to do with the text input precision having loads of digits, or were you seeing visual artifacts?

JoshEngebretson:
Ah, yeah, you referenced this issue there, very nice :)

LANDING!

PR #315, opened 2015-09-26
Adding WebSocket support for native, and fixing HttpRequest

Type1J:
This only for native. Emscripten support is not yet available. I noticed that the HttpRequest example didn't work, so I made it work as well.


JoshEngebretson:
Wow, big pull request :)

I'll send this to CI, but need to digest it a little.  Did you see that there is a HTTPRequest class?

https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/Atomic/Network/HttpRequest.h

Thanks!

JoshEngebretson:
ok to test

Type1J:
There were 2 third party library additions, which I added the license files. They should be compatible with your license. That's the bulk of the pull request.

The additions were ASIO (standalone, not the Boost version), and WebSocketPP.

The parts that I wrote are the Atomic/Network files WebSocket.h and WebSocket.cpp. The other changes were just to get the JSBind to work, and to fix HttpRequest to work again.

Yes, I did see HttpRequest. I based my WebSocket implementation on it. I did have some questions about things that my be inherited from Urho3D:

1) I used the Atomic based mutex class regulate access to an Atomic based list class. I know that in STL from C++11 on, the collection classes use atomic (nomenclature clash here, note the lower case ;-)) access, so you don't have to use a mutex for operations such as checking the size and push, pop. I didn't dive into the list code, so I used the mutex to be safe for now, but is it necessary to use it at all?

2) The HttpRequest class isn't an event dispatcher. I expected it to be one. Because of that difference, the interface to WebSocket class was modeled after it where payloads are checked in the handler for the network "update" event. I did that because I thought that might be what's expected. Is this what's expected? I could rewrite it to call an event handler when a new payload is received. It might be good to have both interfaces, and let the user choose which works better for them. What do you think?

Anyway, I'm really excited to contribute to this great project. I hope to make more contributions in the future.

JoshEngebretson:
Thanks :)

1)  Yes, if the list is being access from multiple threads

2)  The best way to handle events is using SendEvent, SubscribeToEvent, as this works very well between native and script code.  There isn't much of an expectation here.  WebSockets could send an event instead of polling.  It might not be a bad idea to change HttpRequest to send an event as well, if it would simplify things.

I am wondering if there shouldn't be a separate subsystem for web requests/sockets, instead of hanging stuff off Network?  The Web subsystem could be in the Network module.  Looking over the Network class, I do think this would be preferable. 

It looks like the build server is having some problems with the PR: http://atomicbuilds.mooo.com:9765/job/AtomicPR_Validate/205/

Type1J:
I've looked at the logs, and I know what the problem is, but I can't get to it today.

Also, since Network was existing, I may just leave it untouched, and make a new module, like you are suggesting, which would nullify this pull request. I don't know what I should call it. "Http" maybe? WebSocket requests do start as http, and are upgraded (101 Switching Protocols) to WebSockets, so it's still fitting. Also, since the URLs for WebSockets start with ws or wss (SSL) instead of http, I could use 1 call to start either type of request, which may be cleaner.

JoshEngebretson:
Cool.  

I think it can go in the Network module, how about calling the new subsystem "Web".  A subsystem is just an Object which is registered with the context as a subsystem (ie. singleton).  If the Web subsystem can handle sockets and http requests, that would make sense.  I think the existing http request functionality on Network is kind of ill conceived, as is having to poll it :)

Type1J:
That sounds fine. I send a new pull request when I'm done.

PR #316, opened 2015-09-26
UIButton now can emulate keyboard buttons.

rsredsq:


Source/Atomic/Input/Input.h
```diff
@@ -290,6 +292,12 @@ class ATOMIC_API Input : public Object
     /// Return whether application window is minimized.
     bool IsMinimized() const;
 
+    /// Binds UIButton element to the given button
+    void BindButton(UIButton* touchButton, int button);
+
+    void FakeButtonDown(int button);
+    void FakeButtonUp(int button);
```
JoshEngebretson: How about "Simulate" instead of "Fake", so SimulateButtonDown and SimulateButtonUp?

JoshEngebretson:
This looks very clean, great work.  I have one comment above... also, did you happen to test out the editor in general with these changes?

rsredsq:
@JoshEngebretson yeah, no worries, Josh, it doesn't influence any ui logic :)

PR #331, opened 2015-10-01
Exposed TouchState in the JSAPI

rsredsq:


Source/Atomic/Input/Input.cpp
```diff
@@ -1604,29 +1606,32 @@ void Input::HandleSDLEvent(void* sdlEvent)
         if (evt.tfinger.touchId != SDL_TOUCH_MOUSEID)
         {
             int touchID = GetTouchIndexFromID(evt.tfinger.fingerId & 0x7ffffff);
-            TouchState& state = touches_[touchID];
-            state.touchID_ = touchID;
+            SharedPtr<TouchState> state(new TouchState());
```
JoshEngebretson: Should reuse state from touches_ if it exists instead of creating new state every time?  How does script holding references to stale TouchStates get affected by new states?

PR #332, opened 2015-10-01
TouchState marshaling through a VariantMap.

rsredsq:


Source/Atomic/Input/Input.h
```diff
@@ -66,7 +63,7 @@ struct TouchState
     /// Finger pressure.
     float pressure_;
     /// Last touched UI element from screen joystick.
-    WeakPtr<UIElement> touchedElement_;
+    UIWidget* touchedWidget_;
```
JoshEngebretson: Maybe want a WeakPtr here? That way if the refcount goes to zero we'll have a null instead of dangling ptr?

Source/AtomicJS/Javascript/JSAPI.cpp
```diff
@@ -480,6 +481,14 @@ void js_push_variant(duk_context *ctx, const Variant& v)
...
         duk_put_prop_index(ctx, -2, 1);
         break;
+    case VAR_INTVECTOR2:
+        intVector2 = v.GetIntVector2();
+        duk_push_array(ctx);
+        duk_push_number(ctx, intVector2.x_);
+        duk_put_prop_index(ctx, -2, 0);
+        duk_push_number(ctx, intVector2.y_);
+        duk_put_prop_index(ctx, -2, 1);
+        break;
```
JoshEngebretson: Nice!

Source/AtomicJS/Javascript/JSInput.cpp
```diff
@@ -0,0 +1,64 @@
...
+        vmap["position"] = touchState->position_;
+        vmap["lastPosition"] = touchState->lastPosition_;
+        vmap["delta"] = touchState->delta_;
+        vmap["pressure"] = touchState->pressure_;
+        vmap["touchedWidget"] = touchState->touchedWidget_;
+
+        js_push_variantmap(ctx, vmap);
+        
+        return 1;
+    }
```
JoshEngebretson: Very nice

Script/Packages/Atomic/Input.json
```diff
@@ -4,7 +4,8 @@
 	"classes" : ["Input"],
 	"overloads" : {
 		"Input" : {
-			"BindButton" : ["UIButton", "int"]
+			"BindButton" : ["UIButton", "int"],
+			"GetTouch" : ["unsigned"]
```
JoshEngebretson: This shouldn't actually be doing anything, because TouchState isn't a scriptable class, so will get ignored.  

So, we'll need to add the custom marshaller to the typescript/haxe defs, and extra points for defining an interface for the object properties/keys it returns

JoshEngebretson:
A couple comments, overall looks really solid.  Good work!

PR #334, opened 2015-10-02
Small fix for issue with scrolling. 

rsredsq:
It was giving an incorrect pointer position.


JoshEngebretson:
Good digging, which issue does this address?

rsredsq:
We don't have an issue for that. I've noticed it, when I was using editor

JoshEngebretson:
Ok, when making a PR please provide a bit more specific information about what is being fixed.  I still have little idea what this PR is fixing :)

rsredsq:
Before that fix, if you try to scroll a scrollbar(by mouse cursor) or move a window, you could see a bug, like widget bumping into some coordinates, and then returns to the current position, and also you could try to select a text it was glitchy too

JoshEngebretson:
Sounds great, thanks!  LANDING!

PR #342, opened 2015-10-09
Adding WebSocket support for native

Type1J:
There's no Emscripten support, yet, but it's planned. Also, a WebRequest, which will replace functionality of Network/HttpRequest will be added with native and Emsciripten support.

(The Network subsystem is not built for Emscripten because it uses kNet, but this new Web subsystem will not have any dependencies that would cause problems in the Emscripten environment.)


JoshEngebretson:
ok to test

Type1J:
Sorry about that. It looks like I didn't merge some stuff from the VM where I did the Emscripten build. I'll push those as soon as I can.

JoshEngebretson:
No problem, passing on Windows, OSX, Linux, Android, iOS.  Does this work on mobile as well or is it desktop only? 

I had a quick look this morning, great work matching standards + getting it hooked up to scripting :1:

I'll have a closer look soon, also adding you to the whitelist so anything pushed will immediately go to CI

JoshEngebretson:
Add to whitelist

Type1J:
It should work on mobile, but I haven't tested. It uses libraries which are known to work on mobile.

JoshEngebretson:
Very solid work, great add, LANDING! :rocket:

Issue #321, opened 2015-09-27
Save function on prefabs requiring multiple clicks


LaraEngebretson:
Bug originally reported here: #159

LaraEngebretson:
The filewatcher needs to see the Cache'd prefab update, and load it into scene. Prefabs are referenced by UUID when adding them to scene, as you can move them around in the project.

There's some sorting that could happen there, the tricky thing being that when laying out prefabs in scenes you need to load by UUID, when scripting, you want to load via path. If you move prefabs around in your project, you don't want to break scene references. It is possible that prefabs should be indexed by UUID somehow (maybe at deploy/play time)

I am relatively certain that this issue is about the file system lag on watcher catching change and updating scene, which makes it seem like you have to hit save multiple times.

From a usability standpoint, the delay should probably cause the button to be disabled until the update.

See issue #386

PR #385, opened 2015-11-02
Reload the cache file

rsredsq:
#321


JoshEngebretson:
Ok, so what was happening here was that by having it look in Cache/ first, it was actually finding the cached prefab as a different resource since the project's root is a resource path.  This is an exciting new issue: https://github.com/AtomicGameEngine/AtomicGameEngine/issues/394

I am going to close this PR out, as I have one that is similar coming up

Issue #341, opened 2015-10-08
Add a loop option to the Sound Source component


LaraEngebretson:
looped is assigned to the Sound resource which isn't ideal because it has to be assigned in script

PR #404, opened 2015-11-03
Added "Looped" attribute to the SoundSource component

rsredsq:
#341


Source/Atomic/Audio/SoundSource.cpp
```diff
@@ -1250,4 +1251,15 @@ void SoundSource::SetSound(Sound* sound)
     if (sound) SetSoundAttr(GetResourceRef(sound, ""));
 }
 
+bool SoundSource::IsLooping() const
```
JoshEngebretson: If you call this GetLooping, there will be a looping property generated

Issue #411, opened 2015-11-06
Check for unsaved changes before closing the editor


LaraEngebretson:
Add a confirmation dialog to allow the user to save changes before closing, or close without saving.

Issue #325, opened 2015-09-29
Check for unsaved changes before closing a project


LaraEngebretson:
If a project gets closed with unsaved changes to a scene or script file, all changes made to that file are gone.

Add a confirmation dialog to allow the user to save changes before closing, or close without saving.

We could also add an auto-save to scene files when the player is launched, the same way the script files behave.

JoshEngebretson:
:+1:

PR #413, opened 2015-11-08
Check for unsaved changes before closing whole project or editor

rsredsq:
#325
#411


rsredsq:
I guess I've checked everything now, and all the things should works fine

Script/AtomicEditor/editor/Editor.ts
```diff
@@ -82,15 +88,29 @@ class Editor extends Atomic.ScriptObject {
         return system.loadProject(event.path);
     }
 
-    handleEditorCloseProject(event) {
-        var system = ToolCore.getToolSystem();
+    checkIfEverythingClosed() {
```
JoshEngebretson: Can we be more descriptive in this function name?   I am not sure what "Everything" is :)

Source/Atomic/UI/UI.cpp
```diff
@@ -150,7 +150,7 @@ void UI::HandleExitRequested(StringHash eventType, VariantMap& eventData)
 
 void UI::Shutdown()
 {
-    SetInputDisabled(true);
+    //SetInputDisabled(true);
```
JoshEngebretson: When commenting out code and leaving it, please make a comment is to why

Script/Packages/Atomic/Core.json
```diff
@@ -10,6 +10,11 @@
 			"SendEvent" : ["StringHash"]
 		}
 	},
+	"overloads" : {
+		"Object" : {
+			"UnsubscribeFromEvent" : ["StringHash"]
```
JoshEngebretson: We're going top need both flavors of this, unsubscribing from global events and from object specific events (like the SubscribeToEvent works)

Script/AtomicEditor/editor/Editor.ts
```diff
@@ -123,15 +143,12 @@ class Editor extends Atomic.ScriptObject {
...
-        Preferences.getInstance().write();
-        EditorUI.shutdown();
-
-    }
-
-    handleEditorEventQuit(data) {
-
-        this.sendEvent("ExitRequested");
-
+      this.subscribeToEvent("AllResourceEditorsClosed", (_) => {
```
JoshEngebretson: The event name "AllResourceEditorsClosed" is confusing here, it could happen during normal use when not exiting, you may actually want to do something when this happens.  This is more of a shutdown event?

Source/AtomicEditor/Editors/ResourceEditor.cpp
```diff
@@ -65,6 +65,7 @@ class EditorTabLayout: public TBLayout
                 }
                 else
                 {
+                    editor_->SendEvent("EditorResourceCloseCanceled");
```
JoshEngebretson: In native code, we need to define events like this: https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/AtomicEditor/Editors/SceneEditor3D/SceneEditor3DEvents.h

... and not use String -> StringHash promotion, on the script side, defining a const var for event name, or using string itself is ok.  The binding generator needs a way to dump all the event interface defs (this would mean modifying the MACRO to capture the event info, which is non-trivial)

There could be a ResourceEditorEvents.h created for resource editor specific events

JoshEngebretson:
I like how small the changes are to get this in.  I have a few comments on it above, let's massage this a bit.  Thanks!

Source/Atomic/UI/UI.cpp
rsredsq: it's really unnecessary to disable UI input after requesting exit. Should I just remove it?

JoshEngebretson: Yes, if you think it is not necessary.  

If we're going to comment out code, but leave it in, we should make a note why.  

BTW, I don't have anyone reviewing my pull requests at the moment, so stuff like this may slip in :)

Source/AtomicEditor/Editors/ResourceEditor.cpp
rsredsq: yeah, I made a ResourceEditorEvents.h file in the first commit, but then when I saw https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/AtomicEditor/Editors/ResourceEditor.cpp#L165 , I thought that it's ok for small event to not define it with macro

Script/AtomicEditor/editor/Editor.ts
rsredsq: AllResourceEditorsClosed event sent only when project close or editor close requested, so it's not gonna be fired during normal use. But, hm, I can't imagine another name for that event

Script/Packages/Atomic/Core.json
rsredsq: Should I implement it in that PR?

Source/AtomicEditor/Editors/ResourceEditor.cpp
JoshEngebretson: Yeah, unfortunately that shouldn't be defined like that either :/  

The problem with letting stuff like this slide is that there ends up being more and more cases of things like this, which leads to it being standard to do.  Let's not introduce more doing this...

Script/AtomicEditor/editor/Editor.ts
JoshEngebretson: Did you see this?

Script/Packages/Atomic/Core.json
JoshEngebretson: We'll skip it for this PR, thanks

Script/AtomicEditor/editor/Editor.ts
JoshEngebretson: Try? :)

Script/AtomicEditor/editor/Editor.ts
rsredsq: yeah, renamed to closeAllResourceEditors

Script/AtomicEditor/editor/Editor.ts
JoshEngebretson: I think what I was saying here is that the name should reflect that it is a shutdown event

Source/AtomicEditor/Editors/ResourceEditorEvents.h
```diff
@@ -0,0 +1,26 @@
...
+{
+
+EVENT(E_EDITORCLOSERESOURCE, EditorCloseResource)
+{
+PARAM(P_EDITOR, Editor);            // ResourceEditor
+PARAM(P_NAVIGATE, NavigateToAvailableResource); // bool
+}
+
+EVENT(E_EDITORRESOURCECLOSECANCELED, EditorResourceCloseCanceled)
+{
```
JoshEngebretson: Here we have:
 
EditorCloseResource
EditorResourceCloseCanceled

Probably should make them consistent so:

EditorResourceClose
EditorResourceCloseCanceled

Script/AtomicEditor/editor/Editor.ts
```diff
@@ -82,7 +95,29 @@ class Editor extends Atomic.ScriptObject {
         return system.loadProject(event.path);
     }
 
+    closeAllResourceEditors() {
+        var editor = EditorUI.getCurrentResourceEditor();
+        if (!editor) {
+          if (this.exitRequested) {
+              Editor.Exit();
```
JoshEngebretson: Editor.exit() here instead of Editor.Exit()?

Source/AtomicEditor/Editors/ResourceEditorEvents.h
rsredsq: EditorCloseResource has already been existed so I haven't renamed it

Script/AtomicEditor/editor/Editor.ts
```diff
@@ -55,6 +60,14 @@ class Editor extends Atomic.ScriptObject {
             Preferences.getInstance().registerRecentProject(data.projectPath);
         });
 
+        this.subscribeToEvent("EditorResourceCloseCanceled", (data) => {
+            //if user canceled closing the resource, then user has changes that he doesn't want to lose
+            //so cancel exit/project close request and unsubscribe from event to avoid closing all the editors again
```
JoshEngebretson: Nice commenting

JoshEngebretson:
A couple comments, otherwise much improved, we can have @LaraEngebretson test this once updated

Source/AtomicEditor/Editors/ResourceEditorEvents.h
JoshEngebretson: Well, then how about EditorCloseResourceCanceled

JoshEngebretson:
Thanks for making this better, LANDING!

Issue #313, opened 2015-09-24
Windows editor crashes on nonexistent Android API refresh


MrMormon:
Runtime context: Windows 10, community editor activated with free product key, Android Studio is not installed

Steps to reproduce: Create new 3D project (may work with any template or openable project), go to Build -> Build Settings -> Android -> (Android API) Refresh button

Result: editor crashes with WIndows dialog popup (program *.exe has stopped working - clicking Cancel exits)

Expected behavior: popup dialog/console message "I can't find any Android development tools with your setup, please accept my condolences.", or nothing

standtall007:
It is because you need to specify the Android SDK path first :) like: "C:\Android\sdk" but you can click the Choose button to specify the location a little easier :) Hope this helps.

MrMormon:
Android Studio is not installed, and since it's the community edition, I chose 3D over Android support. But it shouldn't crash.

standtall007:
if you have installed the 3D add-on instead of Android, why are you then trying to choose Android? doesn't make sense :P

MrMormon:
I was just curious what it would say? I dunno. :)

standtall007:
oh okay :D

JoshEngebretson:
Thanks for the report, we do need to address any case where the editor crashes.

standtall007:
I didn't know for a fact it was a bug :P
but i can perfectly understand you Josh it wasn't my intentions to be to cruel i was just so tired when i wrote to Mr.Mormon and i apoligize to you Mr.Mormon for writing that way to you i am really sorry :)

JoshEngebretson:
No worries, it is good to get all the details :)

MrMormon:
That's fine, standtall007, I didn't interpret your comments as hostile.

standtall007:
thanks to both of you :)

MrMormon:
Um, hugs? :P

standtall007:
Hey i'm no teletubby XD

PR #414, opened 2015-11-08
Fix when editors crashes on nonexistent Android API refresh

rsredsq:
#313


Script/AtomicEditor/ui/modal/build/platforms/AndroidSettingsWidget.ts
```diff
@@ -125,6 +125,8 @@ class AndroidSettingsWidget extends Atomic.UIWidget implements BuildSettingsWind
 
     refreshAndroidTargets() {
 
+        if (ToolCore.toolEnvironment.toolPrefs.androidSDKPath.length <= 0) return;
```
JoshEngebretson: Wouldn't it be better to check if the path exists and possibly report an error (at least to the console, using LOGERRORF) ?  I don't think that check should be here, more like in the native platform.refreshAndroidTargets

JoshEngebretson:
Awesome, LANDING!

PR #415, opened 2015-11-08
When breaking a prefab node, update node to remove the prefab save/un…

rsredsq:
…do/break node ui
#407

### DIFF
--- a/Script/AtomicEditor/ui/frames/inspector/NodeInspector.ts
+++ b/Script/AtomicEditor/ui/frames/inspector/NodeInspector.ts
@@ -123,7 +123,7 @@ class NodeInspector extends ScriptWidget {
         this.node = node;
 
         node.scene.sendEvent("SceneEditSerializable", { serializable: node, operation: 0});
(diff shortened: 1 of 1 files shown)


Script/AtomicEditor/ui/frames/inspector/NodeInspector.ts
```diff
@@ -269,6 +269,8 @@ class NodeInspector extends ScriptWidget {
 
                     prefabComponent.breakPrefab();
 
+                    this.sendEvent("EditorActiveNodeChange", { node: this.node });
```
JoshEngebretson: We're not actually changing the active node here, and while this may reload the inspector UI now, it may not always be the case. (There is work to be done so that the Node/Component inspectors refresh)  

For now, this should disable, make invisible, or remove the prefab UI layout instead.

rsredsq: Break prefab also remove a PrefabComponent. I think it's ok to leave that for now, until we will refactor editor inspectors. The same thing for save/undo buttons on prefab UI layout, they send EditorActiveNodeChange event while they shouldn't do it

JoshEngebretson:
Ok, that's a fair point.  I will land this :)

Issue #361, opened 2015-10-21
Using arrow keys in Hierarchy panel should select node in scene view


LaraEngebretson:
It would also be helpful when selecting an object in the scene view, if its node also became active in the hierarchy panel.

Originally requested here: #159

PR #425, opened 2015-11-09
Use arrows keys to navigate in hierarchy and project panels

rsredsq:
#361

### DIFF
--- a/Source/Atomic/UI/UIInput.cpp
+++ b/Source/Atomic/UI/UIInput.cpp
@@ -427,7 +427,7 @@ void UI::HandleKey(bool keydown, int keycode, int scancode)
     }
     else
     {
-        InvokeKey(this, rootWidget_, keycode, specialKey, mod, keydown);
+        InvokeKey(this, rootWidget_, 0, specialKey, mod, keydown);
     }
 
 }
--- a/Source/Atomic/UI/UIInput.cpp
+++ b/Source/Atomic/UI/UIInput.cpp
@@ -427,7 +427,7 @@ void UI::HandleKey(bool keydown, int keycode, int scancode)
     }
     else
     {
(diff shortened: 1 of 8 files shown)


Source/Atomic/UI/UIListView.cpp
```diff
@@ -230,7 +230,6 @@ ListViewItemWidget::ListViewItemWidget(ListViewItem *item, ListViewItemSource *s
     SetLayoutDistribution(LAYOUT_DISTRIBUTION_GRAVITY);
     SetLayoutDistributionPosition(LAYOUT_DISTRIBUTION_POSITION_LEFT_TOP);
     SetPaintOverflowFadeout(false);
-    SetCapturing(false);
```
JoshEngebretson: This is needed to fix an issue when dragging an item up or down in the list view will cause the list to scroll

Source/Atomic/UI/UIInput.cpp
```diff
@@ -427,7 +427,7 @@ void UI::HandleKey(bool keydown, int keycode, int scancode)
     }
     else
     {
-        InvokeKey(this, rootWidget_, 0, specialKey, mod, keydown);
+        InvokeKey(this, rootWidget_, keycode, specialKey, mod, keydown);
```
JoshEngebretson: Sending the keycode here, causes an issue, which I cannot recall at the moment, however it was recent

JoshEngebretson: Ok, one side effect of this change is that pressing TAB in an inspector field inserts a tab into the field instead of moving to the next widget.  There are likely other side effects. 

InvokeKey currently expects no keycode for when a special key is being invoked.  What about arrow keys is requiring the keycode to be passed?

LaraEngebretson:
Great job, thanks!!!

Issue #427, opened 2015-11-10
Running with --player --project, first open of project fails looking for ToolPrefs.json


mattbenic:
To repro:
-Create a new atomic project
-Commit that project to source control without it's Cache directory
-Clone a fresh copy of the project
-Open this project using a VS2015 atomic solution and --player --project path/to/project commandline arguments
-Note an error popup about failing to find ToolPrefs.json

Workaround:
-Open the project once as normal in AtomicEditor
-Project can then be run using --player --project

JoshEngebretson:
The project has to be loaded at least once into the editor before invoking --player mode, as in player mode the cache will not be populated.  

I made this error clear in the dialog in #433

PR #433, opened 2015-11-10
Don't report missing ToolPrefs as this isn't an error, raise an error…

JoshEngebretson:
… dialog if --player command line is invoked without loading project into editor first at least once


JoshEngebretson:
Closes #427

PR #446, opened 2015-11-16
Support for static class functions in JSBind. A version of ToString that takes an explicit vs_args parameter. 

mattbenic:


JoshEngebretson:
Very nice add, thanks!  LANDING!

PR #448, opened 2015-11-17
Fix build directory reference in GenerateVS2015.bat

Type1J:
I manage a few copies of the engine based on what I'm adding. These copies are in the same root directory, but they are all named differently: "AtomicGameEngineWS", "AtomicGameEngineWR", "AtomicGameEngineWebM", etc.

When I was using CMake directly, that was not an issue. However, now I'm using jake to invoke CMake, and that script has a hard coded "AtomicGameEngine" directory in it for the source directory, even though it had already created a binary directory based on the source directory's name.

This patch passes that directory name through to the .bat file where it was previously hardcoded, so that my directory naming scheme works again.

### DIFF
--- a/Build/Scripts/BuildWindows.js
+++ b/Build/Scripts/BuildWindows.js
@@ -89,7 +89,7 @@ namespace('build', function() {
 
     var cmds = [];
 
-    cmds.push(atomicRoot + "Build/Scripts/Windows/GenerateVS2015.bat");
+    cmds.push(atomicRoot + "Build/Scripts/Windows/GenerateVS2015.bat " + atomicRoot);
 
     jake.exec(cmds, function() {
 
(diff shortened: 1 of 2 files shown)


JoshEngebretson:
add to whitelist
ok to test

JoshEngebretson:
Good improvement, thanks!  LANDING!

JoshEngebretson:
BTW, the bat files are really intended to get a local dev environment bootstrapped (and for artist/script writers who want a really simple way to build the editor from master).  

For development, it is possible to run CMake directly or modify the cmake files and VS will pick up that it needs to reconfigure automatically.  There is also now the "GenerateScriptBindings" project, that when built, parses all the script packages, creates any missing source files for CMake to glob, builds the AtomicTool, and generates the new bindings.  Lemme know if you are having issues or problems!

PR #450, opened 2015-11-18
WebSocket cleanup

Type1J:
Don't get too excited. This doesn't have the Emscripten part, yet.

I have to apologize:

I realized that the WebSocket that I've been using and testing for a while now is not actually compatible with what I sent to you the first time. I cleaned up the code, and made this pull request to fix those incompatibilities.

Basically, this version does a lot less copying of messages. I'd like to get it to no copies, but it still might copy here or there. One main difference is that it doesn't copy a message into JavaScript space until it's requested. (An ID int could be at the beginning of a message, and the JS code coould choose not to read it, which prevents an unnecessary copy. This is done be making the message a Deserializer instead of a String or PODVector (Yes, JavaScript PODVector<unsigned char> support is on it's way from me as well, but not in this pull.). Specifically, it uses a BufferQueue instance, even though messages would probably only ever be 1 buffer. I made the BufferQueue class for a better HTTP interface, not really for WebSockets, but because I had to bind it in JSIO, I wanted to keep the number of mixins necessary down for simplicity.

I promise to never make another incompatible change unless it's really necessary. Just to be clear, the only breaking change is with the WebSocket "message" event handler. Instead of "data" with a String object, you get "message", which is a Deserializer, and from that you can read a String or a Uint8Array, if you want the binary version. It's not present in JS memory until it's read from the Deserializer.

A corresponding AtomicExamples change is on it's way as well. It's only a 1 line change from:

    console.log("WebSocket Message: " + event.data);

to:

    console.log("WebSocket Message: " + event.message.readString());

Again, I'm sorry for not sending this change sooner, since it does require existing code to make a similar change in their "message" event handlers, but I know how the Atomic Game Engine works much better now, and I feel pretty strongly that this is a good change.

### DIFF
--- a/Script/Packages/Atomic/IO.json
+++ b/Script/Packages/Atomic/IO.json
@@ -1,7 +1,7 @@
 {
 	"name" : "IO",
 	"sources" : ["Source/Atomic/IO"],
-	"classes" : ["Log", "File", "FileSystem", "FileWatcher"],
+	"classes" : ["Log", "File", "FileSystem", "FileWatcher", "BufferQueue"],
 	"overloads" : {
 		"File" : {
 			"File" : ["Context", "String", "FileMode"]
--- a/Source/Atomic/Container/Vector.h
+++ b/Source/Atomic/Container/Vector.h
@@ -922,10 +922,10 @@ public:
     /// Return whether vector is empty.
     bool Empty() const { return size_ == 0; }
 
-private:
     /// Return the buffer with right type.
     T* Buffer() const { return reinterpret_cast<T*>(buffer_); }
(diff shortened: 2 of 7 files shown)


Type1J:
I should also mention that there are a few (more than 1) crashes that this patch fixes, even though they were rare.

JoshEngebretson:
This looks really great and I think some API tweaks, especially ones that are well founded, are fine until we're closer to a 1.0 release.

Oooooh, PODVector support in JS? Do tell! :)

Thanks again, really appreciate the contribution.... LANDING!

Issue #375, opened 2015-10-28
Remember player positioning when run from editor


mattbenic:
Artist request:
*When playing from the editor, the player opens in the same position, blocking the editor. It should remember the last place it was moved to.

JoshEngebretson:
@rsredsq Instead of saving the position at exit, how about catching E_SCREENMODE and E_WINDOWPOS and forwarding to the editor process to save?

https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/Atomic/Graphics/GraphicsEvents.h#L31

I think this will be a lot better and not complicate the process exit

Issue #324, opened 2015-09-29
Add a check to prevent multiple Atomic Player windows


LaraEngebretson:
Disable the Play button in the editor when an Atomic Player window is open to prevent multiple player  windows from being open simultaneously.

Issue originally reported here: #159

PR #451, opened 2015-11-18
Remember player positioning when run from editor and Prevent multiple Atomic Player windows

rsredsq:
#324 #375


LaraEngebretson:
- The editor crashes when I hit "stop" on Roboman3D (Windows & OSX). Can you reproduce this?
- The position doesn't save (or restore?) on OSX.
- On SpaceGame the position does not save when I use the "stop" button, however, closing the player window does save the position.

Script/AtomicEditor/editor/Editor.ts
```diff
@@ -34,12 +34,19 @@ class Editor extends Atomic.ScriptObject {
...
         Editor.instance = this;
 
+        Preferences.getInstance().read();
+
+        var editorWindow = Preferences.getInstance().editorWindow;
+
+        if (editorWindow) {
+            Atomic.graphics.setWindowSize(editorWindow.width, editorWindow.height);
+            Atomic.graphics.setWindowPosition(editorWindow.x, editorWindow.y);
+        }
```
JoshEngebretson: It would be good if the preferences for the window size and position were parsed at Editor start, and used during the creation of the editor window.  Otherwise, the editor window is created, then immediately resized and moved.

Script/AtomicEditor/editor/Editor.ts
```diff
@@ -53,6 +60,10 @@ class Editor extends Atomic.ScriptObject {
             this.handleProjectUnloaded(data)
         });
 
+        this.subscribeToEvent("IPCPlayerWindow", (data) => {
```
JoshEngebretson: "IPCPlayerWindow" can this be a more descriptive event name?

Script/AtomicEditor/editor/Preferences.ts
```diff
@@ -77,6 +86,7 @@ class Preferences {
     }
 
     write(): void {
+        if (this._configCorrupted) return;
```
JoshEngebretson: We should save, otherwise you'll get into a situation where prefs will never be saved unless you hand edit them.  If you do hand edit them and break the JSON, that is your problem :)  I do not believe we need the _configCorrupted member, and the log should say "Editor preference file invalid, regenerating default configuration"

Script/AtomicEditor/editor/Preferences.ts
```diff
@@ -100,9 +126,18 @@ class Preferences {
     }
 }
 
+class WindowData {
```
JoshEngebretson: This can be an interface as it isn't instantiated

Source/AtomicEditor/EditorMode/AEEditorMode.cpp
```diff
@@ -140,7 +153,19 @@ bool EditorMode::PlayProject(bool debug)
 
 bool EditorMode::PlayProjectDebug()
 {
-    return PlayProject(true);
+    return PlayProject("", true);
+}
+
+void EditorMode::HandleIPCPlayerExitRequest(StringHash eventType, VariantMap& eventData)
+{
+    if (!playerBroker_) return;
```
JoshEngebretson: When does this happen?

JoshEngebretson:
Alright, overall this is looking good.  There are a few nits that it would be good to address, thanks!

JoshEngebretson:
Another note, the player window should be resizeable, so we're testing not just position preferences here.  

This is where it is set for editor mode:

https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/AtomicEditor/Application/AEEditorApp.cpp#L82

Script/AtomicEditor/editor/Editor.ts
rsredsq: Hm, I guess that jsvm context created after window initialization

Script/AtomicEditor/editor/Editor.ts
rsredsq: IPCPlayerWindowData?

Script/AtomicEditor/editor/Preferences.ts
rsredsq: ok

Script/AtomicEditor/editor/Preferences.ts
rsredsq: ok

Source/AtomicEditor/EditorMode/AEEditorMode.cpp
rsredsq: I guess we don't need that function :)

Script/AtomicEditor/editor/Editor.ts
JoshEngebretson: You would need to peek at the JSON to get the values, I would not use the JSVM

Script/AtomicEditor/editor/Editor.ts
JoshEngebretson: IPCPlayerWindowChanged?

Source/AtomicEditor/Application/AEEditorCommon.cpp
```diff
@@ -150,4 +155,38 @@ void AEEditorCommon::Stop()
 #endif
 }
 
+
+bool AEEditorCommon::ReadPreferences()
+{
+
+    FileSystem* fileSystem = GetSubsystem<FileSystem>();
```
JoshEngebretson: Does the ResourceCache subsystem exist here?  If so, can we use the GetTempResource method?

https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/Atomic/Resource/ResourceCache.h#L142

I am not sure if this is wanting a file in the resource dirs, or if you give it an absolute path, it will load it

JoshEngebretson:
Ok, just a couple code style comments + sanity check, and then this needs to be tested before landing

LaraEngebretson:
I ran a quick test on OSX:
* Resizing the editor doesn't update the UI layout, works on master
* Position of player isn't saving
* Editor position and size seems to be working, yay!

LaraEngebretson:
One additional issue:
* Unable to resize the player window

LaraEngebretson:
I tested on OSX and Windows:

* The Play button should change to the Stop icon when using Ctrl-P/Cmd-P and when selecting Edit/Play from the editor's menu
* Selecting Edit/Play from the editor opens additional players
* Resizing the player spams the player output with "Failed to lock index buffer" (windows only)
* Starting the editor with the player on 2nd monitor and the monitor isn't hooked up, you can't get the player back (windows only)

rsredsq:
@LaraEngebretson thanks for testing! The last two issues better make them as a separate issues

JoshEngebretson:
Good to go, thanks!  LANDING!

PR #453, opened 2015-11-22
Variant to Uint8Array mapping

Type1J:
The `Uint8Array` will now map to `PODVector<unsigned char>` for Variant types. That means that you can now pass binary data to event callbacks. I singled out this typed-array type because many JS serialization formats (like [EJSON](https://www.meteor.com/ejson)) which support binary data work with it exclusively.

I haven't looked into the JSBind code, yet. The last step would be to allow `PODVector<unsigned char>` to be used as a function parameter or return value. I'm not sure where that code is.

@JoshEngebretson: Could you point me to the JSBind code where I would add a type as a parameter/return value? I'd like to get the `WebSocket::SendBinary()` call working from a script, and it takes a `const PODVector<unsigned char>&` as a parameter.

### DIFF
--- a/Source/AtomicJS/Javascript/JSAPI.cpp
+++ b/Source/AtomicJS/Javascript/JSAPI.cpp
@@ -235,6 +235,42 @@ void js_object_to_variantmap(duk_context* ctx, int objIdx, VariantMap &v)
 
 }
 
+duk_bool_t js_check_is_buffer_and_get_data(duk_context* ctx, duk_idx_t idx, void** data, duk_size_t* size)
+{
+    void* temp;
+    if (duk_is_buffer(ctx, idx))
+    {
+        temp = duk_get_buffer_data(ctx, idx, size);
+        if (data)
+        {
+            *data = temp;
+        }
+        return true;
+    }
+    if (!(duk_is_object(ctx, idx) &&
+        duk_has_prop_string(ctx, idx, "length") &&
+        duk_has_prop_string(ctx, idx, "byteLength") &&
+        duk_has_prop_string(ctx, idx, "byteOffset") &&
+        duk_has_prop_string(ctx, idx, "BYTES_PER_ELEMENT")))
+    {
+        if (data)
+        {
+            *data = nullptr;
+        }
+        if (size)
+        {
+            *size = 0;
+        }
+        return false;
+    }
+    temp = duk_require_buffer_data(ctx, idx, size);
+    if (data)
+    {
+        *data = temp;
+    }
+    return true;
+}
+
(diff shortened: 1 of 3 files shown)


Source/AtomicJS/Javascript/JSAPI.cpp
```diff
@@ -473,6 +520,15 @@ void js_push_variant(duk_context *ctx, const Variant& v)
     case VAR_STRING:
         duk_push_string(ctx, v.GetString().CString());
         break;
+    case VAR_BUFFER:
+    {
```
JoshEngebretson: Excellent, I filed an issue to cleanup the scope usage in js_push_variant: https://github.com/AtomicGameEngine/AtomicGameEngine/issues/454

Source/AtomicJS/Javascript/JSAPI.cpp
```diff
@@ -473,6 +520,15 @@ void js_push_variant(duk_context *ctx, const Variant& v)
...
+        const PODVector<unsigned char>& buffer(v.GetBuffer()); // The braces are to scope this reference.
+        duk_push_fixed_buffer(ctx, buffer.Size());
+        duk_push_buffer_object(ctx, -1, 0, buffer.Size(), DUK_BUFOBJ_UINT8ARRAY);
```
JoshEngebretson: Nice, and seeing the UINT8ARRAY usage from duktape makes me happy

JoshEngebretson:
Another solid PR, thanks and also for the license! :)

There is some support for Vector<> currently in JSBind, this was added quite quickly and perhaps isn't as clean as it could be... PODVector for various primitive types mapping to bool. number, string,  could be nice, though if it is added to JSBind solely for the UINT8 case, that is fine too

Here are some code pointers:

Parameter marshal in (read only):

https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/ToolCore/JSBind/JavaScript/JSFunctionWriter.cpp#L182

Return value:

https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/ToolCore/JSBind/JavaScript/JSFunctionWriter.cpp#L501

C++ parsing:

https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/ToolCore/JSBind/JSBHeaderVisitor.h#L80

Const check for parameter:

https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/ToolCore/JSBind/JSBHeaderVisitor.h#L198

Type1J:
I'm glad that you like my work. I have actually had a license for some time now, that license was actually bought for a friend.

PR #459, opened 2015-11-23
Scoping variables in js_push_variant

Type1J:
I looked at this function a bit deeper, and I noticed that there were a good number of copies out of the variant, and it constructed all of the possible temporary variables whether it used them or not. It now only copies the variable out of the Variant once, and it only constructs a temporary type when it's needed.


JoshEngebretson:
This is a really nice cleanup, thanks! LANDING!

PR #466, opened 2015-11-26
Player setup doesn't override engine parameters, constants for .pak file extensions.

eugenegous:


JoshEngebretson:
Ok to test
Add to whitelist

JoshEngebretson:
This looks good, though somehow it is breaking the Android build on CI.  I've looked it over a couple times and don't see how this could possibly be tripping a compiler error in stanhull :/  

I'll have a look at this tomorrow and land it.

JoshEngebretson:

Source/Atomic/Resource/ResourceCache.cpp
```diff
@@ -64,6 +64,8 @@ static const char* checkDirs[] =
 };
 
 static const SharedPtr<Resource> noResource;
+const char* Atomic::PAK_EXTENSION = ".pak";
+const char* Atomic::PAK_FILTER = "*.pak";
```
JoshEngebretson: Ok, Android doesn't like the Atomic:: here, as already in the Atomic namespace.  I'll land this and then fix it up in post :)

PR #472, opened 2015-11-29
Added deselect functionality to the scene view

rsredsq:


Source/AtomicEditor/Editors/SceneEditor3D/SceneView3D.cpp
```diff
@@ -443,6 +447,13 @@ bool SceneView3D::OnEvent(const TBWidgetEvent &ev)
         if (ev.ref_id == TBIDC("close"))
             return false;
     }
+    if (ev.type == EVENT_TYPE_KEY_UP)
+    {
+        if (ev.special_key == TB_KEY_ESC)
+        {
+            sceneEditor_->GetSelection()->Clear();
+        }
```
JoshEngebretson: Does this clear selection if you are in the inspector in like a string field and hit esc?

JoshEngebretson: I see that we're in an OnEvent here, so it shouldn't, though want to check

rsredsq: sceneview is always focused because of https://github.com/rsredsq/AtomicGameEngine/blob/88e12f214835cb636789c5050b22ac1f7acefc6d/Source/AtomicEditor/Editors/SceneEditor3D/SceneView3D.cpp#L340
then all the button events send to the sceneview if it's opened, I reckon we don't need that event

Issue #464, opened 2015-11-24
Unnecessary Editor menus should be disabled when no scene is open


mattbenic:
All menus other than "Atomic Editor", "File" and "Developer" should be disabled when no scene is open in the editor.

mattbenic:
That's what I meant, apologies for unclear issue

PR #483, opened 2015-12-04
Disable unnecessary menus when there is no open project

rsredsq:
This is how menu does looks like, until project is open

#464


JoshEngebretson:
Hey!  Nice and clean, LANDING!  Thanks :)

Issue #326, opened 2015-09-29
Add a hover box to descriptions getting truncated in the Inspector panel


LaraEngebretson:
If the text is too long it gets cut off. This prevents users from seeing the whole name/description of a setting. 

![textcutoff](https://cloud.githubusercontent.com/assets/722928/10177351/9282c5a6-66ae-11e5-8f80-348893de376f.png)

Originally reported in #159

JoshEngebretson:
This probably requires tooltips which TurboBadger does not currently have, I wonder if we could manage these in the UI module without many/any TB modifications

PR #486, opened 2015-12-06
Tooltip implementation

rsredsq:
#326

### DIFF
--- a/Source/Atomic/UI/UI.cpp
+++ b/Source/Atomic/UI/UI.cpp
@@ -36,6 +36,8 @@
 #include <TurboBadger/tb_tab_container.h>
 #include <TurboBadger/tb_toggle_container.h>
 #include <TurboBadger/tb_scroll_container.h>
+#include <TurboBadger/tb_menu_window.h>
+#include <TurboBadger/tb_popup_window.h>
 #include <TurboBadger/image/tb_image_widget.h>
 
 void register_tbbf_font_renderer();
@@ -79,6 +81,8 @@ using namespace tb;
 #include "UISeparator.h"
 #include "UIDimmer.h"
 #include "UISelectDropdown.h"
+#include "UIMenuWindow.h"
+#include "UIPopupWindow.h"
 
 #include "SystemUI/SystemUI.h"
 #include "SystemUI/SystemUIEvents.h"
@@ -535,9 +539,50 @@ void UI::HandleUpdate(StringHash eventType, VariantMap& eventData)
         exitRequested_ = false;
         return;
     }
+    
+    mouseStayedTime += eventData[Update::P_TIMESTEP].GetFloat();
+
+    if (mouseStayedTime >= 0.5f)
+    {
+        UIWidget* hoveredWidget = GetHoveredWidget();
+        if (hoveredWidget && !tooltip_ && (hoveredWidget->GetShorten() || hoveredWidget->GetTooltip().Length() > 0))
+        {
+            tooltip_ = new UIPopupWindow(context_, hoveredWidget, "tooltip");
+            UILayout* tooltipLayout = new UILayout(context_, UI_AXIS_Y, true);
+            if (hoveredWidget->GetShorten())
+            {
+                UIEditField* fullTextField = new UIEditField(context_, true);
+                fullTextField->SetAdaptToContentSize(true);
+                fullTextField->SetText(hoveredWidget->GetText());
+                fullTextField->SetReadOnly(true);
+                tooltipLayout->AddChild(fullTextField);
+            }
+            if (hoveredWidget->GetTooltip().Length() > 0)
+            {
+                UIEditField* tooltipTextField = new UIEditField(context_, true);
+                tooltipTextField->SetAdaptToContentSize(true);
+                tooltipTextField->SetText(hoveredWidget->GetTooltip());
+                tooltipTextField->SetReadOnly(true);
+                tooltipLayout->AddChild(tooltipTextField);
+            }
+            Input* input = GetSubsystem<Input>();
+            IntVector2 mousePosition = input->GetMousePosition();
+            tooltip_->AddChild(tooltipLayout);
+            tooltip_->Show(mousePosition.x_ + 1, mousePosition.y_ + 1);
+        }
+    }
+    else 
+    {
+        if (tooltip_) tooltip_->Close();
+    }
 
     SendEvent(E_UIUPDATE);
     TBMessageHandler::ProcessMessages();
+}
+
+UIWidget* UI::GetHoveredWidget()
+{
+    return WrapWidget(TBWidget::hovered_widget);
 }
 
 bool UI::IsWidgetWrapped(tb::TBWidget* widget)
@@ -639,7 +684,6 @@ UIWidget* UI::WrapWidget(tb::TBWidget* widget)
         widgetWrap_[widget] = sep;
         return sep;
     }
-
 
     if (widget->IsOfType<TBContainer>())
     {
--- a/Source/Atomic/UI/UI.cpp
+++ b/Source/Atomic/UI/UI.cpp
@@ -551,24 +551,20 @@ void UI::HandleUpdate(StringHash eventType, VariantMap& eventData)
             UILayout* tooltipLayout = new UILayout(context_, UI_AXIS_Y, true);
             if (hoveredWidget->GetShortened())
             {
-                UIEditField* fullTextField = new UIEditField(context_, true);
-                fullTextField->SetAdaptToContentSize(true);
+                UITextField* fullTextField = new UITextField(context_, true);
                 fullTextField->SetText(hoveredWidget->GetText());
-                fullTextField->SetReadOnly(true);
                 tooltipLayout->AddChild(fullTextField);
             }
             if (hoveredWidget->GetTooltip().Length() > 0)
             {
-                UIEditField* tooltipTextField = new UIEditField(context_, true);
-                tooltipTextField->SetAdaptToContentSize(true);
+                UITextField* tooltipTextField = new UITextField(context_, true);
                 tooltipTextField->SetText(hoveredWidget->GetTooltip());
-                tooltipTextField->SetReadOnly(true);
                 tooltipLayout->AddChild(tooltipTextField);
             }
             Input* input = GetSubsystem<Input>();
             IntVector2 mousePosition = input->GetMousePosition();
             tooltip_->AddChild(tooltipLayout);
-            tooltip_->Show(mousePosition.x_ + 1, mousePosition.y_ + 1);
+            tooltip_->Show(mousePosition.x_ + 8, mousePosition.y_ + 8);
         }
     }
     else 
--- a/Source/Atomic/UI/UI.cpp
+++ b/Source/Atomic/UI/UI.cpp
@@ -540,16 +540,16 @@ void UI::HandleUpdate(StringHash eventType, VariantMap& eventData)
         return;
     }
     
-    mouseStayedTime += eventData[Update::P_TIMESTEP].GetFloat();
+    tooltipHoverTime_ += eventData[Update::P_TIMESTEP].GetFloat();
 
-    if (mouseStayedTime >= 0.5f)
+    if (tooltipHoverTime_ >= 0.5f)
     {
         UIWidget* hoveredWidget = GetHoveredWidget();
-        if (hoveredWidget && !tooltip_ && (hoveredWidget->GetShorten() || hoveredWidget->GetTooltip().Length() > 0))
+        if (hoveredWidget && !tooltip_ && (hoveredWidget->GetShortened() || hoveredWidget->GetTooltip().Length() > 0))
         {
-            tooltip_ = new UIPopupWindow(context_, hoveredWidget, "tooltip");
+            tooltip_ = new UIPopupWindow(context_, true, hoveredWidget, "tooltip");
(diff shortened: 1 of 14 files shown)


Source/Atomic/UI/UI.cpp
```diff
@@ -535,11 +539,52 @@ void UI::HandleUpdate(StringHash eventType, VariantMap& eventData)
         exitRequested_ = false;
         return;
     }
+    
+    mouseStayedTime += eventData[Update::P_TIMESTEP].GetFloat();
```
JoshEngebretson: This could be implemented as a TB custom event, based on hover time: https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/ThirdParty/TurboBadger/tb_inline_select.cpp#L160

Source/Atomic/UI/UI.cpp
```diff
@@ -535,11 +539,52 @@ void UI::HandleUpdate(StringHash eventType, VariantMap& eventData)
...
+
+    if (mouseStayedTime >= 0.5f)
+    {
+        UIWidget* hoveredWidget = GetHoveredWidget();
```
JoshEngebretson: With the TB custom event for hover, we could send the hover widget and capture the tooltip text without wrapping the widget

Source/Atomic/UI/UI.cpp
```diff
@@ -535,11 +539,52 @@ void UI::HandleUpdate(StringHash eventType, VariantMap& eventData)
...
+        if (hoveredWidget && !tooltip_ && (hoveredWidget->GetShorten() || hoveredWidget->GetTooltip().Length() > 0))
+        {
+            tooltip_ = new UIPopupWindow(context_, hoveredWidget, "tooltip");
```
JoshEngebretson: Let's make one tooltip UI, and then reuse it with visibility gone flag

Source/Atomic/UI/UI.cpp
```diff
@@ -535,11 +539,52 @@ void UI::HandleUpdate(StringHash eventType, VariantMap& eventData)
...
+                tooltipLayout->AddChild(fullTextField);
+            }
+            if (hoveredWidget->GetTooltip().Length() > 0)
+            {
+                UIEditField* tooltipTextField = new UIEditField(context_, true);
+                tooltipTextField->SetAdaptToContentSize(true);
+                tooltipTextField->SetText(hoveredWidget->GetTooltip());
+                tooltipTextField->SetReadOnly(true);
+                tooltipLayout->AddChild(tooltipTextField);
+            }
```
JoshEngebretson: Nice work wrangling the UI :)

Source/Atomic/UI/UI.cpp
```diff
@@ -535,11 +539,52 @@ void UI::HandleUpdate(StringHash eventType, VariantMap& eventData)
...
+        if (tooltip_) tooltip_->Close();
+    }
 
     SendEvent(E_UIUPDATE);
     TBMessageHandler::ProcessMessages();
 }
 
+UIWidget* UI::GetHoveredWidget()
+{
+    return WrapWidget(TBWidget::hovered_widget);
```
JoshEngebretson: I think this is a good add, cool

Source/Atomic/UI/UI.h
```diff
@@ -144,6 +147,8 @@ class UI : public Object, private tb::TBWidgetListener
     HashMap<tb::TBWidget*, SharedPtr<UIWidget> > widgetWrap_;
     HashMap<unsigned, String> tbidToString_;
 
+    WeakPtr<UIPopupWindow> tooltip_;
```
JoshEngebretson: SharedPtr will keep it alive, let's make sure when we shutdown it gets cleared

Source/Atomic/UI/UI.h
```diff
@@ -152,6 +157,8 @@ class UI : public Object, private tb::TBWidgetListener
     bool skinLoaded_;
     bool consoleVisible_;
     bool exitRequested_;
+    
+    float mouseStayedTime;
```
JoshEngebretson: We should be able to remove this, if not let's talk about the name

Source/Atomic/UI/UIPopupWindow.h
```diff
@@ -0,0 +1,51 @@
...
+
+#pragma once
+
+#include "UIWidget.h"
+
+namespace Atomic
+{
+
+class UIPopupWindow: public UIWidget
+{
```
JoshEngebretson: There is currently a trick in adding new widgets, which needs to be cleaned up to be self registering, but in the meantime:

https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/Atomic/UI/UI.cpp#L603

and

https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/AtomicJS/Javascript/JSUI.cpp#L84

Not great, I know :)

Source/Atomic/UI/UIPopupWindow.cpp
```diff
@@ -0,0 +1,73 @@
...
+#include "UIWidget.h"
+#include "UIPopupWindow.h"
+
+using namespace tb;
+
+namespace Atomic
+{
+
+    UIPopupWindow::UIPopupWindow(Context* context, UIWidget* target, const String& id) : UIWidget(context, false)
+    {
```
JoshEngebretson: The constructor needs to be able to wrap a widget, as well as create one, for example:

https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/Atomic/UI/UIButton.cpp#L40

Source/Atomic/UI/UIWidget.h
```diff
@@ -250,10 +258,18 @@ class UIWidget : public Object, public tb::TBWidgetDelegate
     bool GetCaptured();
 
     void SetCapturing(bool capturing);
```
JoshEngebretson: I should have flagged these before, GetCapturing/SetCapturing are unclear as to what they do.  Can you describe them a bit, and they we can work on what the names should be?

Source/Atomic/UI/UIWidget.h
```diff
@@ -250,10 +258,18 @@ class UIWidget : public Object, public tb::TBWidgetDelegate
...
-    bool GetCapturing() { return widget_->GetCapturing(); }
+    bool GetCapturing();
 
     void InvokeShortcut(const String& shortcut);
 
+    bool GetShorten();
```
JoshEngebretson: GetShorten/SetShorten are also a bit unclear, could use a doc string, and GetShorten can be const (just noting, I miss plenty of these too)

JoshEngebretson:
Overall, good work and definitely getting more used to the code conventions.  Let me know if you have questions/thoughts on the hover tooltip custom event

Source/Atomic/UI/UI.cpp
rsredsq: TBPopupWindow is designed to be attached to the `target` widget, and it's hard to deattach it from it an stick with another widget without recreating, so it's simplier to close the window and create a new one to the another widget(realocating doesn't consume a lot of resources and time, I've checked :smile: )
I can do it with realocating, but it needs much more TurboBadger code modification or creating a special widget for that(TBFloatingWindow?)

Source/Atomic/UI/UI.cpp
rsredsq: I'm not sure that we need event for that, because we want have only one tooltip, so no need to  track hovered time in widget, we should have only one global hovered widget(multitouch?) and track time for it globally

Source/Atomic/UI/UI.h
rsredsq: widgetHoveredTime?

Source/Atomic/UI/UIPopupWindow.h
rsredsq: ok :sweat_smile:

Source/Atomic/UI/UIWidget.h
rsredsq: `GetCapturing`/`SetCapturing` it's getters and setters for `needCapturing_` variable (https://github.com/AtomicGameEngine/AtomicGameEngine/blob/0fdfd80624754be1f1652cc804896a33b86cc7dc/Source/ThirdParty/TurboBadger/tb_widgets.h#L988)
`needCapturing` means if widget can be captured(it is related to the multitouch support, if we have captured widget, then nothing else can be pressed and only that event can recieve mouse events, e.g. in the DPad buttons don't need capturing, because I want have its pressed at the same time, all TB widget need capturing by default)
`GetCaptured` returns if widget is captured

Source/Atomic/UI/UIWidget.h
rsredsq: GetShorten returns true if widget has a string and that string has its length > widget rect, so that string has ...(three dots) at the end

Source/Atomic/UI/UI.cpp
JoshEngebretson: That is a great point, ok

Source/Atomic/UI/UI.cpp
JoshEngebretson: Ok, good to know, are we sure we want to use TBPopupWindow?  I guess I haven't looked at how this appears in the editor, @LaraEngebretson will kick off a local build and check it out :)

Source/Atomic/UI/UI.h
JoshEngebretson: Ok, weakptr is good here is we will be recreating

Source/Atomic/UI/UI.h
JoshEngebretson: tooltipHoverTime_ ?

Source/Atomic/UI/UIWidget.h
JoshEngebretson: Ok, we'll try and figure out what to call this later, and at some point maybe there will be a slightly way of handling it anyway, so in the future I'll flag method/var names which may cause some confusion, better :)

Source/Atomic/UI/UIWidget.h
JoshEngebretson: Ok, GetShortened() would be good, does this have to be on UIWidget?  Doesn't it only make sense for text fields?

rsredsq: All the widgets can contain a text, so it's better to have it for all of them

Source/Atomic/UI/UIWidget.cpp
```diff
@@ -810,5 +825,37 @@ void UIWidget::InvokeShortcut(const String& shortcut)
...
+    if (!widget_)
+        return;
+
+    widget_->SetShortened(shortened);
+}
+
+String UIWidget::GetTooltip()
+{
+    if (!widget_)
+        return "";
```
JoshEngebretson: return String::EMPTY;

Source/Atomic/UI/UIWidget.cpp
```diff
@@ -810,5 +825,37 @@ void UIWidget::InvokeShortcut(const String& shortcut)
...
+
+void UIWidget::GetShortened(bool shortened)
+{
+    if (!widget_)
+        return;
+
+    widget_->SetShortened(shortened);
+}
+
+String UIWidget::GetTooltip()
```
JoshEngebretson: const String& GetTooltip() const

Source/Atomic/UI/UIWidget.cpp
```diff
@@ -810,5 +825,37 @@ void UIWidget::InvokeShortcut(const String& shortcut)
...
+{
+    if (!widget_)
+        return "";
+
+    return widget_->GetTooltip().CStr();
+}
+
+void UIWidget::SetTooltip(const String& tooltip)
```
JoshEngebretson: nice :)

Source/Atomic/UI/UIPopupWindow.h
JoshEngebretson: Did you add UIPopupWindow here?

https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/AtomicJS/Javascript/JSUI.cpp#L84

JoshEngebretson: Why, yes you did, I just saw it, great!

Source/ThirdParty/TurboBadger/tb_widgets.h
```diff
@@ -999,6 +999,14 @@ class TBWidget : public TBTypedObject, public TBLinkOf<TBWidget>
 
     TBWidget *GetWidgetByTouchId(unsigned touchId);
 
+    bool GetShortened() { return shortened_; }
+
+    void SetShortened(bool shortened) { shortened_ = shortened; }
+
+    TBStr GetTooltip() { TBStr str; str.Set(tooltip_); return str; }
```
JoshEngebretson: Why do we need to do the str.Set(tooltip_) dance here?

Source/Atomic/UI/UIWidget.cpp
rsredsq: ok

Source/Atomic/UI/UIWidget.cpp
rsredsq: ok

Source/ThirdParty/TurboBadger/tb_widgets.h
rsredsq: hm, right, probably don't need that, const TBStr better

JoshEngebretson:
This is really great, LANDING and thanks :+1:

PR #487, opened 2015-12-07
WebRequest for native

Type1J:
Here's the initial round for PRs for WebRequest. It doesn't have Emscripten, yet, but that should come shortly after upload support, which may already work, but I haven't tested it as thoroughly as I want just yet.

Multiple examples are on the way as well.


Type1J:
I'm looking into the build problem on a Mac later today. I really like CI letting us know these things before a merge, so there's no need to revert.

JoshEngebretson:
Thanks Jay, this Monday is a bit busy.  I'll have a chance to look it over soon.  

The CI was a pain to setup (and I recently redid it from scratch as I learned a couple things on the initial stab), though absolutely necessary :+1:

JoshEngebretson:
One note, on OSX we should be able to continue building against the system libcurl?

Type1J:
If OSX guarantees that libcurl is present, then, yes. I added libcurl to all non-Emscripten platforms, and maybe that wasn't the right way to go.

The way that I made the WebRequest implementation was to make it "a hop, a skip, and a jump" away from being able to do HTTP 2.0.

The reason for compiling libcurl into the engine directly would be because we would eventually want to add libnghttp2, which libcurl optionally links against (we wouldn't use it directly), and which may or may not also be available in the newer OSX distributions.

JoshEngebretson:
I really don't know about libnghttp2, though if it can't be used with the system libcurl, or we would run into problems with system libcurl versions, I am all for compiling it in on Mac from source.  We currently use the system one primarily as it was expedient to work it in :)

Type1J:
I haven't had much time to look at this lately, but it's coming as soon as I can get to it.

Type1J:
Taking this back for now. It needs more support work, first. Support PRs will come soon.

PR #488, opened 2015-12-07
Exposed Texture Unit Names

mattbenic:
Found we needed to use this for exposing inspector dropdowns in C++ components (unless we externed the array in each cpp, but this seemed a better approach)


Source/Atomic/Graphics/Material.cpp
```diff
@@ -758,7 +758,7 @@ Scene* Material::GetScene() const
 
 String Material::GetTextureUnitName(TextureUnit unit)
```
JoshEngebretson: How about adding a static getter on Material instead of exposing the raw string array?  I can do this if there are time constraints

mattbenic: Yeah I can do that now :)

mattbenic:
Requested change made

Issue #495, opened 2015-12-10
Be able to specify source code language for new project and examples


shaddockh:
When specifying a new project, a dropdown of available languages for the project type should be available to select which project template to use.  This should also apply to the Examples so that if an example is written in multiple languages, it is possible to select which version to use.

This will then clone the appropriate project template/example into the path the user specifies.

shaddockh:
FYI, this is something I'm currently working on.

PR #499, opened 2015-12-12
Multi Language Templates for #495

shaddockh:
This provides a dropdown on the new project / clone example project dialog to specify which source code language to generate the project for.  Associated with #495.

Notable changes:
- Added a dropdown to the create project dialog that is fed from entries in the JSON configuration for the template
- created a JSON configuration file for the standard templates (2d, 3d, empty)
- updated the JSON configuration file for the examples to use the new structure 
- specified that the Roboman3d example can have a JavaScript and TypeScript target
- created TypeScript versions of the 3 standard project types
- modified the way that the templates get copied.  Now, instead of just the resources folder getting copied, the entire folder gets copied and a cache folder generated.  The .atomic file then gets renamed
- More languages can easily be added by just updating the template configuration in the JSON (ie, C#, C++, etc)
- The TypeScript example projects store the .ts files alongside the .js files instead of storing them in a separate source directory.  When the TypeScript gets compiled, it will just overwrite the .js file of the same name.

Questions:
There are a few things that may need to be discussed
1) the *.d.ts files for the typescript examples are just stored in the template.  Would it be better to auto-update them on editor build?
2) the editor currently doesn't view .ts files, just .js files.  When the editor pulls in the Chromium framework, would it be possible to display different languages or should the editor go ahead and allow display of .ts files as is?
3) Once number 2 is resolved, that opens up the ability to have the New Component/New Script menu item generate a stub file in the appropriate language.
4) Is there a format for the .atomic project file?  I would like to store the source code language for the project in that file so other elements of the editor can get at it.


JoshEngebretson:
Whoa, heck of a way to wake up, I will build this immediately after morning coffee requirements are met :)

1)  Yes, I think this should be copied or something on editor build instead of duplicated in repo.  All of the definition files, other than the AtomicWork.d.ts (which is currently just a spot to hand add stuff), are auto generated by the native code binding system, so tied to that I'd think.  We have local development builds for working on Atomic itself (uses the repo file tree), and then packaged editor builds

2)  We can add a .ts import type to the editor which opens a basic text editor to start (this works, without js syntax color).  I was playing around with working on the WebView today, and this PR just solidified that plan :)

3) Yes, and so awesome! :+1: 

4)  There is, though it is kind of "under used", it currently only strored the platforms used and is only populated on command line builds using the CLI stuff (which now needs dusting off) https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/ToolCore/Project/ProjectFile.cpp

We should also figure the local node into editor builds, so the editor can invoke the tsc compiler and report compilation errors, this can most definitely be a follow up

JoshEngebretson:
This looks great, I am going to land it, update to TS 1.7 and node 5.2, do some testing and address it in a follow up PR if anything is wonky.  THANKS! :+1:

PR #509, opened 2015-12-13
Splits up AtomicEngine menu to conform with standard program menus

Sleaker:
Moves Quit to File menu.
all other options to the Help menu.
removes AtomicEngine from the menubar.
This format mirrors major editors (Unity, VS, etc)

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -21,6 +21,8 @@
 
 - Matt Benic (https://github.com/mattbenic)
 
+- Nick Minkler (https://github.com/Sleaker)
+
 ### Contribution Copyright and Licensing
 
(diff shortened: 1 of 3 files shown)


JoshEngebretson:
What needs to happen here is a top level system menu on Windows, OSX, Linux, this is a little involved as we need to be able to define native OS accelerators, submenus, etc.

On OSX, it is standard to have the first menu item be the program's name with quit and other services there, this included Atom, Chrome, Unity, etc:

screen shot 2015-12-13 at 7 17 43 am

I do think this is an improvement for now as Windows is the dominate platform and this layout is more what these users would expect.  Can you add yourself to the contributors list in the PR?

https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/AUTHORS.md

@AtomicBuildBot ok to test

Sleaker:
@JoshEngebretson - yah, that might be involved. I think the only difference would actually be OSX, Linux/Windows seem to handle it the same.  Also I believe someone already posted up that OSX has other menu problems.

JoshEngebretson:
Looks great, thanks.  LANDING!

Issue #455, opened 2015-11-22
Closing files sometimes closes the project


LaraEngebretson:
* It happens after making changes in the inspector
* The file doesn't show any unsaved changes
* Once it happens it affects all files in all projects until the editor is restarted

JoshEngebretson:
@rsredsq Could this be related to the modified check when closing resource editors?

rsredsq:
@LaraEngebretson could you please give more information? I can't really reproduce it

LaraEngebretson:
I haven't pinpointed the exact steps to replicate this. I have noticed it happen on multiple occasions though and it's always after I have been making a lot of changes in the Inspector panel.

LaraEngebretson:
I was able to replicate this behavior by doing the following: 

* Open ParticleEmitter3D example
* Open the Scene file
* Close the project
* Open Character Animation 3D example
* Open the Scene file
* Select Robo_01 in the Hierarchy
* Expand the AnimationController component
* Select Animation Resource (click ... next to Animation 0)
* Hit Cancel
* Close the scene file

blueFireGames:
It also happens if no changes are made to the project.

Jason

PR #515, opened 2015-12-13
Fix when editor was closed unexpectedly

rsredsq:
#455


JoshEngebretson:
Thanks, we'll do some testing before pushing... LANDING!

Issue #512, opened 2015-12-13
Dialog to save unsaved changes for a scene shows twice


Sleaker:
1. Open a scene in the editor (I'm using the base 3d example.)
2. make a change.
3. click the X to close the scene window.
4. 2 Save dialogs will pop to save the scene.

Expected: Only 1 dialog should prompt the user to save the scene.

blueFireGames:
This also happens when modifying scripts and components as well.

Jason

PR #516, opened 2015-12-14
UI pointer events clean up

rsredsq:
I've done small clean up for the ui pointer event(again), it's related a bit to the #512, because the click/up event was sent twice

I also fixes a scroll(it was fixed before, but it's more clearly way to do that, because UIListView should be capturing widget)

### DIFF
--- a/Source/Atomic/UI/UISelectList.h
+++ b/Source/Atomic/UI/UISelectList.h
@@ -77,8 +77,6 @@ protected:
     virtual bool OnEvent(const tb::TBWidgetEvent &ev);
 
 private:
-    //Returns true if we have drag and drop element on select list widget
-    bool handlingDragAndDrop_;
 };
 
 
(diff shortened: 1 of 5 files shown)


Source/Atomic/UI/UISelectList.h
```diff
@@ -77,7 +77,8 @@ class UISelectList : public UIWidget
     virtual bool OnEvent(const tb::TBWidgetEvent &ev);
 
 private:
-
+    //Returns true if we have drag and drop element on select list widget
+    bool handlingDragAndDrop_;
```
JoshEngebretson: Can we make sure to initialize this member variable in  the constructor, even if it isn't strictly necessary?

rsredsq:
@AtomicBuildBot retest please

rsredsq:
@AtomicBuildBot test this please

PR #517, opened 2015-12-14
Menubar handles closing itself when it was clicked twice

rsredsq:


rsredsq:
I don't know what the heck is happening with code formatting in turbo badger files, I will fill an issue to convert all tabs to spaces

Source/ThirdParty/TurboBadger/tb_widgets.cpp
```diff
@@ -1355,6 +1355,7 @@ void TBWidget::InvokePointerUp(int x, int y, MODIFIER_KEYS modifierkeys, bool to
 {
     int ox = x;
     int oy = y;
+    TBWidget *old_capture;
```
JoshEngebretson: really should initialize pointers, other than maybe in tiny amounts of code where you can guarantee they are set, I am starting to get suspicious with the amount of changes that are seeming to be necessary here :)

rsredsq: It's the last change, I promise! :mag:

Script/AtomicEditor/ui/frames/MainFrame.ts
```diff
@@ -111,12 +111,18 @@ class MainFrame extends ScriptWidget {
         if (this.menu.handlePopupMenu(target, refid))
             return true;
 
+        if (this.openedMenu && this.openedMenu.id == target.id + " popup") {
+            this.openedMenu.close();
+            this.openedMenu = null;
+            return true;
+        }
+
```
JoshEngebretson: What happens when the menu is closed normally, by selecting something in it or clicking off it?  Does the openMenu member still hold a reference to the last UIMenuWindow?  It seems like it would

rsredsq: close() removes the native pointer, so openedMenu should be GCed,
don't really need set it to null

JoshEngebretson: If script holds onto a reference, it will not be GC'd, how could it be?

Source/ThirdParty/TurboBadger/tb_widgets.cpp
JoshEngebretson: Ok, let's document a little of what is going on though, for ox, oy, and old_capture variables, thanks!

Script/AtomicEditor/ui/frames/MainFrame.ts
JoshEngebretson: I mean the backing TB widget may be released like this, however the UIMenuWindow wrapper will still be held by the openedMenu member, won't be GC'd, and will potentially have a null widget.  I don't think nulling is strictly necessary, but it is something to consider.

rsredsq: hm, do you think that it's necessary to release the openedMenu in that case? It will be overwritten on the next menu click, so it wouldn't cause any mistakes

JoshEngebretson:
Did this get addressed somewhere?

rsredsq:
I'll remake that PR later, it shouldn't contains fixes for ui

JoshEngebretson:
Ok, going to try this in native code right?  

It would be nice if, for example, when clicking the create button on the project frame, and then clicking it again, the menu would go away (without needing special code for Menu bar, create button, create component button, etc).  

Do some research and see what you think?

PR #524, opened 2015-12-15
Refactored pointer up/down function

rsredsq:


Source/Atomic/UI/UISelectList.cpp
```diff
@@ -238,6 +238,13 @@ bool UISelectList::OnEvent(const tb::TBWidgetEvent &ev)
     {
         GetTBSelectList()->SetFocus(WIDGET_FOCUS_REASON_POINTER);
     }
+    if (ev.type == EVENT_TYPE_POINTER_MOVE)
+    {
+        UIDragDrop* dragDrop = GetSubsystem<UIDragDrop>();
+        //if return true, then scroll event will be controlled by that widget itself
+        if (dragDrop->GetDraggingObject())
+            return true;
```
JoshEngebretson: This is really nice clean up

LaraEngebretson:
I will test this first thing tomorrow morning. Thanks @rsredsq :)

LaraEngebretson:
I found a couple things on Windows and OSX:
* When I click on the position/rotation/scale buttons in the inspector, the .scene doesn't recognize that changes have been made.
* I am experiencing inconsistencies with being able to move nodes around in the hierarchy panel. I haven't been able to replicate a specific set of steps that causes it to stop working.

JoshEngebretson:
On the scene not being marked as modified, this is likely something to do with the modified event not being set, which could be an indicator of other problems:

https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/ThirdParty/TurboBadger/tb_inline_select.cpp#L125

The ListView does have some problems, it sometimes scrolls the view while dragging a node (it seemed to be when the node was selected, and dragged through it, which is what the setCapturing(false) was in there for), sometimes dragging a node doesn't work at all as  the drag is not initiated, and sometimes it seems to work.

rsredsq:
1). Why do we have 'if (!ev.target->IsCaptured())'? TBInlineWidget will be captured, when it gets an event. I'll check it later. 
2). Hm, when I've been checking scroll, I didn't find any problems. Ok, I'll retest. Probably need a better scroll prevent, setCapturing(false) doesn't suits in that case

JoshEngebretson:
1)  Because you don't get the modified event until after you release the mouse button.  While holding the button, auto_click events are sent...

2)  This is specifically when dragging nodes, it is now pretty flaky as described above

LaraEngebretson:
Dragging/dropping nodes in the hierarchy panel is still broken. The list scrolls while moving nodes. I was testing in ToonTown as it has a lot of nodes.

dragdrop

JoshEngebretson:
When a node is selected (dark grey horizontal line), beginning a drag and drop sometimes works and sometimes doesn't

These behaviors are quite easy to reproduce

JoshEngebretson:
FYI: I am looking into the drag and drop problem, though the scroll issue is still a problem

JoshEngebretson:
Potential fix for drag drop and improper scrolling behavior here: 

https://github.com/AtomicGameEngine/AtomicGameEngine/tree/JME-ATOMIC-LISTVIEW

This includes the updates from the RED-UI-FIX

JoshEngebretson:
Closing this PR as the UI commits are also contained in this new PR #538, which also addresses some drag/drop issues.  Thanks @rsredsq for UI pointer event fixes

PR #544, opened 2015-12-23
Adding TSLint process to the build

shaddockh:
This adds TSLint to the build process.  I was sitting on these changes until the Linux PR merged in so I could take those into account.  

There are some significant changes to the node modules on here though, so take note...some of them may need to be thought about and if they are too invasive it may just be that TSLint doesn't need to be part of the editor build.  A less invasive option would be just to store a tslint.json file in the root and have the Atom editor configured to read it during development, but not throw errors on build.

**Changes**
- in node 5, the node modules are flattened out to the top level, so now individual modules no longer contain deep hierarchies of duplicate nodes under nested node_modules directories.  This ultimately saves space, but makes your root node_modules much larger.  Pulling in the tslint and typescript modules added to the root of node_modules.  Ideally, `jake` and `fs-extra` should also be updated to do the same.
- Although you don't want to require npm to build, it might be something to contemplate to reduce the source control size.  There is a way to lock in particular versions of a package via `npm shrinkwrap` and that could be an option to lock down dependencies to the version that you know works ( https://docs.npmjs.com/cli/shrinkwrap )
- TSLint has a dependency on TypeScript and requires that it be located in the node_modules directory so it can be resolved, so now there is a TypeScript folder under node_modules  as well as in the build directory.  Ideally, the one in node_modules would stay and the other one would be removed.  `BuildCommon.js` has been updated to use the tsc in node_modules.
- Added a `tslint ignore` option to the generated .d.ts files since they have spacing and content that TSLint doesn't like.  These *should* be addressed in the future.  The spacing shouldn't be difficult to fix in the generator, but there are a couple instances where the same variable names are used and that will take some analysis. `HTTP-ERROR` `HTTP-INITIALIZING` `HTTP-OPEN` and `HTTP-CLOSED` are duplicates.

Let me know if this is the direction you would like to proceed, or if you would rather a more minimalistic approach.

-Shad


JoshEngebretson:
Wow, meaty with a nice PR write up!  

Hm, what do you think about including the npm binary on Windows/OSX and have that pull down the necessary node modules at bootstrap time?  That way we can pull down packages without needing npm to be installed on the system. I would need to look into it a bit more, though that could be doable.  

Another approach would be to have a submodule with the node stuff, that way it wouldn't be in the main repo and we could reboot it if it ever got big.  We're going to have submodules for the WebView and CoreCLR support...

For now, just having the node module sources in the repo isn't a huge concern from a size standpoint, they are just JS text files which compress quite well.  Though, I guess with these three options I would lean towards including npm binaries in  the repo, that could be pretty awesome as the editor could then pull down packages... :)

shaddockh:
Actually, bundling the platform specific version of `npm` with the build might be a good way to go.  I'm not sure what all dependencies it requires, but by doing that it does solve an issue of using node modules that may have platform specific pieces.  By having the build execute an npm install of the modules, the correct platform extensions would come down correctly.

If you want to go with the bundled npm route, then using `npm shrinkwrap` will definitely be something that needs to be done so that incompatible modules don't get pulled down ... only modules versions that have been tested will be.  It might be a good thing to do even if npm isn't bundled, just to keep things sane.

What I can do is update this PR by configuring npm shrinkwrap, and making sure that jake and fs-extra are formatted in the node 5 way with flat dependencies.  Once that's done the stage should be set to be able to bundle NPM if that's the way you want to go.

JoshEngebretson:
I like the flattened modules, a lot... looking over all the npm dependencies and comparing them to what we have now:

https://github.com/shaddockh/AtomicGameEngine/tree/TSH-ATOMIC-EDITOR-TSLINT/Build/node_modules

Our node_modules is svelte is comparison, I think we'll eventually hit a tipping point where having npm in the repo or as a bootstrap download will make sense.  Though, at that point, it might be better to just have node/npm as a prerequisite.

... still chewing on just embracing npm as a prerequisite

JoshEngebretson:
I've been having a closer look here, great work :)

A question, "the node modules are flattened out to the top level, so now individual modules no longer contain deep hierarchies", I am wondering about this as there are still nested node_modules, maybe this only happens when they aren't shared between modules?

I think we should punt on including/requiring npm at the moment, once you start digging through the nested node_modules in the node/npm distro, it really quite a bit of machinery.  I think we should revisit this down the road a bit, especially with how it might play with the editor.  The editor would definitely come with a bundled npm, as we really can't require it in this case.

I attempted to build the PR using Build_AtomicEditor.bat on Windows  and received:

    jake aborted.
    Error: ENOENT: no such file or directory, stat 'C:\Dev\shaddock\AtomicGameEngine\Artifacts\Build\Resources\EditorData\AtomicEditor\EditorScripts    '
    at Error (native)
    at Object.fs.statSync (fs.js:893:18)

This was from a clean checkout.  I can verify that master doesn't have the problem.  The build box doesn't currently remove artifacts (it should), so that folder may have already been there during CI.

I do like the idea of the TSLint process as part of the build.  It should help catch stuff before code review, which is always good :+1:

JoshEngebretson:
Just a note that master did build from a clean checkout on Windows, the CI scripts should remove everything in Artifacts before building... which they will soon #548

shaddockh:
Ok, I'll take a look tonight and see what's going on.  Does the PR build on OSX?

shaddockh:
I did a little research, and the nested node_modules look like they are only when there are version conflicts between dependent modules.

from: https://github.com/npm/npm/issues/9809
> each dependency with a conflicting version on the same dependency would get its own, nested copy of that conflicting dependency. That's why we describe the install tree as maximally flat – it will still allow nesting in the case of conflicts.

Waiting on npm sounds like a good plan....  I agree that it needs more thought and really, until there is a nice plugin api that can take advantage of it, there may not be as much need yet other than trying to expose typescript compilation services to the editor.

JoshEngebretson:
The PR built fine on OSX, I was looking during the build for evidence of linting and didn't see it.  That means it is working right? ;)

shaddockh:
It should only emit errors if there are linting issues and then it will cause jake to abort.  I was getting those on the generation of the *.d.ts files which is why I added the tslint disable to the top.  I'll re-verify that it's also happening on the editor *.ts files too.

shaddockh:
It looks like the glob of `./script/**/*.ts` was not picking up editor files.  The latest commits should resolve that by linting the editor files in a separate pass than the d.ts files.

shaddockh:
These last commits do the following:
- scan the entire AtomicEditor tree for .ts files and run TSLint on each of them.  This is a stopgap measure until TSLint supports passing a directory to it and recursing through it.  The one main issue with this is that the first file that has lint issues will fail the build instead of running through all the files.
- changed the way TSC and TSLint are called in the build.  It was calling the script in the bin directory, but I'm guessing that under windows that won't work right, so it's calling directly into the lib directory.  Let me know if this resolves the issue with the windows build.
- fixes a bunch of small formatting issues in the editor code

It seems that linting the *.d.ts files is frowned upon, so I've already removed the Atomic.d.ts checks in the build process.  Should I also remove the addition of the /* tslint:disable */ line from being generated as well?

JoshEngebretson:
I think this is a win looking over the stuff it caught in the editor scripts, so let's definitely push forward with it!

Maybe the editor scripts aren't laid out like a typical TS project?  I am a little confused why TSLint would be having trouble linting here, do we need to refactor to be more TS "compliant"?

"Should I also remove the addition of the / tslint:disable */ line from being generated as well?"  -  I am unsure and will trust your taste on this one :)  I like the idea of the ambient definitions being linted as well, where do the frowns come in with it?

I cleaned the Windows workspace on the build box, let's see what happens:

shaddockh:
> Maybe the editor scripts aren't laid out like a typical TS project? I am a little confused why TSLint would be having trouble linting here, do we need to refactor to be more TS "compliant"?

I don't think it's a matter of the project not being TS compliant, rather it's the maturity level of TSLint.   Most people either use it in their editor on a single file at a time or pipe it through gulp during a build process a single file at a time.  Once TSLint supports nested directories, we can remove the workaround in the build, but the build will currently work with whatever nesting is in place. 

For the ambient definition files, people have just indicated that they don't want changes in a non-controlled .d.ts file to break their build.  Since they are generated by the build, that doesn't really apply here.  My thought was to remove the disable tslint headers and exclude them from linting in the build until they are compliant. (there are a couple problems with these files currently) With the `disable` lines in place, even if you did try to lint, it wouldn't catch anything.  With them gone, we have the option of linting them in the future if desired.

JoshEngebretson:
This needs to be rebased with some of the latest master changes, are there any other changes you're planning on applying?

shaddockh:
Nope, this should be good to go at this point.

JoshEngebretson:
There are some merge conflicts with master, spare a rebase when you have a chance? :)

shaddockh:
Sure thing, I'll look at that tonight.

shaddockh:
If you get a chance can you take a look at the Linux build?  It appears that it is failing on generating C# bindings, but all the other platforms are working.

JoshEngebretson:
Thanks, I wiped the workspace on the build box, let's give this a whirl

JoshEngebretson:
I have a feeling this has something to do with the version of node on the build box, which is the one from the standard Ubuntu package (not sure the exact version atm).  I will try tracing it tomorrow, I am, pretty sure the bindings are being generated just fine

JoshEngebretson:
Ok, pretty simple case issue on typescript module name here:

https://github.com/shaddockh/AtomicGameEngine/blob/TSH-ATOMIC-EDITOR-TSLINT/Build/Scripts/BuildCommon.js#L19

should be "typescript" not "typeScript"

JoshEngebretson:
OMG!  Getting closer :)  Lint issue I think:

    Script/AtomicEditor/editor/Editor.ts[118, 120]: missing semicolon

Why only on Linux, no clue!

shaddockh:
Ok!!! I think that does it.

JoshEngebretson:
Excellent, and the mystery of why Linux is flagging the lint issue is that on the CI box, the Linux validation is using the "Build_AtomicEditor.sh" which uses the local build scripts, the CI build scripts are used for example on the mac validate task:

./Build/Mac/node/node ./Build/node_modules/jake/bin/cli.js -f ./Build/CIScripts/Build.js build:mac_editor

We need to marry the local and CI builds, this was a bit too much to handle on the first pass.

An issue I would raise here, now that we have it passing (smile), is that when working in Atom/VSCode/VS and in the editor itself, it becomes important that the linting is consistent across user IDE, local builds, and CI builds. If not, and we're generating single lint errors per PR adjustment this could become a real pain.

I am all for the linting and think this PR adds really nice layout changes to node modules, so I would like to get it in.   

Maybe we can not fail the build on a lint issue and make them warnings instead?  I think warnings could be a good way to start with the linting, what do ya think?

shaddockh:
Sure, I think that's a good idea.  I was also concerned that it fails on the first lint error instead of running through and displaying all lint errors.  I'll go ahead and update it so that it bubbles up warnings instead of failing and also displays all linting issues.

shaddockh:
Ok, this should be ready now.  The linter now runs through all the TS files and outputs any errors it finds to the console. Also, it's configurable to whether it should break the build or not on error.  Currently it's set to **not** break the build on errors.

JoshEngebretson:
Awesome, looks great... aaaaand LANDING! :airplane:

Issue #511, opened 2015-12-13
Windows editor should start maximized if exited in this state


LaraEngebretson:
Currently when exiting the editor in a maximized state, the window is roughly the correct size however it is not maximized and slightly smaller than it should be.

PR #545, opened 2015-12-23
Windows editor start maximized if exited in this state

rsredsq:
#511

### DIFF
--- a/Script/AtomicEditor/editor/Editor.ts
+++ b/Script/AtomicEditor/editor/Editor.ts
@@ -54,12 +54,17 @@ class Editor extends Atomic.ScriptObject {
         });
 
         this.subscribeToEvent("IPCPlayerWindowChanged", (data) => {
+            var playerWindow = Preferences.getInstance().playerWindow;
             //if player window is maximized, then we want keep the window size from the previous state
             if (data.maximized) {
-                Preferences.getInstance().playerWindow.maximized = true;
-                return;
+                playerWindow.x = data.posX;
+                playerWindow.y = data.posY;
+                playerWindow.monitor = data.monitor;
+                playerWindow.maximized = true;
+            } else {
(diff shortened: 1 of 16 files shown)


JoshEngebretson:
Mr. Squirrel, this is looking great and will be a nice improvement.  I'll have a closer look once we're back in the office.  Happy holidays! :christmas_tree:

Script/AtomicEditor/editor/Editor.ts
```diff
@@ -170,7 +200,7 @@ class Editor extends Atomic.ScriptObject {
     }
 
     exit() {
-        Preferences.getInstance().write();
+        //Preferences.getInstance().write();
```
JoshEngebretson: Dead code?

JoshEngebretson:
Ok, I looked this over and other than one note it looks really good.  It just needs some testing now.

Source/Atomic/Engine/Engine.cpp
```diff
@@ -379,6 +379,9 @@ bool Engine::Initialize(const VariantMap& parameters)
         ))
             return false;
 
+        if (GetParameter(parameters, "WindowMaximized", false) == true)
+            graphics->Maximize();
```
JoshEngebretson: This should be factored into the SetMode call immediately above:

https://github.com/rsredsq/AtomicGameEngine/blob/RED-EDITOR-511/Source/Atomic/Engine/Engine.cpp#L370

That way the window isn't created and then immediately resized for maximize

LaraEngebretson:
This issue occurs on OSX only:

* Move the editor to 2nd monitor
* Double-click the title bar to maximize the editor window
* Move editor back to primary monitor
* Close the editor
* Relaunch the editor

Issue: the editor reopens on the 2nd monitor

Source/Atomic/Engine/Engine.cpp
rsredsq: ok

rsredsq:
@LaraEngebretson thanks! good catch ;)

Issue #529, opened 2015-12-17
Add a check to the example browser to verify that the templates exist


LaraEngebretson:
If you build from source and do not have a local copy of the examples repo, the following error occurs when loading a template:

![missingtemplates](https://cloud.githubusercontent.com/assets/722928/11875803/49bbb172-a49b-11e5-906e-47cfd6789298.PNG)

PR #546, opened 2015-12-24
Fixes for #529

shaddockh:
* added a check in the welcome screen to verify that an example project exists on the file system before adding it to the example browser
* added a check to the CreateProject dialog to verify that the template to clone from exists before trying to clone it

### DIFF
--- a/Script/AtomicEditor/ui/frames/WelcomeFrame.ts
+++ b/Script/AtomicEditor/ui/frames/WelcomeFrame.ts
@@ -50,6 +50,20 @@ class WelcomeFrame extends ScriptWidget {
 
     addExample(example: ProjectTemplates.ProjectTemplateDefinition) {
 
+        var fileSystem = Atomic.getFileSystem();
+
+        // Verify that at least one of the projects for this example exists, otherwise bounce out
+        let exists = false;
+        example.templates.forEach(template => {
+            if (fileSystem.dirExists(template.folder)) {
+                exists = true;
+            }
+        });
+
+        if (!exists) {
+            return;
+        }
+
         var exlayout = <Atomic.UILayout>this.getWidget("examples_layout");
 
         if (!this.currentExampleLayout) {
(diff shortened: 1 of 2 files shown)


Script/AtomicEditor/ui/frames/WelcomeFrame.ts
```diff
@@ -50,6 +50,20 @@ class WelcomeFrame extends ScriptWidget {
 
     addExample(example: ProjectTemplates.ProjectTemplateDefinition) {
 
+        var fileSystem = Atomic.getFileSystem();
```
JoshEngebretson: This doesn't need to be changed for this PR, though for reference there is a shortcut to the fileSystem on the Atomic object: https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Script/TypeScript/AtomicWork.d.ts#L20

On another note, these should be a getters, or some other fancy TS construct :)

Script/AtomicEditor/ui/modal/CreateProject.ts
```diff
@@ -127,7 +127,15 @@ class CreateProject extends ModalWindow {
...
             } else {
-                var message = "Unable to create project for language:  " + selectedLanguage + "\n\nPlease choose a different language";
+                let message = [
+                    "Unable to create project for:",
+                    "",
+                    `language: ${selectedLanguage}`,
+                    `template: ${templateDetail.folder}`,
+                    "",
+                    "Please choose a different language."
+                ].join("\n");
```
JoshEngebretson: Nice, I am going to need to level up my TS string handling skills :)

JoshEngebretson:
Thanks, this should help when someone builds from source and then is puzzled why the examples don't work :)  LANDING!

Script/AtomicEditor/ui/modal/CreateProject.ts
shaddockh: yeah, I'm still getting into practice using these.  Honestly, that should really be `const message ...` since it's not going to be mutated.  There is also a TSLint rule that can be turned on that bans the use of the `var` statement.  I believe ESLint also has a rule that checks `const` vs `let` usage, but that's not available in TSLint from what I can tell.

PR #547, opened 2015-12-29
WebRequest for native

Type1J:
This WebRequest class uses libcurl to make http/https requests.

It has planned support for Emscripten, which is not included, but it's partially done (I needed to make some of it to make sure that the API will match properly with the native version).

To make it work across iOS, Android, OSX, and Windows, I had to pull zlib out of Assimp and get gzip functionality compiling. Assimp and libcurl now use a common zlib static library.

Also, some of libcurl's CMake build system was added. I plan to make iOS, OSX, and Windows use their own TLS implementations eventually *(Update: This is now done, see commits and comments below.)*, but for now, they all use OpenSSL, which would be needed for new platforms, and is useful to have around for hashs and signatures.

There is an AtomicExamples pull request for example uses.


Type1J:
Hold off on the merge. I just realized that this may affect license verifications because that uses https. I'll get the non-OpenSSL versions working for Windows and OSX so that this will work without having to bypass CA verification.

JoshEngebretson:
This and the example are looking great!  I'll hold off merging + some code review until you can look into the https on Windows/OSX (we can probably bypass CA verification on Linux using the option in the short term?)

I just updated the build box to VS2015 Update 1, so will have the build bot test it again, they updated the frontend to Clang and did move some stuff around so might need some tweaks.

Type1J:
It should be good, now. SChannel and Secure Transport are used in place of OpenSSL on Windows, OSX, and iOS.

JoshEngebretson:
This looks really solid and is a fantastic addition, doing some testing with the activation server and then will land.  THANKS!

JoshEngebretson:
Checks out... aaaaaand LANDING!

Issue #555, opened 2016-01-02
Linux compilation on Manjaro


nifker:
Im getting this error: 
In file included from /home/user/Desktop/AtomicGameEngine/Source/ThirdParty/TurboBadger/tb_renderer.h:12:0,
                 from /home/user/Desktop/AtomicGameEngine/Source/ThirdParty/TurboBadger/tb_renderer.cpp:6:
/home/user/Desktop/AtomicGameEngine/Source/ThirdParty/TurboBadger/tb_linklist.h: In Konstruktor »tb::TBLink::TBLink()«:
/home/user/Desktop/AtomicGameEngine/Source/ThirdParty/TurboBadger/tb_linklist.h:72:21: Fehler: »nullptr« wurde in diesem Gültigkeitsbereich nicht definiert
     TBLink() : prev(nullptr), next(nullptr), linklist(nullptr) {}
                     ^
/home/user/Desktop/AtomicGameEngine/Source/ThirdParty/TurboBadger/tb_linklist.h: In Konstruktor »tb::TBLinkList::TBLinkList()«:
/home/user/Desktop/AtomicGameEngine/Source/ThirdParty/TurboBadger/tb_linklist.h:96:26: Fehler: »nullptr« wurde in diesem Gültigkeitsbereich nicht definiert
     TBLinkList() : first(nullptr), last(nullptr), first_iterator(nullptr) {}
                          ^
Source/ThirdParty/TurboBadger/CMakeFiles/TurboBadger.dir/build.make:86: die Regel für Ziel „Source/ThirdParty/TurboBadger/CMakeFiles/TurboBadger.dir/tb_renderer.cpp.o“ scheiterte
make[2]: *** [Source/ThirdParty/TurboBadger/CMakeFiles/TurboBadger.dir/tb_renderer.cpp.o] Fehler 1
CMakeFiles/Makefile2:608: die Regel für Ziel „Source/ThirdParty/TurboBadger/CMakeFiles/TurboBadger.dir/all“ scheiterte
make[1]: *** [Source/ThirdParty/TurboBadger/CMakeFiles/TurboBadger.dir/all] Fehler 2
Makefile:83: die Regel für Ziel „all“ scheiterte
make: *** [all] Fehler 2

Type1J:
Can you tell us what compiler you are using, and which version?

It looks like this is a compiler that doesn't know the C++11 nullptr keyword.

nifker:
GCC 5.3

[...]

PR #556, opened 2016-01-03
Make CMake in Linux work correctly.

Type1J:
I made a bare Ubuntu VM to see if I could compile Atomic, and I ran into quite a few pitfalls. The first few of those are documented in AtomicGameEngine/AtomicGameEngine#555 . The last few were related to node and libcurl not being present on my bare Ubuntu machine, although they are (or should be) available in the Atomic repo itself. I added node, and linked the correct libcurl. There were also some linker errors due to NETCore and NETScript not being linked in Linux. We could (probably should) automatically set the CMAKE variable LINUX to TRUE in cache if the CMAKE_SYSTEM_NAME is "Linux" instead of requiring it to be set on the cmake command line.

All of that is fixed with this PR, so hopefully this will help the Linux build along.

### DIFF
--- a/Build/CMake/Modules/AtomicLinux.cmake
+++ b/Build/CMake/Modules/AtomicLinux.cmake
@@ -1,5 +1,5 @@
 set (JAVASCRIPT_BINDINGS_PLATFORM "LINUX")
-set (ATOMIC_NODE_JAKE node Build/node_modules/jake/bin/cli.js  -f  Build/Scripts/Bootstrap.js)
+set (ATOMIC_NODE_JAKE Build/Linux/node/node Build/node_modules/jake/bin/cli.js  -f  Build/Scripts/Bootstrap.js)
 
 include(AtomicDesktop)
 
--- a/Build/CMake/Modules/AtomicUtils.cmake
+++ b/Build/CMake/Modules/AtomicUtils.cmake
@@ -26,6 +26,6 @@ macro(GroupSources curdir)
 
 endmacro()
 
-if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
+if(NOT CMAKE_CROSSCOMPILING AND ${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
   set(LINUX TRUE CACHE BOOL "Indicates if host is Linux.")
 endif()
--- a/Build/CMake/Modules/AtomicUtils.cmake
+++ b/Build/CMake/Modules/AtomicUtils.cmake
@@ -25,3 +25,7 @@ macro(GroupSources curdir)
     endif()
 
 endmacro()
+
+if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
+  set(LINUX TRUE CACHE BOOL "Indicates if host is Linux.")
+endif()
--- a/Build/Scripts/BuildCommon.js
+++ b/Build/Scripts/BuildCommon.js
@@ -26,7 +26,7 @@ namespace('build', function() {
         else if (os.platform() == "darwin")
           cmds.push(atomicRoot + "Build/Mac/node/node " + atomicRoot + "Build/TypeScript/tsc.js -p " + atomicRoot + "Script");
         else if (os.platform() == "linux") {
-          cmds.push("node " + atomicRoot + "Build/TypeScript/tsc.js -p " + atomicRoot + "Script");
+          cmds.push(atomicRoot + "Build/Linux/node/node " + atomicRoot + "Build/TypeScript/tsc.js -p " + atomicRoot + "Script");
         }
         jake.exec(cmds, function() {
 
--- a/Build/Scripts/Host.js
+++ b/Build/Scripts/Host.js
@@ -6,8 +6,7 @@ if (os.platform() == "win32") {
 } else if (os.platform() == "darwin") {
   module.exports = require("./HostMac");
   require("./BuildMac");
-}
-else if (os.platform() == "linux") {
+} else if (os.platform() == "linux") {
   module.exports = require("./HostLinux");
   require("./BuildLinux");
 }
--- a/Source/AtomicEditor/CMakeLists.txt
+++ b/Source/AtomicEditor/CMakeLists.txt
@@ -25,9 +25,8 @@ if (APPLE)
     set(ATOMIC_EDITOR_ICON ${CMAKE_SOURCE_DIR}/Build/CMake/Modules/Atomic.icns)
     set_source_files_properties(${ATOMIC_EDITOR_ICON} PROPERTIES MACOSX_PACKAGE_LOCATION "Resources")
 
-elseif(LINUX)
-
 else()
+
     include_directories (${CMAKE_SOURCE_DIR}/Source/ThirdParty/libcurl/include)
     add_definitions(-DCURL_STATICLIB)
 
@@ -36,6 +35,7 @@ else()
     add_definitions(-DATOMIC_WIN32_CONSOLE)
 
     set (SOURCE_FILES ${SOURCE_FILES} ${CMAKE_SOURCE_DIR}/Build/CMake/Modules/Atomic.rc)
+
 endif(APPLE)
 
 add_executable(AtomicEditor ${EXE_TYPE} ${SOURCE_FILES} ${ATOMIC_EDITOR_ICON})
@@ -45,10 +45,10 @@ target_link_libraries(AtomicEditor ToolCore AtomicJS AtomicPlayerJS AtomicNETJS 
 if (APPLE)
     set (TARGET_PROPERTIES MACOSX_BUNDLE_INFO_PLIST MacOSXBundleInfo.plist.template)
 
-    target_link_libraries(AtomicEditor NETCore NETScript curl)
+    target_link_libraries(AtomicEditor NETCore NETScript libcurl)
 
 elseif(LINUX)
-    target_link_libraries(AtomicEditor NETCore NETScript curl nativefiledialog ${GTK3_LIBRARIES})
+    target_link_libraries(AtomicEditor NETCore NETScript libcurl nativefiledialog ${GTK3_LIBRARIES})
 
 else()
     target_link_libraries(AtomicEditor NETCore NETScript libcurl Iphlpapi Wldap32)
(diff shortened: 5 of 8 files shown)


JoshEngebretson:
Thanks, the help on keeping Linux rolling is greatly appreciated.  I will take this PR opportunity to get CI builds back up for Linux.  For reference, the build box is 64 bit Ubuntu 14.04 LTS 

I did try and build using ./Build_AtomicEditor.sh, though the root script is looking to use the system node.  It also looks like maybe the node binary checked in is 32 bit as it won't run on the build box ( could potentially install 32 bit libs to see if it works, though would rather avoid see below).  

I don't think we need to check in the node binary for Linux, as the platform has an excellent package management.  This also solves the problem of 32/64 bit binaries having issues, the main thing is  that we need to document which packages need to be installed as a prerequisite (or strengthen the CMake checks for the packages).

I noticed that I had to install nodejs and nodejs-legacy to have node available and on path as "/usr/bin/node" (https://packages.debian.org/sid/nodejs-legacy)

@CTrauma had the editor cmake locating node on Linux: https://github.com/AtomicGameEngine/AtomicGameEngine/commit/26ec6ab93f68849268be20a83d79a314ee6baf3b

It is probably ok to require nodejs-legacy at the moment and then just use "node" in the build scripts, the other option is to only require the nodejs package, which provides a "usr/bin/nodejs" on path, so change to use "nodejs" binary on Linux.   

I am currently building on the build box, which I think may only be compiling using one make job, once it gets to the TypeScript compile will have to see if the current package version of nodejs on 14.04 LTS is new enough...

Type1J:
I had a list of what packages that I installed. I'll post it here when I'm back at that machine later today. Do you want me to take the internal references out?

JoshEngebretson:
@Type1J That list would be great, hopefully we can boil it down in the Quick and Easy method for folks: https://github.com/AtomicGameEngine/AtomicGameEngine/wiki/Building-the-Atomic-Editor-from-Source

I was able to build and run the editor on the build box!  I just symlinked system nodejs to Build/Linux/node/node for the current PR, tsc ran just fine on the 14.04 packaged node :+1:   

I think we should require nodejs and not nodejs-legacy, this means we would be looking for nodejs on debian based systems... and I think we should focus on making debian based distro's solid first, until there are more resources to put into supporting additional Linux distros.  This means we can assume nodejs will be available and on the path, as nodejs is a required package :)

I haven't tried loading up a project quite yet, though will shortly, this is great progress.  Thanks @CTrauma and @Type1J!

Type1J:
Well, this is a bit nasty.

I made a new 64-bit Ubuntu VM to match the build VM. In the official Linux tarball for Node.JS, you will get an executable called "node", but the Ubuntu package has a "nodejs" executable. We'll have to add a check for "nodejs" first, for Ubuntu distros, then fall back to "node", which on all Linux distros execpt Ubuntu, would be correct.

Type1J:
Here's the command with the package list. The only other things that I had installed were "vim" and "git" up to this point:

    sudo apt-get install build-essential cmake nodejs libgtk-3-dev libasound2-dev libxrandr-dev libgl1-mesa-dev libglu1-mesa-dev

Type1J:
I've tested building with both GNU Make and Ninja, and they both work. Assuming that you're in the repo directory, and the repo directory's name is "AtomicGameEngine" Here's 2 GNU Make build variants:

    cd ..
    mkdir AtomicGameEngine-Linux
    cd AtomicGameEngine-Linux
    cmake -DATOMIC_DEV_BUILD=1 ../AtomicGameEngine
    cmake --build . --config Release --target GenerateScriptBindings
    cmake --build . --config Release

Or for speed:

    cd ..
    mkdir AtomicGameEngine-Linux
    cd AtomicGameEngine-Linux
    cmake -DATOMIC_DEV_BUILD=1 ../AtomicGameEngine
    make -j9 GenerateScriptBindings
    make -j9

Or for more speed, using Ninja (1 line difference from the 1st GNU Make version):

    cd ..
    mkdir AtomicGameEngine-Linux
    cd AtomicGameEngine-Linux
    cmake -G Ninja -DATOMIC_DEV_BUILD=1 ../AtomicGameEngine
    cmake --build . --config Release --target GenerateScriptBindings
    cmake --build . --config Release

Type1J:
That last commit replaced the 32-bit node binary with the following script:

    #!/bin/sh
    
    NODE1=$(which nodejs)
    NODE2=$(which node)
    
    if   [ -x "$NODE1" ]; then
      NODE=$NODE1
    elif [ -x "$NODE2" ]; then
      NODE=$NODE2
    fi
    
    if [ ! -x "$NODE" ]; then
      echo "Please install Node.JS"
      return 1
    fi
    
    $NODE $@

This script takes care of the differences between Linux installations where the Node.JS could be either "nodejs" or "node", but it assumes that it is in the path. If it's not found, the build will stop (because it returns 1), and the user will see "Please install Node.JS" to let them know why it stopped.

JoshEngebretson:
Fantastic, ok landing and then will make a branch to fire up the Linux CI task for PR validation.  The CI and local build scripts are planning to get married, it just hasn't quite happened yet.  THANKS!

nifker:
cloned the repo and got this error: ../../AtomicJS/libAtomicJS.a(JSAtomic.cpp.o): In function `Atomic::jsapi_init_atomic(Atomic::JSVM*)':
JSAtomic.cpp:(.text+0xa2c): undefined reference to `Atomic::jsb_package_atomic_init(Atomic::JSVM*)'
../../AtomicPlayerJS/libAtomicPlayerJS.a(AtomicPlayerJS.cpp.o): In function `AtomicPlayer::jsapi_init_atomicplayer(Atomic::JSVM*)':
AtomicPlayerJS.cpp:(.text+0x24): undefined reference to `Atomic::jsb_package_atomicplayer_init(Atomic::JSVM*)'
collect2: Fehler: ld gab 1 als Ende-Status zurück
Source/AtomicPlayer/Application/CMakeFiles/AtomicPlayer.dir/build.make:125: die Regel für Ziel „Source/AtomicPlayer/Application/AtomicPlayer“ scheiterte
make[2]: *** [Source/AtomicPlayer/Application/AtomicPlayer] Fehler 1
CMakeFiles/Makefile2:1948: die Regel für Ziel „Source/AtomicPlayer/Application/CMakeFiles/AtomicPlayer.dir/all“ scheiterte
make[1]: *** [Source/AtomicPlayer/Application/CMakeFiles/AtomicPlayer.dir/all] Fehler 2
Makefile:83: die Regel für Ziel „all“ scheiterte
make: *** [all] Fehler 2

JoshEngebretson:
As a sanity check, please make sure you are running: ./Build_AtomicEditor.sh in the root of the source tree, it appears the JS script bindings are not being generated or added to the build properly.

Type1J:
Please see https://github.com/AtomicGameEngine/AtomicGameEngine/pull/556#issuecomment-168549642

JoshEngebretson:
@Type1J I don't believe those steps work in a clean build tree (ahead of the bootstrapping process, which interrogates the script system and generates the source files in Artifacts which need to be picked up by cmake)

JoshEngebretson:
@Type1J The GenerateScriptBindings target used to be a dependency of the executable targets, the problem being that it would generate the bindings every build with it being difficult to "smartly" generate the bindings only when necessary.  In practice, I find forcing the script bindings to be regenerated better workflow (and pretty infrequent).

On the generated script sources, during the "bootstrap" which is before cmake is invoked, this is where the script files are generated, so they may be picked up by the cmake targets: 

https://github.com/AtomicGameEngine/AtomicGameEngine/blob/b671a8e48b84c5ad34c4a20d1691bda73a0dc150/Build/Scripts/BuildLinux.js#L25

The sources are either created as empty files or if they already exist are left unmodified.

If there is a better way to generate/add the sources to CMake, I am all ears!

Type1J:
@JoshEngebretson You're right. It doesn't work in a clean build tree.

@nifker I have reproduced the errors that you show above. The problem is that when Atomic builds, it does generate some of it's C++ source for script binding (C# and JavaScript) from C++ headers and some JSON files to indicate which headers to use for that generation, and it doesn't store those generated sources in the CMake binary directory; it instead stores them back in the source tree in the Artifacts directory. That's fine because it means that you can build for multiple platforms, but only have to generate the bindings once. However, in order for CMake to have the names of all of the files that it needs to add to the projects (Makefiles) to build, the Build_AtomicEditor.sh script generates empty copies of the files that will soon be generated.

@JoshEngebretson I'm not sure that there's a better way to make CMake work for the generated source unless:

1. AtomicTool gave more responsibility to the build system by requiring invocations for each source file generated instead of finding that out itself.
or
2. AtomicTool took on more of responsibility of the build system, and was invoked for every build, but it could determine which files that it needed to regenerate by storing modified dates and making a quick compare.

I'm not sure if I like either option.

However, I do have a hackish workaround that @nifker can use. This hack was found accidentally, but I'll explain why it works. First, here's the command sequence (you could put this in a script):

    cd ..
    mkdir AtomicGameEngine-Linux
    cd AtomicGameEngine-Linux
    cmake -G Ninja -DCMAKE_BUILD_TYPE=Release -DATOMIC_DEV_BUILD=1 ../AtomicGameEngine
    cmake --build . --target GenerateScriptBindings
    cmake ../AtomicGameEngine
    cmake --build .

I believe that I did something along these lines accidentally, which is what led me to believe that the previous comment's command sequences were correct, and they are, but not for a clean checkout. The key here is the second to the last line, which reruns "cmake" to generate the build scripts (Makefiles/ninja files) with all of the currently cached variables (from the first run) after the script binding sources have been generated. It does run very quickly (about 2 seconds), and the file glob for the generated sources isn't in cache, so it will be reevaluated in that run, and it will pick up the generated sources before the final line that builds the "all" target, which doesn't modify most of the build scripts, so those builds are not run again, just the remaining targets and the binding sources are built.

I'm not sure whether or not to recommend that method in a wiki, but I've been unknowingly using it, it seems.

Type1J:
It does, however, show that there's a well defined place where one could make another CMake build directory and build when cross-compiling. I think the [Mastering CMake](http://smile.amazon.com/Mastering-CMake-Ken-Martin/dp/1930934262/) book even had something like the above sequence in the "Adding a generated file and generator" section in chapter 12, but I don't have my copy with me to check.

PR #593, opened 2016-01-13
Fixing UIView resize behavior.

Type1J:
A UIView size object never gets updated after it's created. Since many other UIWidget objects size themselves based on their parent's size, any resize causes the UI to misbehave.

Many of the AtomicExamples suffer from this problem, so for an example of what this PR fixes, try ParticleEmitter3D, which has buttons on the right side of the window. Re-size the window, or make it fullscreen, and the buttons stay where they were relative to the top-left corner of the window, so they no longer are on the right side of the screen. Another example is SpaceGame, where after startup, the menu screen (and Ok button) have similar behavior.

The AtomicEditor seems pretty good about keeping things in the right place, but even it suffers from this problem: Try making a subwindow appear, such as when the Player Output window appears when running the player, then resize the main editor window. The subwindow stays relative to where it was, which is fine, even expected. Then, close the subwindow and make it reappear (relaunch the player), and although .center() was called on it in the .ts file, it centers based on the original window size, not the new window size.

This PR fixes everything described above.

### DIFF
--- a/Source/Atomic/UI/UIView.cpp
+++ b/Source/Atomic/UI/UIView.cpp
@@ -24,6 +24,7 @@
 
 #include "UI.h"
 #include "UIView.h"
+#include "../Graphics/GraphicsEvents.h"
 
 using namespace tb;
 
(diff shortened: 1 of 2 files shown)


JoshEngebretson:
I was having a similar issue yesterday with the content root of a tab container not updating (which is also a TBWidget instance, and not a layout, like the root widget.   Here's that PR for reference:

https://github.com/AtomicGameEngine/AtomicGameEngine/pull/592

I came up with pretty much the same solution.  Though, there may be a better way here, I am thinking there is a bug in tb when resizing non-TBLayout parents, the children are not being updated properly. 

We do handle screen mode here and set the root TBWidget size:

https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/Atomic/UI/UI.cpp#L528

I'll look into this a bit before landing and see if there is a more universal fix to be had... thanks!

JoshEngebretson:
Ok, the good news is that the direct widget (without layout) parenting is in fact sane, I think what was missing on UIView was setting the gravity to all (so it pulls in all directions on resize instead of the default which is just a left/top positioning), this seemed to fix the issues mentioned in the desc
 
Would you mind checking #596 when you get a chance to make sure it addresses this... I tested the editor (window.center is working right there on resize, for like build window, player window, etc), I also tested ParticleEmitter3D, and SpaceGame (SpaceGame menu is a window with ".center()" called on it, so doesn't update again until the loop, though it does work).  

The only change, other than the cleanups, is setting the gravity: https://github.com/AtomicGameEngine/AtomicGameEngine/pull/596

If this checks out I will land it and close out this PR.

- Josh

Type1J:
Sounds good. There are sizing issues with the Breakout clone example as well, but I think that it uses sprites, not UI.

Type1J:
I'll check as soon as I can. It will be a few hours from now.

Type1J:
#596 Checks out.

JoshEngebretson:
Excellent, thanks for raising issue and doing the leg work in this PR :)

PR #611, opened 2016-01-15
Center window if it's outside of desktop resolution

rsredsq:
#599

### DIFF
--- a/Source/Atomic/Graphics/Direct3D11/D3D11Graphics.cpp
+++ b/Source/Atomic/Graphics/Direct3D11/D3D11Graphics.cpp
@@ -2807,6 +2807,12 @@ bool Graphics::GetMaximized()
 
     return SDL_GetWindowFlags(impl_->window_) & SDL_WINDOW_MAXIMIZED;
 }
+IntVector2 Graphics::GetMonitorResolution(int monitorId) const
+{
+    SDL_DisplayMode mode;
+    SDL_GetDesktopDisplayMode(monitorId, &mode);
+    return IntVector2(mode.w, mode.h);
+}
 // ATOMIC END
 
 }
--- a/Source/Atomic/Graphics/Direct3D11/D3D11Graphics.cpp
+++ b/Source/Atomic/Graphics/Direct3D11/D3D11Graphics.cpp
@@ -2796,7 +2796,7 @@ int Graphics::GetCurrentMonitor()
     return SDL_GetWindowDisplayIndex((SDL_Window*) this->GetSDLWindow());
 }
 
-int Graphics::GetMonitorsNumber()
+int Graphics::GetNumMonitors()
 {
     return SDL_GetNumVideoDisplays();
 }
--- a/Source/AtomicEditor/Application/AEEditorCommon.cpp
+++ b/Source/AtomicEditor/Application/AEEditorCommon.cpp
@@ -264,14 +264,46 @@ void AEEditorCommon::ValidateWindow()
     IntVector2 windowPosition = graphics->GetWindowPosition();
     int monitors = graphics->GetMonitorsNumber();
     IntVector2 maxResolution;
+
     for (int i = 0; i < monitors; i++)
     {
         IntVector2 monitorResolution = graphics->GetMonitorResolution(i);
         maxResolution += monitorResolution;
     }
-    if (windowPosition.x_ > maxResolution.x_ || windowPosition.y_ > maxResolution.y_ || windowPosition.x_ < maxResolution.x_ || windowPosition.y_ < maxResolution.y_)
+
+    if (windowPosition.x_ >= maxResolution.x_ || windowPosition.y_ >= maxResolution.y_ || windowPosition.x_ < 0 || windowPosition.y_ < 0)
     {
-        graphics->CenterWindow();
+        bool editor = this->GetTypeName() == "AEEditorApp";
+        JSONValue window;
+        window["x"] = 0;
+        window["y"] = 0;
+        window["width"] = 0;
+        window["height"] = 0;
+        window["monitor"] = 0;
+        window["maximized"] = editor ? true : false;
+
+        FileSystem* fileSystem = GetSubsystem<FileSystem>();
+        String path = fileSystem->GetAppPreferencesDir("AtomicEditor", "Preferences");
+        path += "prefs.json";
+
+        JSONValue prefs;
+
+        SharedPtr<File> file(new File(context_, path, FILE_READWRITE));
+        SharedPtr<JSONFile> jsonFile(new JSONFile(context_));
+
+        jsonFile->BeginLoad(*file);
+        prefs = jsonFile->GetRoot();
+
+        prefs[editor ? "editorWindow" : "playerWindow"] = window;
+
+        jsonFile->Save(*file, "   ");
+        file->Close();
+
+        graphics->SetMode(0, 0);
+        if (editor)
+        {
+            graphics->Maximize();
+        }
     }
 }
 
--- a/Source/AtomicEditor/Application/AEEditorCommon.cpp
+++ b/Source/AtomicEditor/Application/AEEditorCommon.cpp
@@ -7,6 +7,7 @@
 
 #include <Atomic/Engine/Engine.h>
 #include <Atomic/Input/Input.h>
+#include <Atomic/Graphics/Graphics.h>
 
 #include <Atomic/IPC/IPC.h>
 
@@ -55,6 +56,8 @@ AEEditorCommon::AEEditorCommon(Context* context) :
 
 void AEEditorCommon::Start()
 {
+    ValidateWindow();
+
     Input* input = GetSubsystem<Input>();
     input->SetMouseVisible(true);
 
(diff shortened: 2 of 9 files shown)


Source/AtomicEditor/Application/AEEditorCommon.cpp
```diff
@@ -255,4 +258,53 @@ bool AEEditorCommon::ReadPreferences()
...
+
+    for (int i = 0; i < monitors; i++)
+    {
+        IntVector2 monitorResolution = graphics->GetMonitorResolution(i);
+        maxResolution += monitorResolution;
+    }
+
+    if (windowPosition.x_ >= maxResolution.x_ || windowPosition.y_ >= maxResolution.y_ || windowPosition.x_ < 0 || windowPosition.y_ < 0)
+    {
+        bool editor = this->GetTypeName() == "AEEditorApp";
```
JoshEngebretson: Better to use context_->GetEditorContext()

https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/Atomic/Core/Context.h#L120

Source/Atomic/Graphics/Direct3D11/D3D11Graphics.cpp
```diff
@@ -2807,6 +2807,12 @@ bool Graphics::GetMaximized()
 
     return SDL_GetWindowFlags(impl_->window_) & SDL_WINDOW_MAXIMIZED;
 }
+IntVector2 Graphics::GetMonitorResolution(int monitorId) const
+{
+    SDL_DisplayMode mode;
+    SDL_GetDesktopDisplayMode(monitorId, &mode);
+    return IntVector2(mode.w, mode.h);
+}
 // ATOMIC END
```
JoshEngebretson: Nice add, we really need a common Graphics base class to put this stuff in, there's a ticket for that... this is what we have to do for now

Source/AtomicEditor/Application/AEEditorCommon.cpp
```diff
@@ -255,4 +258,53 @@ bool AEEditorCommon::ReadPreferences()
     return true;
 }
 
+void AEEditorCommon::ValidateWindow()
+{
+    Graphics* graphics = GetSubsystem<Graphics>();
+    IntVector2 windowPosition = graphics->GetWindowPosition();
+    int monitors = graphics->GetMonitorsNumber();
```
JoshEngebretson: Ok, let's fix "GetMonitorsNumber" from a previous PR, this should be: GetNumMonitors() (which is short for get number of monitors, sounds better in english)

Source/AtomicEditor/Application/AEEditorCommon.cpp
```diff
@@ -255,4 +258,53 @@ bool AEEditorCommon::ReadPreferences()
...
+    IntVector2 maxResolution;
+
+    for (int i = 0; i < monitors; i++)
+    {
+        IntVector2 monitorResolution = graphics->GetMonitorResolution(i);
+        maxResolution += monitorResolution;
+    }
+
+    if (windowPosition.x_ >= maxResolution.x_ || windowPosition.y_ >= maxResolution.y_ || windowPosition.x_ < 0 || windowPosition.y_ < 0)
+    {
```
JoshEngebretson: Does window position go below zero for monitors to left of main monitor?  When does this happen?

Source/AtomicEditor/Application/AEEditorCommon.cpp
```diff
@@ -255,4 +258,53 @@ bool AEEditorCommon::ReadPreferences()
...
+        SharedPtr<File> file(new File(context_, path, FILE_READWRITE));
+        SharedPtr<JSONFile> jsonFile(new JSONFile(context_));
+
+        jsonFile->BeginLoad(*file);
+        prefs = jsonFile->GetRoot();
+
+        prefs[editor ? "editorWindow" : "playerWindow"] = window;
+
+        jsonFile->Save(*file, "   ");
+        file->Close();
```
JoshEngebretson: This code block is duplicating a fair amount of code, should split out the save/load to separate methods and adjust other places this is done, also you may want to setup a function that creates default window data, so you don't have to duplicate it here and then add/remove to multiple code areas when/if we have more defaults...

Source/AtomicEditor/Application/AEEditorCommon.cpp
```diff
@@ -255,4 +258,53 @@ bool AEEditorCommon::ReadPreferences()
...
+
+        graphics->SetMode(0, 0);
```
JoshEngebretson: Let's add a note here that by specifying 0 width and height here that default values will be used for future code spelunkers

Source/AtomicEditor/Application/AEEditorCommon.cpp
JoshEngebretson: Another reason to not duplicate code like this is to have better error checking, this code does no error checking if file is actually opened, whether it loaded without error, etc

JoshEngebretson:
Ok, reviewed with some notes, thanks

Source/AtomicEditor/Application/AEEditorCommon.cpp
```diff
@@ -236,22 +298,31 @@ bool AEEditorCommon::ReadPreferences()
             prefs = jsonFile->GetRoot();
         }
 
+        file->Close();
```
JoshEngebretson: File will be closed by SharedPtr going out of scope

Source/AtomicEditor/Application/AEEditorCommon.cpp
```diff
@@ -174,46 +177,105 @@ bool AEEditorCommon::CreateDefaultPreferences(String& path, JSONValue& prefs)
...
+}
+
+void AEEditorCommon::GetDefaultWindowPreferences(JSONValue& windowPrefs)
+{
+    windowPrefs["x"] = 0;
+    windowPrefs["y"] = 0;
+    windowPrefs["width"] = 0;
+    windowPrefs["height"] = 0;
+    windowPrefs["monitor"] = 0;
+    windowPrefs["maximized"] = context_->GetEditorContext() ? true : false;
```
JoshEngebretson: You cannot get both the editor and player preferences like this, as it will either always be true (editor process) or false (player process)

Source/AtomicEditor/Application/AEEditorCommon.cpp
```diff
@@ -174,46 +177,105 @@ bool AEEditorCommon::CreateDefaultPreferences(String& path, JSONValue& prefs)
...
+
+        if (!LoadPreferences(prefs))
+            return;
+
+        JSONValue window;
+        GetDefaultWindowPreferences(window);
+
+        prefs[context_->GetEditorContext() ? "editorWindow" : "playerWindow"] = window;
+
+        graphics->SetMode(0, 0);
```
JoshEngebretson: I wanted a comment here explaining that setting the mode to 0 width/height will use engine defaults for window size and layout

Source/AtomicEditor/Application/AEEditorCommon.cpp
JoshEngebretson: Any answer for this question?

Source/AtomicEditor/Application/AEEditorCommon.cpp
JoshEngebretson: Let's leave the explicit close in though, it is fine

Source/AtomicEditor/Application/AEEditorCommon.cpp
```diff
@@ -174,46 +177,105 @@ bool AEEditorCommon::CreateDefaultPreferences(String& path, JSONValue& prefs)
...
     String path = fileSystem->GetAppPreferencesDir("AtomicEditor", "Preferences");
     path += "prefs.json";
+    return path;
+}
 
-    JSONValue prefs;
+bool AEEditorCommon::LoadPreferences(JSONValue& prefs)
+{
+    FileSystem* fileSystem = GetSubsystem<FileSystem>();
+    String path = GetPreferencesPath();
```
JoshEngebretson: GetPreferencesPath being split out to it's own method is good

JoshEngebretson:
The pull changes the script API and the editor scripts don't compiler, for example this line:

https://github.com/rsredsq/AtomicGameEngine/blob/RED-EDITOR-599/Script/AtomicEditor/ui/Shortcuts.ts#L33

uses "getMonitorsNumber"

Source/AtomicEditor/Application/AEEditorCommon.cpp
rsredsq: yes, no need that check because window position can't unexpectedly go below zero

Source/AtomicEditor/Application/AEEditorCommon.cpp
rsredsq: ok

Source/AtomicEditor/Application/AEEditorCommon.cpp
rsredsq: ok

Source/AtomicEditor/Application/AEEditorCommon.cpp
rsredsq: ok

JoshEngebretson:
Great, thanks this is looking much better.  Just needs some testing before landing...

Source/AtomicEditor/Application/AEEditorCommon.cpp
JoshEngebretson: On OSX I just received: "Window was moved to -669,171" (I am thinking this is how it would be on Windows too), the position can go below zero and I am not sure how this is handled for monitors to the left of main display, in short I am not sure how to handle this one, perhaps absolute value or perhaps leave until it can be looked at closer

LaraEngebretson:
I found the following two issues:

On Windows - (primary monitor is 1440x900, secondary monitor is 1680x1050)

* Delete prefs
* Launch the editor
* Move editor to secondary monitor and maximize window
* Open a project
* Press play and maximize the player
* Close the player and the editor
* Disconnect the secondary monitor
* Launch the editor

Issue: The editor does not start maximized

---

On Mac - (primary monitor is 2560x1440, secondary monitor is 1920x1080)

* Delete prefs
* Launch the editor
* Move editor to secondary monitor, maximize window
* Open a project
* Press play, move player to secondary monitor, and maximize window
* Close the player and the editor
* Launch the editor

Issue: The editor restarts on the primary monitor

JoshEngebretson:
What's the status on the PR?  It would be good to get this landed...

rsredsq:
@JoshEngebretson I've fixed windows issue. I also did the test on OS X and I couldn't reproduce the issue described above, editor window opens on the secondary monitor after that actions.

JoshEngebretson:
@rsredsq Great, thanks.  As long as Windows is solid, can tweak on OSX here...

@LaraEngebretson Can you test the latest PR hash for your issue cases above?

JoshEngebretson:
Checks out on Windows, OSX still has an issue, though filing a followup to address it.

LANDING! :airplane: , THANKS!

PR #619, opened 2016-01-18
Cubemap & Lightmap techniques, 0b file build fix, EngineSettings support in player

mattbenic:


Script/AtomicEditor/ui/frames/inspector/MaterialInspector.ts
```diff
@@ -57,6 +57,8 @@ var techniqueLookup = {
     "Techniques/DiffAlpha.xml": "Alpha",
     "Techniques/DiffAlphaMask.xml": "Alpha Mask",
     "Techniques/DiffAdd.xml": "Additive",
+    "Techniques/NoTexture.xml": "No Texture",
+    "Techniques/DiffLightMap.xml": "Lightmap",
     "Techniques/NoTexture.xml": "No Texture"
 };
```
JoshEngebretson: Compilation error here:

Script/AtomicEditor/ui/frames/inspector/MaterialInspector.ts(60,5): error TS2300: Duplicate identifier '"Techniques/NoTexture.xml"'.
Script/AtomicEditor/ui/frames/inspector/MaterialInspector.ts(62,5): error TS2300: Duplicate identifier '"Techniques/NoTexture.xml"'.

Resources/CoreData/Shaders/HLSL/LitSolid.hlsl
```diff
@@ -225,33 +225,41 @@ void PS(
...
+                finalColor = diff * lightColor * diffColor.rgb;
+            #endif
+
+            #ifdef AMBIENT
+                finalColor += cAmbientColor * diffColor.rgb;
+                finalColor += cMatEmissiveColor;
+                oColor = float4(GetFog(finalColor, fogFactor), diffColor.a);
+            #else
+                oColor = float4(GetLitFog(finalColor, fogFactor), diffColor.a);
+            #endif
```
JoshEngebretson: Lightmaps + shadows, very nice, will need to add an example :)

We'll need to update the GLSL too (can do that as a follow up, will make issue for it), is it possible to match whitespace formatting here, it is hard to see the code delta, shader changes really need to be well marked/commented too... if time is short right now, say the word and I'll do the HLSL update in the GLSL issue :)

Resources/CoreData/Techniques/DiffUnlitEnvCube.xml
```diff
@@ -0,0 +1,8 @@
+<technique vs="LitSolid" ps="LitSolid" psdefines="DIFFMAP" >
+    <pass name="base" vsdefines="ENVCUBEMAP" psdefines="ENVCUBEMAP" />
+    <pass name="prepass" psdefines="PREPASS" />
+    <pass name="material" vsdefines="ENVCUBEMAP" psdefines="MATERIAL ENVCUBEMAP" depthtest="equal" depthwrite="false" />
+    <pass name="deferred" vsdefines="ENVCUBEMAP" psdefines="DEFERRED ENVCUBEMAP" />
+    <pass name="depth" vs="Depth" ps="Depth" />
+    <pass name="shadow" vs="Shadow" ps="Shadow" />
+</technique>
```
JoshEngebretson: When is this technique useful?

Script/AtomicEditor/ui/frames/inspector/MaterialInspector.ts
```diff
@@ -274,7 +276,7 @@ class MaterialInspector extends ScriptWidget {
         section.contentRoot.addChild(attrsVerticalLayout);
 
         // TODO: Filter on technique
-        var textureUnits = [Atomic.TU_DIFFUSE, Atomic.TU_NORMAL, Atomic.TU_SPECULAR]; // ,Atomic.TU_EMISSIVE, Atomic.TU_ENVIRONMENT,
+        var textureUnits = [Atomic.TU_DIFFUSE, Atomic.TU_NORMAL, Atomic.TU_SPECULAR ,Atomic.TU_EMISSIVE];//, Atomic.TU_ENVIRONMENT,
         //Atomic.TU_CUSTOM1, Atomic.TU_CUSTOM2];
```
JoshEngebretson: Good add

Source/Atomic/Engine/EngineConfig.cpp
```diff
@@ -190,6 +190,14 @@ bool EngineConfig::LoadWindowConfig(const JSONValue& jwindow)
...
         else if (key == "resizable")
             engineConfig_["WindowResizable"] = GetBoolValue(jvalue, false);
+        else if (key == "width")
+            engineConfig_["WindowWidth"] = GetIntValue(jvalue, false);
+        else if (key == "height")
+            engineConfig_["WindowHeight"] = GetIntValue(jvalue, false);
+        else if (key == "positionx")
+            engineConfig_["WindowPositionX"] = GetIntValue(jvalue, false);
+        else if (key == "positiony")
+            engineConfig_["WindowPositionY"] = GetIntValue(jvalue, false);
```
JoshEngebretson: More settings options! :confetti_ball:

Source/AtomicPlayer/Application/AtomicPlayer.cpp
```diff
@@ -258,5 +262,18 @@ void AtomicPlayerApp::HandleJSError(StringHash eventType, VariantMap& eventData)
...
+    String filename = fileSystem->GetProgramDir() + "Settings/Engine.json";
+
+    if (!fileSystem->FileExists(filename))
+        return;
+
+    if (EngineConfig::LoadFromFile(context_, filename))
+    {
+        EngineConfig::ApplyConfig(engineParameters_);
+    }
+}
```
JoshEngebretson: Thanks, this is great, settings used by the player, like in the deal!

Source/ToolCore/Build/ResourcePackager.cpp
```diff
@@ -201,11 +201,6 @@ void ResourcePackager::GeneratePackage(const String& destFilePath)
             return;
         }
 
-        if (!file.GetSize())
-        {
-            return;
-        }
-
```
JoshEngebretson: Bye bye 0b file problem

Source/ToolCore/Build/BuildWindows.cpp
```diff
@@ -152,6 +152,12 @@ void BuildWindows::Build(const String& buildPath)
     if (buildFailed_)
         return;
    
+    if (!BuildCreateDirectory(buildPath_ + "/Settings"))
+        return;
+
+    if (!BuildCopyFile(GetSettingsDirectory() + "/Engine.json", buildPath_ + "/Settings/Engine.json"))
```
JoshEngebretson: This is great too, settings on deployment, this is pretty simple/powerful really...

JoshEngebretson:
Nice pull request!  

One check you can do to test for CI issues locally is to run Bulld_AtomicEditor.bat on from the root checkout, in this case it should have caught the TS script issue (let me know if it doesn't or the windows console is eating the error messages).  Thanks!

mattbenic:
Yeah terribly sorry for missing that. Will check It out tomorrow. Comes of
working on a separate branch.

JoshEngebretson:
No worries, I know you're super busy and this is good stuff :)

Resources/CoreData/Techniques/DiffUnlitEnvCube.xml
mattbenic: When your artists try to work around not having working lightmaps by using an unlit diffuse technique instead of just telling you lightmaps aren't working :p But it works now, so I thought I'd leave it in, since it works now and someone else might want it. Can remove it otherwise.

Resources/CoreData/Shaders/HLSL/LitSolid.hlsl
mattbenic: As far as I can tell the whitespace is matched? 4 spaces per indent. Note that this change did introduce another level of indent.

Source/Atomic/Engine/EngineConfig.cpp
mattbenic: Yeah we really need to configure starting location and size externally :) Looking at the config/settings being read from the commandline between player, engine and embedded player it really looks like some kind of shared code might be good there as well.

Source/ToolCore/Build/ResourcePackager.cpp
mattbenic: Yup, having it exclude 0b items was pretty simple too (just remove them from the list at this point). Is there any reason you'd prefer to include them?

The build system is going to need something a bit more complex than just including all files at some point. It's quite reasonable to expect there to be files in the resources directory that are for testing only or need to be excluded for some other reason. Unity's solution is to explicitly specify scenes in your build settings and it determines what to include from there, I think Unreal does something similar. A whitelist or blacklist (probably simpler) would probably be just fine though, and would make for customizable builds by passing different lists to the build process. The "default" list could be modified by marking particular resources (or even directories) as excluded in the inspector for those objects. .

mattbenic:
No, that's no excuse for not checking the build for compiler errors :) Is there a way I can kick off the actual CI build that reports as correct/incorrect here (after checking it with the bat)?

Source/ToolCore/Build/ResourcePackager.cpp
JoshEngebretson: Good thoughts, I was thinking about the folder approach too, maybe something like: "Include in all Builds", "Exclude from all Builds", "Include in Build" (with a combo box of what builds to include it in)

I'd like to look into how Unreal handles asset build tracking, Unity's being scene reference based... with a special folder for stuff to copy into build.  We can do some asset dependency stuff based on scene/prefab references, though I want customizable builds and to be able to drive assets solely from code... (I also want a code entry point and not a scene entry point, er, wait we already have that :stuck_out_tongue_winking_eye:) 

There should be a way to configure builds, much like in a compiler IDE, possible with preprocessor style tags, like "DEMO BUILD" or "GDC Presentation Build", or "SHIPPING", these tags could be applied to assets/folders to include in a build.  If you delete the build, then the tags would all automatically be removed, etc.

I guess the include/exclude/include in these platform builds, at a folder level, would be most expedient

Resources/CoreData/Shaders/HLSL/LitSolid.hlsl
JoshEngebretson: Ok, could you comment on what is changed?  I have to apply it over in GLSL-land and we may have to update this when pulling from Urho and that will help.  I also really need to plop in an example so I can be sure it is working...

Resources/CoreData/Techniques/DiffUnlitEnvCube.xml
JoshEngebretson: If there is a use case for it, great, otherwise let's remove it

JoshEngebretson:
You're pretty safe if you just run the Build_AtomicEditor.bat, this generates a full editor, including compiling all TS, etc minus the CI stuff required for building all the platform binaries, pulling down examples, generating API docs, merging all the binaries and resources and then building and code signing installers.  

The PR vaiidation task currently builds the platform binaries with the Linux build running ./Build_AtomicEditor.sh which compiles TS, etc (and is what caught the issue on the PR).  I could switch the PR validation to do a full editor build at this point... though this will slow down pull validations (not a big issue at this point)

Source/ToolCore/Build/ResourcePackager.cpp
mattbenic: Yeah, TBH Unity's native handling of different builds is pretty weak. It's _ok_ at handling things on different platforms, but realistically today that's just not enough. As an example on Android, you need to be able to target different chipsets with different prebuilt texture sets and now even on iOS there's native support for addon resources and if I'm not mistaken they can be refined based on device capabilities. You want to be able to have builds tailored for each of those. First step towards that would be a build that takes a list of resources to include, second would be a file format for that list. 

On the editor side, lots of possible options for making it easy to control how that gets generated, I suppose looking at what others do as a start would be great and then having a serious think about which way you want to go.

mattbenic:
I saw that Jenkins auto ran the build after a few minutes, that's fine really. It caught my (second :/ ) screwed up version of the material inspector.

Resources/CoreData/Techniques/DiffUnlitEnvCube.xml
mattbenic: Ok, I'll make sure we're not using it and remove if we aren't.

JoshEngebretson:
Great PR! 

I filed #620 to track some follow up tasks with the lightmap + shadow shader enhancements

Thanks... aaaaand LANDING! :airplane:

Issue #607, opened 2016-01-14
"Reveal in Finder" should be something different depending on the platform


Type1J:
Context menus on files in the editor have a menu item called "Reveal in Finder", which uses Finder on Mac, Explorer on Windows, and something else (Nautilus, GNOME Files, Dolphin, Thunar, etc.) on Linux (I haven't checked if it works on Linux).

I think a generic "Reveal in File Manager" could be used for Linux if a name can't be retrieved somehow (like if `xdg-mime query default inode/directory` doesn't work because xdg-utils isn't installed), but Windows certainly needs "Explorer" instead of "Finder".

JoshEngebretson:
This treads into "localization" which is better done sooner than later...

LaraEngebretson:
Closed by #651

PR #651, opened 2016-02-09
Issue #607

raheelx:
"Reveal in Finder" changes based on the platform being used. 

PS - Please test this for Mac and Linux as we're working with Windows


JoshEngebretson:
Hi @raheelx , thanks for the PR! :)

@shaddockh is going to help do some code reviews on TypeScript related PR's, he should have notes for you soon

@AtomicBuildBot ok to test

shaddockh:
Hey @raheelx, good work on this!  I was looking over the changes and have some suggestions on how to streamline it a bit and reduce some of the duplication. Typescript has a feature called ```computed keys``` that you can use on those associative arrays for the menu, so instead of needing to reconstruct both arrays based upon platform, you could do something like the following:
```
var showInFs = "Reveal in File Manager";
if (Atomic.platform == "Windows") {
  showInFs = "Reveal in Explorer";
} 
... more platforms ...
```

Then in your associative array do:
```
var assetGeneralContextItems = {
      "Rename": ["rename_asset", undefined, ""],
     [showInFs]: ["reveal_folder", undefined, ""],
     "-1": null,
    "Delete": ["delete_asset", undefined, ""]   
  };
```
The ```[showInFs]``` will get expanded at run-time to whatever value is contained in that variable.

Eventually, these menu structures will need to be refactored to be more extensible, but this should work for now.

raheelx:
Hey @JoshEngebretson  and @shaddockh : )

Thanks for the tip! I just sent a new PR now with those changes :+1:

shaddockh:
Awesome!!

@JoshEngebretson this looks good to me if you want to scan it and land it. :+1:

JoshEngebretson:
Excellent, thanks guys! LANDING :airplane:

Issue #497, opened 2015-12-11
Provide a way to jump to objects referenced in the inspector 


mattbenic:
For example when a static model is selected, clicking on the name of the applied material- or on a button next to the name of the material- exposes and highlights the material in the project hierarchy.

JohnnyWahib:
Currently busy with this task specifically for the material inspector. I have added small buttons next to each of the texture buttons that fire off an event that the ProjectFrame.ts subscribes to, sending through the path of the asset selected, and highlighting the relative path in the project frame. I'm assuming we are going to use this for other items in the different inspectors (material,models etc) so we would probably need to implement a similar case for each eventually. Does this method seem alright or is there a better way to go about this that I've completely missed?

raheelx:
Hi @JoshEngebretson :)

@JohnnyWahib and I have been working on this Material Referencer issue. So far we've got it to highlight the item in the Project Frame when it's clicked in the Material Inspector. The last issue we have with this is that the Project Frame needs to automatically scroll to and open the correct folder with the highlighted item. Do you by any chance have an example where you've done something similar to guide us?

LaraEngebretson:
Closed by #664

PR #664, opened 2016-02-18
Atomic file reference

JohnnyWahib:
Hey @JoshEngebretson, 

@raheelx and I have implemented a solution to issue #497: Provide a way to jump to objects referenced in the inspector.

These are our concerns:
- EditField Components trigger events onFocusChanged --> We would prefer it onClick.
- The ScrollTo implementation in the ProjectFrame.ts might be unnecessarily complicated.
- Animation components in array widget are not referencing.

Besides these concerns our solution is functional. :) Please let us know what you think!


JoshEngebretson:
Hi @JohnnyWahib and @raheelx , I scrolled through this, looking good!  I'll look a bit closer later today and followup to clarify the concerns.  In the meantime, let's get the build bot chewing

@AtomicBuildBot add to whitelist

Script/AtomicEditor/ui/frames/ProjectFrame.ts
```diff
@@ -50,6 +54,7 @@ class ProjectFrame extends ScriptWidget {
         this.subscribeToEvent("ResourceAdded", (ev: ToolCore.ResourceAddedEvent) => this.handleResourceAdded(ev));
         this.subscribeToEvent("ResourceRemoved", (ev: ToolCore.ResourceRemovedEvent) => this.handleResourceRemoved(ev));
         this.subscribeToEvent("AssetRenamed", (ev: ToolCore.AssetRenamedEvent) => this.handleAssetRenamed(ev));
+        this.subscribeToEvent(EditorEvents.InspectorProjectReference, (ev: EditorEvents.InspectorProjectReferenceEvent) => { this.handleInspectorProjectReferenceHighlight(ev.path) });
```
JoshEngebretson: Nice, definitely the preferred pattern instead of using raw string event names.

Script/AtomicEditor/ui/frames/ProjectFrame.ts
```diff
@@ -382,6 +391,18 @@ class ProjectFrame extends ScriptWidget {
 
     }
 
+    // Shows referenced file in projectframe
+    handleInspectorProjectReferenceHighlight(path: string): void {
+        this.assetReferencePath = path;
+        var db = ToolCore.getAssetDatabase();
+        var asset = db.getAssetByPath(this.resourceFolder.getPath() + "/" + path);
+
```
JoshEngebretson: Cool, I always get excited by new asset database and toolcore usage :)  These will also eventually work purely from the CLI :+1:

Script/AtomicEditor/ui/frames/ProjectFrame.ts
```diff
@@ -399,12 +420,17 @@ class ProjectFrame extends ScriptWidget {
...
         for (var i in assets) {
 
             var asset = assets[i];
-
             container.addChild(this.createButtonLayout(asset));
+            this.containerScrollToHeightCounter++;
         }
+       
+        var containerScroll: Atomic.UIScrollContainer = <Atomic.UIScrollContainer>this.getWidget("contentcontainerscroll");
+        containerScroll.scrollTo(0, this.containerScrollToHeight);
```
JoshEngebretson: nice API add

Script/AtomicEditor/ui/frames/inspector/AttributeInfoEdit.ts
```diff
@@ -700,6 +701,37 @@ class ResourceRefAttributeEdit extends AttributeInfoEdit {
                     }
                 }
                 this.editField.text = text;
+
+                this.editField.subscribeToEvent(this.editField, "UIWidgetFocusChanged", (ev: Atomic.UIWidgetFocusChangedEvent) => {
+
```
JoshEngebretson: More quality event handling... very nice

Source/Atomic/UI/UIScrollContainer.h
```diff
@@ -63,6 +63,8 @@ class UIScrollContainer : public UIWidget
     void SetAdaptContentSize(bool adapt);
     bool GetAdaptContentSize();
 
+    void ScrollTo(int x, int y);
```
JoshEngebretson: Good add, pretty nice JSBind automatically generates the JS/TS bindings, huh? :smile:

JoshEngebretson:
This is a really great PR, I am looking forward to trying it out!

1)  Maybe the event should be EVENT_TYPE_POINTER_DOWN instead of on focus?

2)  Looked good to me, can always circle back :)

3)  Need a little more context on this, seems like it might be a followup

I made the hard decision some months ago to rewrite major editor systems from C++ to TS.  Really glad to see such productivity with the TypeScript, especially in terms of UI work!

raheelx:
Hi @JoshEngebretson :)

Thanks for the feedback! We're glad you like it :) 
We did have a look at the mouse events but for some reason we couldn't get it to work. I'm currently busy with another issue now but we'll get back to mouse event on Monday.

A bit more on the 3rd concern:
In the animation controller  component (in the inspector view) ,when we have an object selected with multiple animations, only the current animation applied references the fbx where it was created. If there are multiple animations loaded onto the model in  the list created below the current animation slot, those aren't referencing.

JoshEngebretson:
@raheelx Great, once #667 lands, you should be able to capture pointer down widget events, something like this:

        var search = <Atomic.UIEditField>this.getWidget("filter");

        search.subscribeToEvent(search, "WidgetEvent", (ev: Atomic.UIWidgetEvent) => {

            if (ev.type == Atomic.UI_EVENT_TYPE_POINTER_DOWN) {
                console.log("pointer down");
            }

        });

I think the animation controller issue may be a good follow up issue, if it requires significant changes.

THANKS!

raheelx:
Hi Josh :)

We've implemented a "None" option in the Resource Selector Dialog Box, enabled multi-select with the shift key, and fixed some issues with the Asset Referencer.

Source/Atomic/UI/UIListView.cpp
```diff
@@ -852,6 +852,13 @@ void UIListView::SendItemSelectedChanged(ListViewItem* item)
 
 }
 
+void UIListView::selectItem(ListViewItem* item, bool select)
```
JoshEngebretson: Case should be "SelectItem" here

Source/Atomic/UI/UIListView.cpp
```diff
@@ -883,25 +890,51 @@ bool UIListView::OnEvent(const tb::TBWidgetEvent &ev)
...
                     {
-
-                        item->SetSelected(true);
-                        UpdateItemVisibility();
-
-                        SendItemSelectedChanged(item);
+                        selectItem(item, true);
                     }
 
                     SetValueFirstSelected();
```
JoshEngebretson: Cool, the ListView is a bit fragile (there are a number of reasons for this), testing changes, especially with the hierarchy frame is a good thing to do when modifying it.  We'll do some testing here before landing.

JoshEngebretson:
This is looking great, one small case issue and we'll do some testing before landing

raheelx:
thanks! I just fixed that case issue :+1:

LaraEngebretson:
@raheelx @JohnnyWahib I just tested this and it is a huge improvement. Great work! High Five :clap: !!! Landing!

Issue #625, opened 2016-01-24
libcurl CMake configuration 


JoshEngebretson:
Currently the libcurl in ThirdParty runs a slew of configuration setup and generates configuration headers when configuring the build via CMake.   This takes a lot of time, is fragile, and shouldn't be necessary as we should be able to specify the config for the platforms supported as with other third party dependencies.

Type1J:
... sorry... :frowning: It was quick to get it working for all the platforms that way. iOS is already (mostly) in the state that you describe: There are no checks. I had planned to go back and do that for all of the platforms, and basically remove most, if not all of the CMake checks, but I haven't had the time.

JoshEngebretson:
No problem, it might be a little bit before I can look at it... if you're interesting in taking the issue, lemme know, all part of the fun :balloon:

Type1J:
If you want assign the issue to me, then I'll get to it eventually.

JoshEngebretson:
Awesome, thanks... it looks like I had to invite ya to the organization in order to assign, should get an email shortly :)

JoshEngebretson:
#686 addressed the Windows side of this by adding a really simple CMake script for the platform: 

https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/ThirdParty/libcurl/LibCurlWindows.cmake

I also disabled openssl from building on Windows as it isn't currently used.  

The libcurl CMake files are a maintenance disaster and aren't currently doing much on OSX/iOS other than generating header files based on hard set configuration values.  I haven't looked into what is up on Android, though on Linux it would almost certainly be better to go off the package manager's version than a local build using libcurl's fragile CMake system:

    message(WARNING "the curl cmake build system is poorly maintained. Be aware")

I can't recall if there was a reason to not use the libcurl system framework on OSX?

Type1J:
The reason is nghttp2 integration, but everything that I do with the network involves WebSockets, so that's less of an issue if other people follow that trend (http2 makes many tiny requests fast by using a single connection, and not making you repeat headers, guaranteed compression and encryption.). I think that http2 will be included with the Apple libcurl eventually, but 1.1 with pipelining is probably fine for now. The user of the API doesn't need to do anything special to utilize any upgrades that may happen later.

In otherwords, we can scrap the official libcurl CMake code for a simplified version with pregenerated headers, as you were saying before.

JoshEngebretson:
[...]

PR #679, opened 2016-02-26
Reducing CMake configure time for OSX.

Type1J:
This is a partial for #625 for OSX only. It doesn't yet address Linux and Windows. I'll have to check Android as well, but I think that it does some try_compile calls, which could be removed.

### DIFF
--- a/Source/ThirdParty/libcurl/CMakeLists.txt
+++ b/Source/ThirdParty/libcurl/CMakeLists.txt
@@ -192,7 +192,8 @@ mark_as_advanced(DISABLED_THREADSAFE)
 option(ENABLE_IPV6 "Define if you want to enable IPv6 support" ON)
 mark_as_advanced(ENABLE_IPV6)
 if(ENABLE_IPV6 AND NOT WIN32)
-  if(IOS)
+  if(APPLE)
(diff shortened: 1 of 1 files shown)


JoshEngebretson:
What is generally done on Third Party libs is generate the required header file on the platform and then use it instead of generating the header via compiler interrogation locally and on CI.  I can see the need to do this on various Linux distros, though not VS2015/XCode/NDK.  Do you think that is an option here?  THANKS!

Type1J:
We're basically doing that, except the generation of the defines and config header is being done statically at configure time instead of interrogating the compiler. This happens as sub-second speeds.

The reason for generating at configure time instead of making a file per platform is that I did want to be able to interrogate if needed for Linux, BSD, or consoles (the SDKs changes sometimes), and there's not a noticeable difference at configure time for the platforms that can have a static configuration.

Xcode is covered by this PR; CMake configures for OSX or iOS in less than 15 seconds for the entire Atomic source tree. This should be similar for VS2015 and NDK after the PRs targeting those platforms.

JoshEngebretson:
I tried out the branch, great work!  Get's right down to compiling :+1: , Thanks!

LANDING! :airplane:

PR #693, opened 2016-03-10
Auto load submodules fix #687

### LINKED ISSUE #687: Initialize submodules automatically from build scripts, if submodules are missing
Build_AtomicEditor.* and CMake_*.* scripts in the repo root should automatically initial submodules if they are missing (repo was cloned without submodules)

honigbeutler123:
Hi just wanted to build this and found this. It works this way. You might want to reconsider the location if you want to add more submodules though.

### DIFF
--- a/Build/CMake/Modules/AtomicDesktop.cmake
+++ b/Build/CMake/Modules/AtomicDesktop.cmake
@@ -9,39 +9,22 @@ set (ATOMIC_LINK_LIBRARIES ${ATOMIC_LINK_LIBRARIES} LibCpuId SQLite)
 
 # Check whether the CEF submodule is available
 if (EXISTS ${CMAKE_SOURCE_DIR}/Submodules/CEF)
-    #Check if CEF got pulled
+    #Check if CEF got pulled by looking if the foldes is empty
     file(GLOB RESULT ${CMAKE_SOURCE_DIR}/Submodules/CEF)
     list(LENGTH ${RESULT} RES_LEN)
     if(RES_LEN EQUAL 0)
-     message(STATUS "Trying to initalise CEF submodule")
+     message(STATUS "Initialising CEF submodule")
 
-     set(PATH_STRING $ENV{PATH})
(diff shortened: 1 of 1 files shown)


honigbeutler123:
#687

Build/CMake/Modules/AtomicDesktop.cmake
```diff
@@ -9,6 +9,14 @@ set (ATOMIC_LINK_LIBRARIES ${ATOMIC_LINK_LIBRARIES} LibCpuId SQLite)
 
 # Check whether the CEF submodule is available
 if (EXISTS ${CMAKE_SOURCE_DIR}/Submodules/CEF)
+    #Check if CEF got pulled
+    file(GLOB RESULT ${CMAKE_SOURCE_DIR}/Submodules/CEF)
+    list(LENGTH ${RESULT} RES_LEN)
+    if(RES_LEN EQUAL 0)
+     # DIR is empty, so init the submodule and checkout master
```
JoshEngebretson: Another thought, should be able to detect whether git is on path, if so attempt to clone submodules.  If not, should exit with a message

honigbeutler123:
Okay i googled and found no real good way for searching if git is in the path, i made my own solution. If git is not there it makes the cmake file without webview. i did not test if that build is working.

honigbeutler123:
ah i did not read this with the should exit message. Should i change it ? would make the code less ugly because of this ultra if else case

JoshEngebretson:
Thanks :)  How about the FindGit module? https://cmake.org/cmake/help/v3.0/module/FindGit.html

honigbeutler123:
Hmm i wonder how google did not provide this on the front page with "cmake git find" ... okay so i obviously use this now and made it Required if cef is not initialising to stop the build in that case.

JoshEngebretson:
Awesome, I'll give it a test soon, THANKS! :+1:

JoshEngebretson:
Works great, thanks! LANDING! :airplane:

Issue #576, opened 2016-01-07
Prevent exclusive fullscreen mode when running the editor


LaraEngebretson:
Steps to replicate:

* Open a project
* Alt-Enter to go fullscreen
* Press Play

![fullscreencrash](https://cloud.githubusercontent.com/assets/722928/12176134/cb1dc662-b51a-11e5-84ff-a2bbf7b53cd5.PNG)

JoshEngebretson:
I do not believe exclusive fullscreen mode should be supported by the editor (though possibly should be supported by play mode), it will not play nicely with multiple top level windows, makes tabbing between applications difficult, etc.

As a follow up, we may want to support borderless fullscreen mode for editor (and almost certainly the editor's player mode)

LaraEngebretson:
Closed by #703

PR #703, opened 2016-03-16
Issue #576 User can no longer enter exclusive fullscreen mode during …

mattbenic:
…Editor mode, but can enter fullscreen when in Play Mode.


JoshEngebretson:
LANDED! :airplane: 

Closes #576

Issue #437, opened 2015-11-11
Can't add a class without a default public constructor to JSBind


mattbenic:
In some cases classes might have constructors that are not intended to be called from atomic scripts (eg using a static factory method on the class to construct the instance). 

To deal with this case, if JSBind can't find a public constructor with arguments it can handle, it should not output a constructor mapping. Currently it still tries to, which causes uncompilable code to be generated.

Vlamboljant:
We couldn't find or reproduce the case that was causing this problem, but we did handle the case of a private constructor and if a new case arises that it can be handled in the same way.

PR #705, opened 2016-03-16
Issue #437: Can't add a class without a default public constructor to JSBind

mattbenic:


Source/ToolCore/JSBind/JSBHeaderVisitor.h
```diff
@@ -475,6 +478,9 @@ class JSBHeaderVisitor : public SymbolVisitor
 
             Declaration* decl = symbol->asDeclaration();
 
+            if (name == "RefCounted" && getNameString(symbol->name()) == "RefCounted")
+                int i = 0;
+
```
JoshEngebretson: This seems to have reinserted itself...

Source/ToolCore/JSBind/JSBClass.h
```diff
@@ -136,6 +136,9 @@ class JSBClass : public Object
     JSBModule* GetModule() { return module_; }
     JSBPackage* GetPackage() { return module_->GetPackage(); }
 
+    bool HasPrivateConstruct() { return hasPrivateConstruct_; }
+    void SetHasPrivateConstruct(bool value = true) { hasPrivateConstruct_ = value; }
+
```
JoshEngebretson: I am not seeing this API addition used anywhere, also a bit questionable, skipping the constructor in the header traversal will mean there isn't a constructor function in the class, which can be tested without needing a local member variable

Source/ToolCore/JSBind/JSBHeaderVisitor.h
```diff
@@ -487,10 +493,6 @@ class JSBHeaderVisitor : public SymbolVisitor
                 if (function->isPureVirtual())
                     jclass->SetAbstract();
 
-                // only want public functions
-                if (!symbol->isPublic())
-                    continue;
-
```
JoshEngebretson: We should not be processing private methods at all as they will never be bound.  This adds overhead and will process many functions which will never be used.

Source/ToolCore/JSBind/cplusplus/Bind.cpp
```diff
@@ -1910,6 +1910,7 @@ bool Bind::visit(SimpleDeclarationAST *ast)
 
             if (Function *funTy = decl->type()->asFunctionType()) {
                 funTy->setMethodKey(methodKey);
+                funTy->setVisibility(_visibility);
```
JoshEngebretson: It seems odd that this modification would be necessary, if it is, let's make sure to mark it as modified and the reason for the modification

Script/Packages/Atomic/Container.json
```diff
@@ -1,5 +1,10 @@
 {
 	"name" : "Container",
 	"sources" : ["Source/Atomic/Container"],
-	"classes" : ["RefCounted"]
+	"classes": [ "RefCounted" ],
+	"overloads" : {
+		"RefCounted" : {
+			"RefCounted" : []
+		}
```
JoshEngebretson: Do we need access to the RefCounted constructor?  These should never be instantiated, perhaps missing something here

Source/ToolCore/JSBind/JSBHeaderVisitor.h
```diff
@@ -284,6 +284,9 @@ class JSBHeaderVisitor : public SymbolVisitor
         if (function->isStatic())
             jfunction->SetStatic(true);
 
+        if (!function->isPublic())
+            jfunction->SetSkip(true);
+
```
JoshEngebretson: This goes along with the comment about processing private methods, that is an early out which should be kept instead of setting it to skip here...

JoshEngebretson:
It seems the goal here is a follow up to static member support?  If we don't have a constructor available, the binding generation code will skip it in this PR.  So, getting whatever is needed during traversal/setup for this needs to happen.  This should be a pretty minimal set of changes?

mattbenic:
Yes, that's the idea. Currently the bindings that get generated always include a constructor, even if there isn't one on the class. I also thought it would be minimal, but it turned out to be a series of changes to get this information. If you prefer something simple like a "HasDefaultConstructor" earlier on, removing the need for these changes then we can do that instead. That just felt like a hacked way to get it done.

JoshEngebretson:
Closing as now pretty out of date, can be reopened/refactored if necessary

PR #739, opened 2016-04-04
Fix for missing keystrokes in the code editor on OSX.

keithjohnston:
Turns out this is a bug when using CEF with SDL: https://bugzilla.libsdl.org/show_bug.cgi?id=3107.
I have tested this fix and it works great for me - no more missing keystrokes in the code editor.

### DIFF
--- a/Source/ThirdParty/SDL/src/video/cocoa/SDL_cocoaevents.m
+++ b/Source/ThirdParty/SDL/src/video/cocoa/SDL_cocoaevents.m
@@ -50,6 +50,8 @@
     SDL_SendQuit();
 }
 
+// ATOMIC: Fixes missing WebView keystrokes on OSX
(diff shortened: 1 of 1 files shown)


JoshEngebretson:

@keithjohnston Great find!!!  SDL 2.0.4 is coming up, I wonder if this fix made it into there.  When patching SDL or other third party sources, marking them as modified helps maintenance when merging in new versions, and is a requirement in some of the third party licenses.  Would you mind adding something like this above the modification?

    // ATOMIC: Fixes missing WebView keystrokes on OSX (see https://bugzilla.libsdl.org/show_bug.cgi?id=3107)

Thanks! :)

keithjohnston:
Will do!

keithjohnston:
I think this is ready now?

JoshEngebretson:
Roger that, LANDING! Thanks!

PR #740, opened 2016-04-04
Dragging files into hierarchy

raheelx:
Hi @JoshEngebretson :)

A small fix to prevent the engine from crashing when you drag files into the Hierarchy Frame. I couldn't find a logged issue to reference for this though!

-Raheel


JoshEngebretson:
Nice catch, landing!

PR #742, opened 2016-04-05
Ease of use fixes for some TS Bindings

shaddockh:
While building some examples, I made some hand modifications to atomic.d.ts.  This incorporates them into the official script bindings as well as fixes some linting issues with the generated bindings.

Currently, getting a strongly typed component from a node requires you to cast the result:
``` typescript
let node = <Atomic.Sprite2D>this.node.getComponent("Sprite2D");
```

This change adds the following signature overload in addition to the current signature:
```typescript
let node = this.node.getComponent<Atomic.Sprite2D>("Sprite2D");
```
What this does is auto-casts the result to the component and ensures that the generic type being cast in is actually a descendent of Atomic.Component.  The same goes for ```Node.getJSComponent```, ```Node.createComponent```, ```Node.getOrCreateComponent```, and ```Component.getComponent```

Without this check, the TypeScript compiler would not catch such mis-uses as:
```typescript
let node = <string> this.node.getComponent("Sprite2D");
```

### DIFF
--- a/Script/Packages/Atomic/Core.json
+++ b/Script/Packages/Atomic/Core.json
@@ -19,8 +19,8 @@
 
 		"Object" : [
 			"sendEvent(eventType:string, data?:Object);",
-			"subscribeToEvent(eventType:string, callback:(data:any)=>void);",
-			"subscribeToEvent(sender:AObject, eventType:string, callback:(data:any)=>void);"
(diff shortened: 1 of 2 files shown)


JoshEngebretson:
May all your casts be true... very nice!

Issue #691, opened 2016-03-10
Profiler should exclude data from profiler output


mattbenic:
Ideally the profiler/debug hud should not include triangles and batches from rendering the hud itself.

PR #746, opened 2016-04-06
Issue #691 Profiler should exclude data from profiler output

weinandvv:
Divided the triangles into categories "All pass", "Single pass" and "Editor" (D3D9Graphics.cpp & h, DebugHud.cpp & h, Renderer.cpp & h, View.cpp)

Created a profiler widget to be added to the scene that lets you view the information easier when a scene is open (DebugHud.cpp & h, ResourceEditor.cpp & h, SceneEditor3D.cpp & h, SceneView.cpp, skin.tb.txt, UI.cpp)

### DIFF
--- a/Resources/EditorData/AtomicEditor/resources/default_skin/skin.tb.txt
+++ b/Resources/EditorData/AtomicEditor/resources/default_skin/skin.tb.txt
@@ -15,7 +15,7 @@ defaults
 	placeholder
 		opacity 0.2
 	disabled
-		opacity 0.3
+		opacity 1
 elements
 	TBButton
 		text-color #fefefe
--- a/Resources/EditorData/AtomicEditor/resources/default_skin/skin.tb.txt
+++ b/Resources/EditorData/AtomicEditor/resources/default_skin/skin.tb.txt
@@ -15,7 +15,7 @@ defaults
 	placeholder
 		opacity 0.2
 	disabled
-		opacity 1
+		opacity 0.3
 elements
 	TBButton
 		text-color #fefefe
--- a/Source/Atomic/Graphics/Direct3D11/D3D11Graphics.h
+++ b/Source/Atomic/Graphics/Direct3D11/D3D11Graphics.h
@@ -293,6 +293,9 @@ public:
     /// Return whether Direct3D device is lost, and can not yet render. Always false on D3D11.
     bool IsDeviceLost() const { return false; }
 
+    /// Return number of True primitives drawn this frame.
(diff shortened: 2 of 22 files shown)


Vlamboljant:
Hi Josh.

First part of the profiler included dividing the triangles up into 
a few categories.The amount of triangles seem to be multiplied by
the number of lights in the scene. So now the "All Passes" is the 
original triangles value. "Single Pass" excludes the multiplication
 of lights as well as the shadowmaps so it should be the actual number
  of tri's you currently have in a scene."Editor" is the tri's made up by the editor. 

Second part was to change the position of the UIElement profiler text. 
It was difficult to read when the profiler is toggled. So currently
using the UIWidgets I created a profiler that sets the UIElement text
inside the scene view so it could be easier to read. The profiler I 
made can also be toggled with the "Toggle Profiler" button when a scene is open.

An issue I had with this which seems a bit hacky was to disable the widget
I placed my UIEditFields on and changing the Disable text opacity to show
the text clearer. I used UIEditField because the UITextFields are quite 
limited.

I'm pretty sure there is a more effective way of doing these two changes 
so any help and advice would really be appreciated.

Vlamboljant:
Seems the check have failed. Problem is on mine side. Sorry for the trouble.

JoshEngebretson:
Hello!

The failure is related to the new methods being added solely to the Direct3D9 renderer, on a GL compile:

Source/Atomic/Graphics/Renderer.cpp:707:47: error: no member named 'GetSinglePassPrimitives' in 'Atomic::Graphics'
     
We need to watch the delta between renderers, this is one reason I really want to look into a bgfx renderer to consolidate the Graphics subsystem.  We also have to be really careful with moving renderer logic around as might have side effects on GL, D3D11, that don't show up on D3D9

I have a really busy couple days here, will get to reviewing this soon.  I like how a better profile display UI sounds :+1:

Vlamboljant:
I was able to add the necessary functions to OGL and D3D11 classes, but I wasn't able to test it. How do you do a GL compile?

JoshEngebretson:
Wow, that is some mighty kung-fu getting D3D11 and GL passing on the CI box :)  We do have a general issue with code duplication in having multiple Graphics subsystems.  I did see some rendering logic moved around too, which can have side effects in other renderer API's and need to be tested independently.

We'll be merging upstream changes in soon too, which will likely lead to some conflicts.  So, it may be better to handle the upstream merge before landing this PR.  In any event, thanks for the PR and I should be able to review it soon.

PR #747, opened 2016-04-08
Flip infacing normals

raheelx:
Hi @JoshEngebretson :)

This PR adds the ability to add import settings when importing models. It was needed because of the problem when adding planar models to the scene. 

How it Works:
1. Create an Import.json file in the Settings Folder of your project.
2. The Import.json allows you to set each of the postprocess flags to true or false. This is what it looks like:

```
{
  "desktop": {

    "aiFlagsDefault" : {
      "convertToLeftHanded" : true,
      "joinIdenticalVertices" : true,
      "triangulate" : true, 
      "genSmoothNormals" : true,
      "limitBoneWeights" : true,
      "improveCacheLocality" : true,
      "fixInfacingNormals" : false,
      "findInvalidData" : true,
      "genUVCoords" : true,
      "findInstances" : true,
      "optimizeMeshes" : true
    }
  }

}
```

NB: If there isn't an Import.json in the Settings folder of the game, the engine will set the flags to its default behaviour (where everything is true).

Is this okay? Please let me know if there is anything you'd like to be different. 

-Raheel


raheelx:
Please don't land this just yet, we've found an issue.

raheelx:
The issue was solved, so it can be tested now :+1:

JoshEngebretson:
Cool, having these exposed is great and a good bootstrap to expose the functionality per project.  We do need better per model import settings.  Assimp has some FBX limitations, may have to look into Autodesk's FBX SDK though thinking the license may be an issue there.

Source/Atomic/Engine/EngineConfig.h
```diff
@@ -24,43 +24,37 @@
...
+#include "../Resource/Configuration.h"
 
 namespace Atomic
 {
 
 class Context;
 
 /// Atomic engine configuration
-class EngineConfig
+class EngineConfig :
```
JoshEngebretson: Good introduction of inheritance

Source/Atomic/Resource/ImportConfig.cpp
```diff
@@ -0,0 +1,89 @@
+//
```
JoshEngebretson: This file should be in Source/ToolCore/Import most likely

Source/Atomic/Resource/ImportConfig.cpp
```diff
@@ -0,0 +1,89 @@
...
+            valueMap_["aiProcess_FixInfacingNormals"] = GetBoolValue(jvalue, true);
+        else if (key == "findInvalidData")
+            valueMap_["aiProcess_FindInvalidData"] = GetBoolValue(jvalue, true);
+        else if (key == "genUVCoords")
+            valueMap_["aiProcess_GenUVCoords"] = GetBoolValue(jvalue, true);
+        else if (key == "findInstances")
+            valueMap_["aiProcess_FindInstances"] = GetBoolValue(jvalue, true);
+        else if (key == "optimizeMeshes")
+            valueMap_["aiProcess_OptimizeMeshes"] = GetBoolValue(jvalue, true);
+    }
```
JoshEngebretson: You've GOT to have options! :+1:

Source/Atomic/Resource/ImportConfig.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+    return true;
+}
+
+bool ImportConfig::LoadDesktopConfig(JSONValue root)
+{
+    const JSONValue& jdesktop = root["desktop"];
```
JoshEngebretson: I wonder if we should have quality levels here, which I am thinking "desktop" implies high quality settings?

Source/ToolCore/Import/OpenAssetImporter.cpp
```diff
@@ -860,6 +866,70 @@ void OpenAssetImporter::CollectAnimations(OutModel* model)
...
 }
 
+void OpenAssetImporter::ApplyFlag(int flag, bool active)
+{
+    aiFlagsDefault_ &= ~flag;
+    if (active)
+        aiFlagsDefault_ |= flag;
+}
+
+void OpenAssetImporter::SetOveriddenFlags(VariantMap aiFlagParameters)
```
JoshEngebretson: Can this be (const VariantMap& aiFlagParameters, otherwise making a copy when passing to the method

JoshEngebretson:
A nit and a couple questions, otherwise looking good :)

Source/Atomic/Resource/ImportConfig.cpp
mattbenic: Do you mean like Unity's high/medium/low levels? If so would that be under each pleatform? So you'd have the reading of each set of settings under "Desktop" for example, then optional overrides for "Desktop-High", "Desktop-Medium", "Desktop-Low"? And down the line of course, not holding up this PR, right?

JoshEngebretson: The only thing I can really think of on the OpenAssetImporter is limiting bone weights or something based on quality settings and that is a stretch, so definitely not holding up PR on it...

JoshEngebretson:
Looks good, this will probably get refactored a bit in the future when exposing to UI.  Thanks, LANDING!

PR #749, opened 2016-04-11
Shift drag clone bug fix

raheelx:
Hi @JoshEngebretson 

The artists at Luma found a bug with the Shift+Drag cloning feature (it was cloning when you clicked anywhere in the editor) so this is a fix for that issue. 

-Raheel


JoshEngebretson:
@raheelx The drag clone bug fix looks good, however it appears that the other PR commits were dragged in.  You should be able to create a new branch off master and then cherry pick just the clone fix commit into it for a new PR.

Issue #623, opened 2016-01-24
Textures should have compression settings and be cached as compressed


mattbenic:
Currently textures are added to the project cache in the same format as the original, and loaded that way. So typical working formats like png and tga are loaded uncompressed. Ideally the following should happen:
1) Textures should be added to the project cache as a compressed format appropriate for the current target platform
2) The cache should support multiple target platforms, so that when switching between targets, or building for multiple targets a re-import of all resources is not required (this is a major shortfall of Unity's project cache)
3) Project platform settings should include preferred compression settings
4) Textures asset files should include optional per-platform overrides for texture compression settings
5) Editor controls for the above :)

JoshEngebretson:
Initial support will be for DDS for desktop deployment (though solution should support PVR, etc)

PR #758, opened 2016-04-14
#623 windows texture compression

mattbenic:
This PR introduces DXT compression for windows projects for HUGE savings in runtime memory use. Compression is performed using the newly included crunch library (https://github.com/richgel999/crunch.git).
We needed this in place fairly quickly, so while there really is a need for a better abstraction of the compression process (actually for cache-conversion of all assets) this was added directly to TextureImporter. On non-windows platforms, the compression setting is simply ignored.
Compression is disabled by default, but can be enabled with a setting in Settings\Import.json:

```
{
    "desktop": {

        "TextureImporter": {
            "compressTextures": true
        }
    }

}
```


mattbenic:
@JoshEngebretson It's likely the failure is due to building crunch on non-windows platforms. I'm seeing if we can get our mac build node to have a build task on it for Atomic as well, so we can run our builds through that first. 

In the meantime, please do your copy-paste magic on the build error :)

mattbenic:
Ok, we'll be setting up an OSX build tomorrow, so from then on we'll be able to run PRs through that :)

JoshEngebretson:
We do need a public CI setup, I should be able to forward build information somehow.  That stuff takes chunks of time to figure out is all.  

The build error is on Linux, though oddly am not seeing the exact error info in Jenkins log.  I can get it patched up to build on Linux and submit a PR back if it isn't obvious, lemme know. 

Error: In file included from /var/lib/jenkins/jobs/AtomicDS_BuildBinaries_Linux/workspace/Source/ThirdParty/crunch/crnlib/crn_core.h:177:0,
                 from /var/lib/jenkins/jobs/AtomicDS_BuildBinaries_Linux/workspace/Source/ThirdParty/crunch/crnlib/crn_qdxt1.cpp:3:
/var/lib/jenkins/jobs/AtomicDS_BuildBinaries_Linux/workspace/Source/ThirdParty/crunch/crnlib/crn_timer.h:52:40: warning: deprecated conversion from string constant to ‘char*’ [-Wwrite-strings]
(See full trace by running task with --trace)
Build step 'Execute shell' marked build as failure
Finished: FAILURE

CMakeLists.txt
```diff
@@ -25,7 +25,7 @@ endif()
 # this is here as QtCreator is having trouble picking up #include <Atomic/*> without it
 include_directories(${CMAKE_SOURCE_DIR}/Source ${CMAKE_SOURCE_DIR}/Source/AtomicEditor/Source)
 
-set (ATOMIC_LINK_LIBRARIES Atomic Box2D Duktape TurboBadger FreeType JO LZ4 PugiXml STB)
+set (ATOMIC_LINK_LIBRARIES Atomic Box2D Duktape TurboBadger FreeType JO LZ4 PugiXml STB crunch)
```
JoshEngebretson: These libraries are linked with pretty much everything, do we need crunch here or just in stuff used by ToolCore? If we're linking to crunch from the runtime, that will increase the runtime size significantly to support DDS saving?

Source/Atomic/Resource/Image.cpp
```diff
@@ -246,7 +247,9 @@ Image::Image(Context* context) :
     components_(0),
     cubemap_(false),
     array_(false),
-    sRGB_(false)
+    sRGB_(false),
+    pot_(false),
+    hasAlphaChannel_(false)
 {
```
JoshEngebretson: These new cached member variables which are invalidated if IsPowerOfTwo or components > 3 change, can these be simple getters which return whether image is pot or has alpha channel querying IsPowerOfTwo(width_) && IsPowerOfTwo(height_) and components > 3

Source/Atomic/Resource/Image.cpp
```diff
@@ -1211,6 +1216,91 @@ bool Image::SaveJPG(const String& fileName, int quality) const
         return false;
 }
 
+bool Image::SaveDDS(const String& fileName) const
+{
```
JoshEngebretson: I am thinking saving DDS files should be in the domain of the tooling and not runtime.  Otherwise, SaveDDS should only be available on desktop and not increase iOS/Android/Emscripten runtimes

Source/Atomic/Resource/Image.cpp
```diff
@@ -2034,7 +2124,12 @@ unsigned char* Image::GetImageData(Deserializer& source, int& width, int& height
 
     SharedArrayPtr<unsigned char> buffer(new unsigned char[dataSize]);
     source.Read(buffer.Get(), dataSize);
-    return stbi_load_from_memory(buffer.Get(), dataSize, &width, &height, (int*)&components, 0);
+    // #623 BEGIN TODO: Crunch expects all 4 components. This isn't ideal, but since
+    // precompressed versions should always be loaded anyway it doesn't matter.
+    // Might need to handle differently per platform.
+    components = 4;
+    return stbi_load_from_memory(buffer.Get(), dataSize, &width, &height, nullptr, components);
+    // #623 END TODO: Crunch expects all 4 components. This isn't ideal, but since
```
JoshEngebretson: Does GetImageData have many call sites?  Wondering about possible side effects here with something expecting/requiring different bit depths?

Source/Atomic/Resource/Image.h
```diff
@@ -138,12 +138,18 @@ class ATOMIC_API Image : public Resource
     bool SaveTGA(const String& fileName) const;
     /// Save in JPG format with compression quality. Return true if successful.
     bool SaveJPG(const String& fileName, int quality) const;
+    /// Save in DDS format. Return true if successful.
```
JoshEngebretson: Thanks for commenting :)  I need to get better about this...

Source/Atomic/Resource/Image.h
```diff
@@ -138,12 +138,18 @@ class ATOMIC_API Image : public Resource
...
     /// Whether this texture is detected as a cubemap, only relevant for DDS.
     bool IsCubemap() const { return cubemap_; }
     /// Whether this texture has been detected as a volume, only relevant for DDS.
     bool IsArray() const { return array_; }
     /// Whether this texture is in sRGB, only relevant for DDS.
     bool IsSRGB() const { return sRGB_; }
+    /// Whether this texture has power of two dimensions
+    bool IsPOT() const { return pot_; }
+    /// Whether this texture has an alpha channel
+    bool HasAlphaChannel() const { return hasAlphaChannel_; }
```
JoshEngebretson: So we have the getters, should be able to lose the members?

Source/Atomic/Resource/ResourceCache.cpp
```diff
@@ -595,7 +595,26 @@ Resource* ResourceCache::GetResource(StringHash type, const String& nameIn, bool
...
+        if (file)
+            LOGDEBUG("Loaded cached DDS " + name + ".dds");
+    }
+    if (!file)
+        file = GetFile(name, sendEventOnFailure);
+#else
+    // #623 END TODO
+    file = GetFile(name, sendEventOnFailure);
+#endif
+
```
JoshEngebretson: Nice foot work :+1:

JoshEngebretson:
Really great PR!  I have a few notes and let me know if the Linux fix takes you more than 5 minutes and I'll thump it :rabbit:

mattbenic:
It looks like it's falling over on a deprecation warning? If you could check that out on your linux box (if you have one) and PR, that would let me get the other issues in the PR sorted so long, hopefully get this landed today still :)

CMakeLists.txt
mattbenic: If we move the Save* methods out of Image, then yeah this could move to be a ToolCore only dependency.

Source/Atomic/Resource/Image.cpp
mattbenic: Sure, I can move both to be just the accessors, they're only using already cached data so that should be fine.

JoshEngebretson:
I don't think the warning should have caused a failed build, will check warning level there.  I'll kick off a build on the Linux box and see if the log was truncated for some reason.

Source/Atomic/Resource/Image.cpp
mattbenic: What about jpg, png and tga saving? Would that not make sense in tooling as well?

Source/Atomic/Resource/Image.cpp
mattbenic: The only other call point is LoadColorLUT, but I can introduce a GetImageDataForceComponents and use that in BeginLoad guarded by a windows check? BTW BeginLoad is a serious refactoring candidate :) I didn't want to do it because I wasn't sure if it was originally an Urho class, and I know merging something like that can be a royal PITA.

CMakeLists.txt
JoshEngebretson: We want to be able to save PNG and JPEG screenshots, DDS is not a common capture format.

mattbenic: Ok, just DDS it is then :)

Source/Atomic/Resource/Image.cpp
JoshEngebretson: instead of moving it we might want to keep SaveDDS on Image class for desktop builds for consistency and where increasing the runtime size doesn't matter so much.  Though, this does complicate the script bindings (somehow need per platform API bindings where they diverge).  

A preference would probably be to have SaveDDS on Image, though only available on desktop builds (so don't build crunch on Android/iOS/Emscripten)

Source/Atomic/Resource/Image.cpp
JoshEngebretson: This will probably be tripped by  the changes:

https://github.com/AtomicGameEngine/AtomicGameEngine/blob/f4c3ac360ce7376a091dc714ceff33cd17415285/Source/Atomic/Resource/Image.cpp#L883

GetImageDataForceComponents with a modification to BeginLoad sounds pretty hacky, there's a better way?

Source/Atomic/Resource/Image.cpp
mattbenic: Can leave it in on all platforms, but guard for windows and have other platforms return false and log a warning or error?

JoshEngebretson: Brilliant!, though guard for Windows/OSX/Linux :)  We should introduce a ATOMIC_PLATFORM_DESKTOP define which makes this simple... not for this PR

Source/Atomic/Resource/Image.cpp
mattbenic: It needs to only happen on Windows, and only from BeginLoad. Short of introducing the knowledge of a forced alpha layer up the chain and passing it down which I would think is worse, I don't think there's really a good alternative. The ultimate solution would be to fix crunch's expectation of 4 components.

Source/Atomic/Resource/Image.cpp
mattbenic: Perfect. doing it now.

Source/Atomic/Resource/Image.cpp
JoshEngebretson: Add a bool parameter to GetImageData called forceComponents that defaults to false, if true it will force the components to whatever was passed in (by reference) otherwise will use component size of image and maintain compatible behavior?  If this sounds good, add some checks for 0 and > 4 components and log error return null.

JoshEngebretson:
lzma_Threads.h line 13 HANDLE not defined, pretty sure Linux thinks its Windows on the crunch stuff

Source/Atomic/Resource/Image.cpp
mattbenic: With SaveDDS being suppoorted on all desktop platforms, that does mean we'd need to do this on all of those as well, just a guard change, but just FYI.

Resulting code in BeginLoad would be:
`
#if defined(ATOMIC_PLATFORM_WINDOWS) || defined (ATOMIC_PLATFORM_LINUX) || defined (ATOMIC_PLATFORM_OSX)
        // #623 BEGIN TODO: Crunch expects all 4 components. This isn't ideal, but since
        // precompressed versions should always be loaded anyway it doesn't matter.
        // Might need to handle differently per platform.
        // #623 END TODO: Crunch expects all 4 components. This isn't ideal, but since
        unsigned char* pixelData = GetImageDataForceAlphaLayer(source, width, height, components);
#else
        unsigned char* pixelData = GetImageData(source, width, height, components);
`

Still including components as a reference param is a tossup, but it keeps this cleaner.

mattbenic:
The quality on libsquish seems fine. The processing definitely feels like it takes longer, though I haven't done a side-by-side comparison. It's ok, we can live with that, particularly with the network cache tool currently under development and likely to be completed next week :)

mattbenic:
And failing again :( I assume this is another issue with multiplatform builds. If you haven't had a chance to check it out by then, when I'm next onsite I'll try it against our mac. Still need to make a plan for a linux build node as well.

JoshEngebretson:
Great, thanks for testing it out.  It is one of those cases where debug build takes a lot longer.  I didn't time it either, though in debug I was concerned about speed.  Switching to release, it was a lot faster. 

The build failure was on Windows, which looks to be a msbuild flakiness that strikes once in awhile.  Let's try a build again :)

@AtomicBuildBot test this please

JoshEngebretson:
.... and passed, yay!

Great feature add, thanks!  LANDING!

mattbenic:
NP, sorry about the initial choice of compression lib.

JoshEngebretson:
Yeah, crunch is cool and does a bunch of stuff, just a tad gnarly on the dependency side of things for Atomic

libsquish was missing the DDS container stuff which bogged me down, though Image.cpp already had the dds desc structure defined in it!  That was a weird convergence!  

For everyone not on the CI Box LAN, needing to guess why a build failed is a problem to solve.  This is probably going to come down to forwarding failed logs somewhere in the short term.

mattbenic:
Would emails of committers not be in the git logs? You could send mails to those?

JoshEngebretson:
It could be an opt in for email notifications possibly, there are jokers like me who put bogus email addresses into their git infos.  I'd like to avoid email and have the build info someplace central.  

Microsoft uses Jenkins for all the dotnet foundation repositories: http://dotnet-ci.cloudapp.net/

They just have a security team is all and I don't trust myself to secure it + keep it secured with a close eye on the open port for intrusion.  The nature of CI is a bunch of connected boxes which makes it all that much more of a security concern

Maybe there is a way to mirror the build info to an external Jenkins install, that would be cool.  Also, now that we're MIT we qualify for various CI and other free hosting stuff.  So, should look into options here...

mattbenic:
Just one more thing on the plate ;) Have you posted on gitter, facebook, twitter and your forums to see if there isn't someone that would be interested in contributing in that way-by working on the CI? I'm sure it's not just coders that would be keen on helping a project like Atomic

PR #761, opened 2016-04-20
Editor plugin context menus

mattbenic:
Just a quickie to add context menu support to the editor plugins


JoshEngebretson:
Nice work!  LANDING! :)

Script/AtomicEditor/ui/frames/menus/HierarchyFrameMenu.ts
```diff
@@ -110,7 +115,8 @@ class HierarchyFrameMenus extends Atomic.ScriptObject {
                 node.scene.sendEvent("SceneEditNodeCreated", { node: newnode });
             }
 
-            return true;
+            // Let plugins handle context
+            return ServiceLocator.uiServices.hierarchyContextItemClicked(target, refid);
```
shaddockh: I know I'm coming in late to the PR, but would it make more sense to pass the ```node``` attached to the target instead of the UIWidget?  I would think that most plugins would be interested in that node and not the menu item.  Also, in the other menu handler, if the menu was already handled, I didn't pass it on to the service locator.

mattbenic: Great suggestion, that'll simplify use a lot. And good catch on the return. I'll make those changes tomorrow, I'm out of office today.

PR #767, opened 2016-04-26
Alphabetic ts bindings

mattbenic:
Sorts items in generated .d.ts alphabetically. This is to avoid different machines with different directory sorting default generating different versions which are a pain for source control.


JoshEngebretson:
Awesome!  Let the *.d.ts improvements begin! :)

I'll look into this bizarro build error on Linux, which has absolutely no bearing on this PR
```
Error: /var/lib/jenkins/jobs/AtomicDS_BuildBinaries_Linux/workspace/Source/ThirdParty/kNet/src/unix/UnixEvent.cpp: In member function ‘void kNet::Event::Set()’:
/var/lib/jenkins/jobs/AtomicDS_BuildBinaries_Linux/workspace/Source/ThirdParty/kNet/src/unix/UnixEvent.cpp:157:32: warning: ignoring return value of ‘ssize_t read(int, void*, size_t)’, declared with attribute warn_unused_result [-Wunused-result]
  read(fd[0], &val, sizeof(val));
(See full trace by running task with --trace)
Build step 'Execute shell' marked build as failure
Finished: FAILURE
```

JoshEngebretson:

JoshEngebretson:
The linux console was obscuring the clang error, which is being reported on OSX. 
```
/Users/jenkins/Home/workspace/AtomicDS_BuildAtomicTool_Mac/Source/ToolCore/JSBind/JSBTypeScript.cpp:291:21: error: non-const lvalue reference to type 'Vector<[...]>' cannot bind to a temporary of type 'Vector<[...]>'
    Vector<String>& constants = module->GetConstants().Keys();
```

I'll apply a fix in a local branch and land this later today

JoshEngebretson:
Thanks for the PR, merged via quick fix in #770

PR #768, opened 2016-04-26
Adding extra event loop support for per frame events

Type1J:
Adding requestAnimationFrame() and cancelAnimationFrame().

These offer per frame callbacks, and have an API that matches the web browser API exactly, so external tweening libraries (like GSAP) work in Atomic Game Engine without modification.

...To be perfectly honest, GSAP doesn't work out of the box. Here's a frontend script that I use:

    require("AtomicEventLoop");
    (function (globalVar) {
        if (globalVar.window === undefined) {
            globalVar.window = globalVar;
        }
        if (globalVar.global === undefined) {
            globalVar.global = globalVar;
        }
    })(new Function('return this')());
    require("GSAP/TweenLite");

This script assumes that:
1, "TweenLite.js" is in "Modules/GSAP/TweenLite.js".
2. The version of Atomic Game Engine used has this pull request.

I am going to go through the UI library soon and make another PR for the modifications to make most tweening libraries (Tween.js, GSAP, and others) work without update callback hacks necessary for some values now.

I think you could add TimelineLite/TweenMax/TimelineMax to the GSAP.js script above, but I haven't tried.

### DIFF
--- a/Resources/CoreData/AtomicModules/AtomicEventLoop.js
+++ b/Resources/CoreData/AtomicModules/AtomicEventLoop.js
@@ -16,6 +16,8 @@
  *      - Modified the way that setTimeout,clearTimeout,setInterval,clearInterval are loaded into global scope
  *      - Implemented setImmediate/clearImmediate
  *
+ *  Jay Sistar - April 2016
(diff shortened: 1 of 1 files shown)


JoshEngebretson:
Great work!  Thanks!

PR #771, opened 2016-04-27
Combine d.ts files

### LINKED ISSUE #764: Combine generated d.ts files into a single file
On build, all of the d.ts files that are generated should be combined into a central atomic.d.ts file.  Something to look into in order to accomplish this might be ```dts-generator``` at https://github.com/SitePen/dts-generator

shaddockh:
This PR adds an extra build step that will combine all of the generated d.ts files into a single Atomic.d.ts and place it under Script/TypeScript/dist and also placed in the distribution under Data/AtomicEditor/TypeScriptSupport.

A new dependency was brought in to accomplish this: https://github.com/SitePen/dts-generator which is under the New BSD, so if that's not compatible then I'll determine another way to do this.

This closes #764


JoshEngebretson:
Neat!  I'll have a closer look at this tomorrow, looking forward to scrolling through the amalgamated Atomic.d.ts :+1:   

The New BSD license is fine and don't fret about the CI error.  I am having difficulties with the build box,  Winders(tm)!!!

JoshEngebretson:

shaddockh:
Cool. I'm starting to dive back into the TS integrations and this should help clean things up.  😀

Build/CIScripts/GenEditorData.js
```diff
@@ -137,10 +137,12 @@ namespace('build', function() {
     process.chdir(atomicRoot);
 
     var tsc = "./Build/node_modules/typescript/lib/tsc";
+    var dtsGenerator = "./Build/node_modules/dts-generator/bin/dts-generator";
 
     cmds = [
       atomicRoot + "Build/Mac/node/node " + tsc + " -p ./Script",
-      atomicRoot + "Build/Mac/node/node " + tsc + " -p ./Script/AtomicWebViewEditor"
```
JoshEngebretson: +18,856 lines and -1 line... here be that line

JoshEngebretson:
Great, checks out here and the amalgamated Atomic.d.ts is freaking wunderbar ❤️ 

One thing that would be good is to have all generated stuff filtered into ./Artifacts instead of in the source tree.  I think there might be a few places to address this and won't hold up this PR for it... LANDING! Thanks! :)

shaddockh:
It should be easy enough to generate it somewhere else (just change the --out param).  I've got it generating in the source tree and then copied out to a location in artifacts.  Currently, it copies to:
```./Data/AtomicEditor/TypeScriptSupport/Atomic.d.ts``` . Is this correct?  Or should the process that copies all the files into the app bundle, also copy the generated files into the correct directory structure?

Might be good to start an issue to track build protocol.

JoshEngebretson:
I think temporary files should definitely be generated under Artifacts, this way when doing a non-Dev Build everything will be built/generated/copied in the Artifacts folder.  

Dev Builds run from the resources/data in the source tree, which is really helpful while developing.  This fact complicates keeping the source tree clean, wrt generated files.  We probably do have to copy required files into the source tree for the time being.

I'll paste this into an issue :)

PR #778, opened 2016-04-30
User preferences system for extensions

shaddockh:
This PR adds the ability for host extensions to store and retrieve preferences from the project user preferences file.  These preferences are also propagated to the web client so that editor extensions can retrieve the settings.  Additionally, whenever preferences are changed from the host, a preference change event is sent to the web client so the web client can respond appropriately.

The other changes here are some lint whitespace changes and also refactoring of the webview client extension names and structure to more closely match the changes made the host extension naming done in a previous PR.

I did have a question on the user preference file.  I'm using: 
```
ToolCore.toolSystem.project.userPrefsFullPath 
```
to store the preferences in and this points to a file called ```projectname.userpreferences```.  It appears that the Snap Settings dialog stores settings in a different file called ```UserPrefs.json``` .  Should't those also be stored in the ```projectname.userpreferences``` file?  

I updated the routine that saves the Snap Settings to not overwrite the file, but to retain any extraneous settings that may also be in there and just update with the known settings, so switching over to the ```projectname.userpreferences``` file should be seamless if that's the way you would like to go.  Either that or have ```ToolCore.toolsSystem.project.userPrefsFullPath``` point to the ```UserPrefs.json``` file.

I'm going to be using this to coordinate some settings in the TypeScript extension, but thought I would break out just the user settings piece into it's own PR so it's easier to digest.

### DIFF
--- a/Script/AtomicEditor/editor/Editor.ts
+++ b/Script/AtomicEditor/editor/Editor.ts
@@ -161,6 +161,18 @@ class Editor extends Atomic.ScriptObject {
         this.sendEvent(EditorEvents.UserPreferencesChangedNotification);
     }
 
+    /**
+     * Sets a group of user preference values in the user settings file located in the project.  Elements in the
+     * group will merge in with existing group preferences.  Use this method if setting a bunch of settings
+     * at once.
+     * @param  {string} settingsGroup name of the group the preference lives under
+     * @param  {string} groupPreferenceValues an object literal containing all of the preferences for the group.
+     */
+    setUserPreferenceGroup(settingsGroup: string, groupPreferenceValues: Object) {
+        Preferences.getInstance().setUserPreferenceGroup(settingsGroup, groupPreferenceValues);
+        this.sendEvent(EditorEvents.UserPreferencesChangedNotification);
+    }
+
     handleEditorLoadProject(event: EditorEvents.LoadProjectEvent): boolean {
 
         var system = ToolCore.getToolSystem();
--- a/Script/AtomicEditor/editor/Editor.ts
+++ b/Script/AtomicEditor/editor/Editor.ts
@@ -145,7 +145,10 @@ class Editor extends Atomic.ScriptObject {
      * @param  {number | boolean | string} defaultValue value to return if pref doesn't exist
      * @return {number|boolean|string}
      */
-    getUserPreference(extensionName: string, preferenceName: string, defaultValue?: number | boolean | string): number | boolean | string {
+    getUserPreference(settingsGroup: string, preferenceName: string, defaultValue?: number): number;
+    getUserPreference(settingsGroup: string, preferenceName: string, defaultValue?: string): string;
+    getUserPreference(settingsGroup: string, preferenceName: string, defaultValue?: boolean): boolean;
+    getUserPreference(extensionName: string, preferenceName: string, defaultValue?: any): any {
         return Preferences.getInstance().getUserPreference(extensionName, preferenceName, defaultValue);
     }
 
--- a/Script/AtomicEditor/editor/Editor.ts
+++ b/Script/AtomicEditor/editor/Editor.ts
@@ -55,7 +55,7 @@ class Editor extends Atomic.ScriptObject {
 
         this.editorLicense = new EditorLicense();
 
-        EditorUI.initialize();
+        EditorUI.initialize(this);
 
         this.playMode = new PlayMode();
 
@@ -136,6 +136,29 @@ class Editor extends Atomic.ScriptObject {
         Preferences.getInstance().saveEditorWindowData(editorWindowData);
 
         return true;
+    }
+
+    /**
+     * Return a preference value or the provided default from the user settings file
+     * @param  {string} extensionName name of the extension the preference lives under
+     * @param  {string} preferenceName name of the preference to retrieve
+     * @param  {number | boolean | string} defaultValue value to return if pref doesn't exist
+     * @return {number|boolean|string}
+     */
+    getUserPreference(extensionName: string, preferenceName: string, defaultValue?: number | boolean | string): number | boolean | string {
+        return Preferences.getInstance().getUserPreference(extensionName, preferenceName, defaultValue);
+    }
+
+
+    /**
+     * Sets a user preference value in the user settings file
+     * @param  {string} extensionName name of the extension the preference lives under
+     * @param  {string} preferenceName name of the preference to set
+     * @param  {number | boolean | string} value value to set
+     */
+    setUserPreference(extensionName: string, preferenceName: string, value: number | boolean | string) {
+        Preferences.getInstance().setUserPreference(extensionName, preferenceName, value);
+        this.sendEvent(EditorEvents.UserPreferencesChangedNotification);
     }
 
     handleEditorLoadProject(event: EditorEvents.LoadProjectEvent): boolean {
--- a/Script/AtomicEditor/editor/Preferences.ts
+++ b/Script/AtomicEditor/editor/Preferences.ts
@@ -145,6 +145,63 @@ class Preferences {
     static getInstance(): Preferences {
         return Preferences.instance;
     }
+
+    /**
+     * Return a preference value or the provided default from the user settings file located in the project
+     * @param  {string} settingsGroup name of the group these settings should fall under
+     * @param  {string} preferenceName name of the preference to retrieve
+     * @param  {number | boolean | string} defaultValue value to return if pref doesn't exist
+     * @return {number|boolean|string}
+     */
+    getUserPreference(settingsGroup: string, preferenceName: string, defaultValue?: number | boolean | string): number | boolean | string {
+        const prefsFileLoc = ToolCore.toolSystem.project.userPrefsFullPath;
+        if (Atomic.fileSystem.fileExists(prefsFileLoc)) {
+            let prefsFile = new Atomic.File(prefsFileLoc, Atomic.FILE_READ);
+            try {
+                let prefs = JSON.parse(prefsFile.readText());
+                if (prefs && prefs[settingsGroup]) {
+                    return prefs[settingsGroup][preferenceName] || defaultValue;
+                }
+            } finally {
+                prefsFile.close();
+            }
+        }
+
+        // if all else fails
+        return defaultValue;
+    }
+
+    /**
+     * Sets a user preference value in the user settings file located in the project
+     * @param  {string} settingsGroup name of the group the preference lives under
+     * @param  {string} preferenceName name of the preference to set
+     * @param  {number | boolean | string} value value to set
+     */
+    setUserPreference(settingsGroup: string, preferenceName: string, value: number | boolean | string) {
+
+        const prefsFileLoc = ToolCore.toolSystem.project.userPrefsFullPath;
+        let prefs = {};
+
+        if (Atomic.fileSystem.fileExists(prefsFileLoc)) {
+            let prefsFile = new Atomic.File(prefsFileLoc, Atomic.FILE_READ);
+            try {
+                prefs = JSON.parse(prefsFile.readText());
+            } finally {
+                prefsFile.close();
+            }
+        }
+
+        prefs[settingsGroup] = prefs[settingsGroup] || {};
+        prefs[settingsGroup][preferenceName] = value;
+
+        let saveFile = new Atomic.File(prefsFileLoc, Atomic.FILE_WRITE);
+        try {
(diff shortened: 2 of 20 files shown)


Script/AtomicEditor/editor/Editor.ts
```diff
@@ -138,6 +138,41 @@ class Editor extends Atomic.ScriptObject {
         return true;
     }
 
+    /**
+     * Return a preference value or the provided default from the user settings file
+     * @param  {string} extensionName name of the extension the preference lives under
+     * @param  {string} preferenceName name of the preference to retrieve
+     * @param  {number | boolean | string} defaultValue value to return if pref doesn't exist
+     * @return {number|boolean|string}
+     */
```
JoshEngebretson: The extra effort on the doc strings really looks great...

Script/AtomicEditor/editor/Editor.ts
```diff
@@ -138,6 +138,41 @@ class Editor extends Atomic.ScriptObject {
...
+    getUserPreference(extensionName: string, preferenceName: string, defaultValue?: number | boolean | string): number | boolean | string {
```
JoshEngebretson: Wow, TS has logical OR type definitions?!?!?!?!  Cool!

Script/AtomicEditor/editor/Editor.ts
```diff
@@ -138,6 +138,41 @@ class Editor extends Atomic.ScriptObject {
...
+     * group will merge in with existing group preferences.  Use this method if setting a bunch of settings
+     * at once.
+     * @param  {string} settingsGroup name of the group the preference lives under
+     * @param  {string} groupPreferenceValues an object literal containing all of the preferences for the group.
+     */
+    setUserPreferenceGroup(settingsGroup: string, groupPreferenceValues: Object) {
+        Preferences.getInstance().setUserPreferenceGroup(settingsGroup, groupPreferenceValues);
+        this.sendEvent(EditorEvents.UserPreferencesChangedNotification);
+    }
+
```
JoshEngebretson: This is great and there is a laundry list of things to hook up to prefs :)

Script/AtomicEditor/editor/Preferences.ts
```diff
@@ -145,6 +147,111 @@ class Preferences {
...
+        try {
+            saveFile.writeString(JSON.stringify(prefs, null, "  "));
+        } finally {
+            saveFile.flush();
+            saveFile.close();
+        }
+
+        // Cache the update
+        this.cachedProjectjPreferences = prefs;
+    }
```
JoshEngebretson: We likely should be moving some ToolCore prefs/JSON handling up to TS scripts.  This is so much easier to parse than in C++

Script/AtomicWebViewEditor/clientExtensions/ClientExtensionServices.ts
```diff
@@ -189,4 +202,51 @@ export class ExtensionServiceRegistry extends ServiceRegistry<Editor.ClientExten
             }
         });
     }
+
+    /**
+     * Called when prefeerences changes
```
JoshEngebretson: Typo, though can catch it in next go around :)

Script/AtomicWebViewEditor/interop.ts
```diff
@@ -213,6 +218,22 @@ export default class HostInteropType {
...
         editorCommands.resourceDeleted(path);
     }
+
+    /**
+     * Host is notifying client that there are preferences to load and passing us the path
+     * of the prefs.
+     * @param  {string} prefUrl
+     */
+    loadPreferences(prefUrl: string) {
+        console.log("Load preferences called for :" + prefUrl);
```
JoshEngebretson: The log is starting to have a fair bit of output in it.  So, we should probably start trying to catch debugging code before landing,  A way to set debug level logging would be good, we already have this native side, just needs to be wrapped.

Source/Atomic/Resource/ResourceEvents.h
```diff
@@ -95,6 +95,11 @@ EVENT(E_PROJECTUNLOADEDNOTIFICATION, ProjecUnloadedNotification)
 {
 }
     
+/// User Preferences Changed
+EVENT(E_USERPREFERENCESCHANGEDNOTIFICAITON, UserPreferencesChangedNotification)
```
JoshEngebretson: Does this strictly need to be added as a native event?  Event types can be defined in script... I'll have a look at the handler and see about some thoughts there.

Source/AtomicEditor/Editors/JSResourceEditor.cpp
```diff
@@ -107,33 +110,44 @@ String getNormalizedPath(const String& path)
...
+void JSResourceEditor::HandleUserPreferencesChangedNotification(StringHash eventType, VariantMap& eventData)
+{
+    ToolSystem* tsys = GetSubsystem<ToolSystem>();
+    Project* proj = tsys->GetProject();
+    FileSystem* fileSystem = GetSubsystem<FileSystem>();
+    if (fileSystem->FileExists(proj->GetUserPrefsFullPath()))
+    {
+        webClient_->ExecuteJavaScript(ToString("HOST_loadPreferences(\"atomic://%s\");", proj->GetUserPrefsFullPath().CString()));
+    }
+}
```
JoshEngebretson: I don't see anything here that couldn't be accomplished in script.  Do you have thoughts on having this event defined/handled native side?

JoshEngebretson:
I reviewed the PR and have a few notes.  

On the pref file question, I think we need global and project prefs, snap settings are probably a global editor setting.  

I like the idea of handling prefs in the TS scripts instead of native code.  It looks really clean and is good to sort now, as we don't have many prefs hooked up yet!

Source/Atomic/Resource/ResourceEvents.h
shaddockh: The only reason I have this as a native event is that I can't determine how to propagate an event from the editor host to the web client instance directly.  The only way I've found to do it is through the C++ side.  If there is a way to trigger events in the web client from the editor typescript, or even call webclient javascript from the editor, that would make my life TONS easier.

Script/AtomicEditor/editor/Editor.ts
shaddockh: I actually use https://atom.io/packages/docblockr .  It stubs out the doc comments when you type ```/**``` and you just have to fill in the details.  Very handy.

shaddockh:
added a few line comments.  If there is a way to move away from having the native event, I'd definitely vote for that.

JoshEngebretson:
I'll branch the PR and have a go at converting to a script event sometime tomorrow and we'll see what happens :)

JoshEngebretson:
Really, looking over JSResourceEditor and seeing as the UI & TS bindings have leveled up so much since I initially wrote the control, there isn't much reason this is native at all.  That's a good thing!

Source/AtomicEditor/Editors/JSResourceEditor.cpp
```diff
@@ -188,6 +191,16 @@ void JSResourceEditor::HandleWebMessage(StringHash eventType, VariantMap& eventD
...
             }
+            else if (message == EDITOR_GET_USER_PREFS)
+            {
+                ToolSystem* tsys = GetSubsystem<ToolSystem>();
+                Project* proj = tsys->GetProject();
+                FileSystem* fileSystem = GetSubsystem<FileSystem>();
+                if (fileSystem->FileExists(proj->GetUserPrefsFullPath()))
+                {
+                    webClient_->ExecuteJavaScript(ToString("HOST_loadPreferences(\"atomic://%s\");", proj->GetUserPrefsFullPath().CString()));
+                }
```
shaddockh: The reason that this is still here is that there needs to be some kind of notification event to the typescript indicating that the page is fully loaded.  Currently, there is not so the typescript can't execute any javascript on editor creation because those methods don't exist yet.  Ideally this will get removed, I just need to think on how to notify the typescript when the page is done loading...possibly in the EDITOR_LOAD_COMPLETE block above.  Maybe do a sendEvent with EDITOR_LOAD_COMPLETE that can be hooked into.

shaddockh:
I'm definitely on board with moving those events out of the native and into the typescript.  Should that happen in this PR or another PR?

JoshEngebretson:
Maybe we should just get the native event removed and moved over to be handled solely in script.  Land early, land often? :)

JoshEngebretson:
Sorry, upon reflection that probably isn't entirely clear as I thought you were asking about the native editor source for some reason.  Let's get the native event moved over to script for this PR if possible.

shaddockh:
Sounds good.

E_USERPREFERENCESCHANGEDNOTIFICAITON has already been removed and is being handled in the TypeScript for the most part other than the lines in JSResourceEditor.cpp I commented on above.  I'll look this weekend to see how I can remove that part as well as try to migrate the other native events (ResourceRenamed, ResourceDeleted, and ProjectUnloaded) to the Typescript.

Script/AtomicEditor/editor/Editor.ts
```diff
@@ -138,6 +138,44 @@ class Editor extends Atomic.ScriptObject {
...
+     * Return a preference value or the provided default from the user settings file
+     * @param  {string} extensionName name of the extension the preference lives under
+     * @param  {string} preferenceName name of the preference to retrieve
+     * @param  {number | boolean | string} defaultValue value to return if pref doesn't exist
+     * @return {number|boolean|string}
+     */
+    getUserPreference(settingsGroup: string, preferenceName: string, defaultValue?: number): number;
+    getUserPreference(settingsGroup: string, preferenceName: string, defaultValue?: string): string;
+    getUserPreference(settingsGroup: string, preferenceName: string, defaultValue?: boolean): boolean;
+    getUserPreference(extensionName: string, preferenceName: string, defaultValue?: any): any {
```
JoshEngebretson: Nice TypeScript-fu, once again

Script/AtomicEditor/ui/frames/ResourceFrame.ts
```diff
@@ -226,6 +226,23 @@ class ResourceFrame extends ScriptWidget {
         }
     }
 
+    handleUserPreferencesChanged() {
```
JoshEngebretson: This is way better now with script access to the view, though we really need to refactor the initial version of editors to make it more pluggable... not for this PR!

JoshEngebretson:
I missed the commit with the move from native to script for the event, much nicer.  We can circle back on the load complete bit.  Do you have the events you need for that? The PR looks great, landing :)

shaddockh:
I may need to create a native event for when the web client is fully loaded, but once I have that, then I can move all of the other events that I was hooking into in the native side to the TypeScript and it should make it much cleaner.  The issue is that the web client loads up the page and all of the additional javascript asynchronously, so the native doesn't know when it's all loaded until the web client tells it.  At that point, the native should be able to broadcast a web client loaded event to the typescript and then the typescript can handle the post-load processes such as telling the client to load the source, give the client the user preferences, etc.

JoshEngebretson:
Ok, if this event isn't cutting it, I can look into one that will, so no need to spend time on it:

https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/AtomicWebView/WebViewEvents.h#L47

Hopefully, that's the one!  Never can be too sure in async-land though!

shaddockh:
Hmm.  That looks promising.  I'll see if that gets fired after everything is loaded or only once the main window is created.

shaddockh:
Nevermind, I don't think that will work with async.  That's why I added the event from the client in the first place.
https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/AtomicEditor/Editors/JSResourceEditor.cpp#L140-L144

shaddockh:
It looks like there needs to be an event for DOM Ready.  I found some information here:
http://magpcss.org/ceforum/viewtopic.php?f=6&t=10277

JoshEngebretson:
Do you think you could file an issue referencing that CEF thread with what is needed for the DOM Ready event for our usage? Lara and I are on vacation next week and will try and get it landed before leaving.

shaddockh:
I've been experimenting with some things and have some ideas for a solution that might work. I need to try some things out this weekend, but if it doesn't work then I'll file an issue.  From what I'm reading, onLoadEnd should work, but the system.js used in the client may be causing some problems because it's streaming additional content in after the page is loaded.

shaddockh:
fyi, the idea I had will work.  I'll clean things up and hopefully have a PR for you by the end of the weekend.  If not, you can take a look at it when you guys get back and have a good trip! :)

PR #782, opened 2016-05-09
Added SetSubimage to copy contents of an image to a rect in another image

mattbenic:
Added this as part of the lightmapping work, but someone on gitter was looking for this functionality specifically so decided to PR separately.


JoshEngebretson:
Great, landing :)

PR #784, opened 2016-05-10
Generate API documentation with TypeDoc

### LINKED ISSUE #668: Investigate TypeScript API doc generation with TypeDoc
Once initial TS API doc generation is hooked up, can integrate with build so available in Atomic Editor binary distro.  The Atomic API is large enough this should be a "multi-page" project, if TypeDoc supports it.  If not, will probably need an alternative.

The minimal theme seems nice: http://typedoc.io/themes/minimal

shaddockh:
This PR should resolve #668 

This generates a multi-page API site from the d.ts files.  It also updates the comment parser to be able to handle multi-line formal comment extraction in case there is a need for more verbose documentation of API methods.
```
/**
This is a formal comment
*/
```

I'm unsure if this will work correctly with the CI builds, so I'm going to need some help with that.


JoshEngebretson:
Hey!  I am in the Sierra Nevada's with the saddest of internet connections (which is perfect!)

The CI boxes are down while we are out of town.  I'll land this and we can address the CI build in post, with notes!

The multiline comments will be good, I was thinking about this last week!

PR #787, opened 2016-05-11
Referenced typescript dependencies in generated definition files

JohnnyWahib:
Generated .d.ts files were missing references to dependencies, causing ts files that used them to fail to compile.

### DIFF
--- a/Source/ToolCore/JSBind/JSBPackage.cpp
+++ b/Source/ToolCore/JSBind/JSBPackage.cpp
@@ -195,6 +195,7 @@ bool JSBPackage::Load(const String& packageFolder)
                 return false;
             }
 
+            dependencies_.Push(depPackage);
         }
 
     }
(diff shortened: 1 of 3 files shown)


mattbenic:
This highlighted a more general issue that the .d.ts files aren't tested in CI. Perhaps a in inclusion test script that gets built and compiled as part of the AtomicTool bind command would sort that out?

mattbenic:
Actually, if @shaddockh's combined .d.ts includes all the editor defines as well as the atomic ones it should sort things out.

JoshEngebretson:
I'll investigate whether the combined .d.ts includes everything, this could still be a useful addition with some factoring, if we do split up dependencies.   They would just need to be marked in the package (ie. not core or something so not in Atomic.d.ts)

shaddockh:
FYI, The combined d.ts includes everything in the d.ts files in the TypeScript directory, including the editor and work definitions.

JoshEngebretson:
Excellent, I will close this for now.  @JohnnyWahib If we still need this, let me know and can reopen

PR #814, opened 2016-05-27
fix for editor crash on start in Linux

CTrauma:
changed the array size to '2' to reflect actual size


JoshEngebretson:
@CTrauma Good catch, thanks!

@AtomicBuildBot add to whitelist, test this please

PR #839, opened 2016-06-07
pre-set compile on save for new TypeScript projects

shaddockh:
To make things more seamless for new TypeScript projects, auto set the Compile on Save setting.  This is so that when you get a new project, you can just edit and run without needing to toggle any settings.


JoshEngebretson:
Nice add, thanks

PR #842, opened 2016-06-08
Fix for Material and Scene dialogs 

shaddockh:
This fixes the issue where the new scene and the new material dialogs were re-using the new component dialog which has been changed.


JoshEngebretson:
Good catch :)

PR #847, opened 2016-06-09
Copy paste tool for components

Vlamboljant:
Hi @JoshEngebretson 

Made it possible to let you copy and paste the settings of one component on a node to a similar component on a different node.

Can not copy and paste to different components as expected.

If nothing is copied nothing will paste as also expected


JoshEngebretson:
Awesome, here's the compilation error which is the same on Linux and Mac:

/Users/jenkins/Home/workspace/AtomicDS_BuildBinaries_MacEditor/Source/AtomicEditor/Editors/SceneEditor3D/SceneSelection.cpp:440:77: error: cannot pass object of non-trivial type 'Atomic::String' through variadic function; call will abort at runtime [-Wnon-pod-varargs]
        LOGERRORF("Component Copy: Unable to copy component from node: %s", component->GetAttribute("type").ToString());

JoshEngebretson:
Great work! One improvement for the UI would be to have copy/paste next to each other using a layout, instead of full button for each which increases vertical space usage

Excellent code, small fix over in https://github.com/AtomicGameEngine/AtomicGameEngine/pull/853 which will land from there, THANKS! ✈️

PR #848, opened 2016-06-09
Orthographic views

Vlamboljant:
Hi @JoshEngebretson 

Added switching from perspective- to orthographic views. 
Added snapping to a specific view if an object is selected. If no object is selected the camera wil snap to the view relative to its position.

Things you are able to do:

1. Switch between perspective and orthographic using "O" key.

2. Snap to an orthographic view using keys 1-6.
    (I was very unsure as to which keys I should assign this to)
    1-Top
    2-Bottom
    3-Left
    4-Right
    5-Front
    6-Back

3. You can switch back to perspective by pressing the letter "O" key even after pressing keys 1-6.

4. It will remember the perspective position you were in before switching to orthographic. You can revert to that position by pressing the letter "P" key.

5. Can zoom while in orthographic using the mouse wheel


JoshEngebretson:
Great, we definitely need multiple viewports.  I am not sure how intuitive the P key is, though if there are editing packages that use that convention, cool.  We'll want viewport controls in the UI, though 1-6 for selection works now.

One thing that would be good in the future would be to squash commits, I have to get better about this myself.  It makes it quite a bit more difficult to rebase when there are conflicts.

There were conflicts with the other PR's so I rebased and will land from here: https://github.com/AtomicGameEngine/AtomicGameEngine/pull/854

Thanks! :)

mattbenic:
Next step on this would definitely be to get some kind of axis widget in place like in other editors, but the shortcuts met our needs for the time being and we thought it better to PR as is rather than hold off (we don't know if/when we'll get to a widget), since it _is_ useful :)

PR #850, opened 2016-06-09
New navigation

Vlamboljant:
Hi @JoshEngebretson 

Changed how to increase and decrease camera speeds. You no longer use shift key. You can simply hold down the right mouse button and adjust the speed using the mouse wheel.

No longer have to use alt to zoom in and out. Adjust the zoom using only the mouse wheel. It will zoom in on where the mouse pointer is or the object selected.


JoshEngebretson:
This feels great!  Very nice usability improvements, one small issue with OSX mouse wheel being backwards, fixed in #851 and will land from there

Thank you!

PR #865, opened 2016-06-14
fix for compiling with gcc 6.1.1 on Arch Linux

CTrauma:
Tested with Ubuntu  GCC 5.3 and has no effect in older compilers


JoshEngebretson:
Thanks!  I filed #866 to deal with the crusty local copy of stb_image and stb_truetype in TurboBadger

PR #869, opened 2016-06-15
Basic editing key bindings for Linux

JimMarlowe:
Add the arrow, backspace, delete to the Linux Ace editor, making for an enjoyable editing experience.
To say CEF support for linux is a little weak would not be an understatement. After trying to find the right combinations of key events, states, values, I found, by looking at the CEF browser unit tests for GTK, that they were translating GTK key scancodes to Windows scancodes, then using the methodology that windows uses to talk to the embedded browser. Now if that had been documented, it would have been done faster.
There will need to be still more follow on work to handle translation for the function and keypad keys, and making sure the full set of ace key bindings work.

### DIFF
--- a/Source/AtomicWebView/WebClient.cpp
+++ b/Source/AtomicWebView/WebClient.cpp
@@ -688,11 +688,13 @@ void WebClient::SendKeyEvent(const StringHash eventType, VariantMap& eventData)
 #endif
 
 #ifdef ATOMIC_PLATFORM_LINUX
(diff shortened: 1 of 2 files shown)


JoshEngebretson:
> by looking at the CEF browser unit tests for GTK, that they were translating GTK key scancodes to Windows scancodes, then using the methodology that windows uses to talk to the embedded browser.

Wow, that is nuts, thanks for the hard work, I know how futzy this particular key mapping can be and it looks like you hit the mother load!  LANDING!

PR #871, opened 2016-06-16
remove unused dependency on ATL

CTrauma:
This should make building with VC++ Build Tools and mingw easier.


JoshEngebretson:
Awesome, thanks for digging!  LANDING! ✈️

PR #877, opened 2016-06-17
TypeScript enhancements

### LINKED ISSUE #845: Support TS and ES6 style 'export default' for components
To make components cleaner, an enhancement would be to support ```export default``` for components.  This would change components from being defined as:
```typescript
class MyComponent extends Atomic.JSComponent {
...
}
export = MyComponent;
```
to simply
```typescript
export default class MyComponent extends Atomic.JSComponent {
...
}
```

The logic would need to look for a property called ```default``` on the module because the transpiled JS looks like
```javascript
exports.default = MyComponent
```

shaddockh:
This PR adds some additional overloads to make some APIs work better as generics so that the resulting return value is auto-cast to the right type.  Namely:
```
ResourceCache.getResource<T>
ResourceCache.getTempResource<T>
ResourceCache.getExistingResource<T>
UI.getWidgetAt<T>
UIWidget.getWidget<T>
```

The other thing this PR does is allow for ```export default <classname>``` which is used in ```ES6``` and ```TypeScript``` which closes #845.  The project templates have been updated to use ```export default``` for the ```TypeScript``` projects as this is a cleaner and more preferable way of doing things.

@JoshEngebretson can you double check the C++ changes?  They work, but I'm still getting my bearings in C++ and there could be a better way of handling that.

### DIFF
--- a/Data/AtomicEditor/ProjectTemplates/Project2D/TypeScript/Resources/Components/Star.ts
+++ b/Data/AtomicEditor/ProjectTemplates/Project2D/TypeScript/Resources/Components/Star.ts
@@ -3,7 +3,7 @@
 /**
  * Component that will rotate a node at a configurable speed.
  */
-class Star extends Atomic.JSComponent {
+export default class Star extends Atomic.JSComponent {
 
     /**
      * Fields witihin the inspectorFields object will be exposed to the editor
@@ -28,5 +28,3 @@ class Star extends Atomic.JSComponent {
 
     }
 }
-
-export = Star;
--- a/Script/Packages/Atomic/Resource.json
+++ b/Script/Packages/Atomic/Resource.json
@@ -7,5 +7,12 @@
 			"GetPixel": ["int", "int"],
 			"SetSize": ["int", "int", "int", "unsigned"]
 		}
+	},
+	"typescript_decl" : {
+		"ResourceCache" : [
+			"getResource<T extends Resource>(type: string, name: string, sendEventOnFailure?: boolean): T;",
+			"getTempResource<T extends Resource>(type: string, name: string, sendEventOnFailure?: boolean): T;",
+	        "getExistingResource<T extends Resource>(type: string, name: string): T;"
+		]
 	}
 }
--- a/Source/AtomicJS/Javascript/JSComponent.cpp
+++ b/Source/AtomicJS/Javascript/JSComponent.cpp
@@ -280,12 +280,21 @@ void JSComponent::InitInstance(bool hasArgs, int argIdx)
             return;
         }
 
-        duk_get_prop_string(ctx, -1, "component");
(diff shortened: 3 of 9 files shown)


Data/AtomicEditor/ProjectTemplates/Project2D/TypeScript/Resources/Components/Star.ts
```diff
@@ -3,7 +3,7 @@
 /**
  * Component that will rotate a node at a configurable speed.
  */
-class Star extends Atomic.JSComponent {
+export default class Star extends Atomic.JSComponent {
```
JoshEngebretson: Very nice

Script/Packages/Atomic/Resource.json
```diff
@@ -7,5 +7,12 @@
 			"GetPixel": ["int", "int"],
 			"SetSize": ["int", "int", "int", "unsigned"]
 		}
+	},
+	"typescript_decl" : {
+		"ResourceCache" : [
+			"getResource<T extends Resource>(type: string, name: string, sendEventOnFailure?: boolean): T;",
+			"getTempResource<T extends Resource>(type: string, name: string, sendEventOnFailure?: boolean): T;",
+	        "getExistingResource<T extends Resource>(type: string, name: string): T;"
+		]
```
JoshEngebretson: Great stuff on the generics support here

Source/AtomicJS/Javascript/JSComponent.cpp
```diff
@@ -280,12 +280,21 @@ void JSComponent::InitInstance(bool hasArgs, int argIdx)
...
+            duk_pop(ctx);
+
+            // If "default" doesn't exist, look for component
+            duk_get_prop_string(ctx, -1, "component");
+
+            if (!duk_is_function(ctx, -1))
+            {
+                duk_set_top(ctx, top);
+                return;
+            }
```
JoshEngebretson: Cool, this looks right

Source/AtomicJS/Javascript/JSComponentFile.cpp
```diff
@@ -145,16 +153,26 @@ bool JSComponentFile::InitModule()
...
+            {
+                LOGERRORF("Component file export object does not export a key \"component\" function: %s", GetName().CString());
+                duk_set_top(ctx, top);
+                return false;
+            }
+
+            scriptClass_ = false;
+        } else {
+            // this is a TS or ES2015 default class
+            scriptClass_ = true;
```
JoshEngebretson: This looks good too

JoshEngebretson:
Awesome, and congrats on pulling around some Duktape :)  Thanks, LANDING! :+1:

Type1J:
Thanks @shaddockh! That's a great add.

PR #880, opened 2016-06-18
Fix for user provided modules starting with "Atomic"

shaddockh:
This fixes the issue where a module starting with "Atomic" pulls from the wrong location if it is not a built-in file.


JoshEngebretson:
Good fix, need to evaluate this moving forward, an early convenience hack whose time has maybe come and gone.  Thanks!  Landing

Issue #844, opened 2016-06-08
Android support on all platforms


JimMarlowe:
A proposal for enabling full android support on Win, OSX, Linux.

All platforms can produce debug and release apks. (#31)
If devices are connected, attempt installation, with later versions of Android things get complicated.
The paths are available on all platforms, some platforms can use system installed programs and paths. (#81)
Support direct typing into the path fields (#536)
All the paths and settings will be put into project BuildSettings.json file. (#818)
The icon is a button with current icon, and the filename next to it. Pressing the icon button brings up a finder to seek a new icon. (#754)

The [x] Release Path checkbox signifies that the user wants to build a release apk. It is unselected by default. It can be made insensitive until a release path has a value. 
Make the "Release Path" be the place where the user stores the ant.properties which contains the keystore and password info to make a release apk.
If the user selects Release and does not provide needed keystore information, a debug apk is created.

Build Settings Android layout

App Name :            [_____________]
App Package :        [_____________]
Company Name :   [_____________]
Product Name :      [_____________]

Android SDK Path :
[____________________] [Choose]

Android API Level:
[____________________] [Refresh]

Ant Path :
[____________________] [Choose]

Java JDK Path :
[____________________] [Choose]

[x] Release Path :
[____________________] [Choose]

[____________________] [Icon]  

Would you like to know more?

JoshEngebretson:
This sounds great, I especially like the linking to existing issues that will be closed in  the process :)

PR #893, opened 2016-06-29
Andriod toolchain for debug and release apks from #844

JimMarlowe:
This is the result of the proposal in #844. Everything in #844 was implemented, in addition to the OK and Cancel buttons on the Build Settings dialog perform their assigned functions. This gives a fairly wide path for users to deploy both debug and release APKs from the editor. The build processing changed very little, and the error messages have been enhanced to tell what the problem is and to give possible fixes, in some cases. The release APKs are done by the user modifying ant.properties to include the keystore location and the passwords, so the build process does not need to be stopped to enter these, and since the directory where the user keeps this file is set into ToolPrefs, they only needs to do it once.
I need help testing it on windows and OSX, I can not build on these platforms.
And when it lands, the wiki page can be updated to reflect the new capabilities.


JoshEngebretson:
Excellent, release APKs!!! I breezed through the code and looking good, will have a closer look soon.

We can have @LaraEngebretson do some Android deployment testing on Windows/Mac :)

JoshEngebretson:
"9 successful checks" - Congrats on compiling on all platforms in one go :)

JimMarlowe:
Some notes from looking over the wiki, which can help with the testing.
Note the minimum Android API level is 12, anything below this will not compile successfully.  ( from anecdotal testing, you need API level 15 to make the Galaxy S4 not act weird ).

An android app has a set of icons, and may use different ones depending on the situation. These are the folder names and the sizes of the icons that are expected to be present.
   drawable-ldpi/con.png  size = 36x36 pixel 
   drawable-mdpi/icon.png  size = 48×48 pixels
   drawable-hdpi/icon.png  size = 72×72 pixels
  drawable/logo_large.png  size =  512x512 pixels  (Google Play store) 
You can create your own set of icons, and set the icon path to the directory containing these directories and files as shown above. If the selected path is correct, the icon in the finder button will change to the drawable-ldpi icon.  If no selection is made, the default Atomic icons are used for the app.

Release Apps.
In order to create a release app, you will need a certificate to apply to the app as it is being built. Certificates may be purchased, or they can be self-signed. They are put in a keystore, with passwords for the store and alias. This information can be put in the ant.properties to make the authorization process automatic. When the Release Path checkbutton is set, you must enter the directory that contains the  ant.properties or the build will fail.  (This is as short as I could make it, it needs to be expanded. It has enough keywords that you can google your way thru this maze, or you make a debug apk. )

Resources/EditorData/AtomicEditor/editor/ui/buildsettings_android.tb.txt
```diff
@@ -20,40 +21,49 @@ TBLayout: axis: y, distribution: gravity, position: left
...
 			TBEditField: id: product_name, autofocus: 1
 				lp: min-width: 220
+	TBLayout: axis: y, position: left, distribution: gravity
+		TBLayout: gravity: left right, distribution-position: right bottom
+			TBEditField: id: icon_root, autofocus: 0
+				lp: min-width: 265
+			TBButton: id: choose_icon, lp: min-width: 48, min-height:48
+				tooltip location of the drawable folders for app icons
+				TBImageWidget: id: and_icon, filename: "Resources/ToolData/Deployment/Android/res/drawable-ldpi/icon.png"
 	TBSeparator: gravity: left right, skin: AESeparator
```
JoshEngebretson: The default TBImageWidget path should derived in code as a development build pulling assets from source tree, and release build pulling from packaged editor assets (possibly in an app bundle) may be different, I am a little fuzzy on this, though the default image isn't showing up on a packaged build, I can look into this... may be unrelated

Resources/EditorData/AtomicEditor/editor/ui/buildsettings_android.tb.txt
```diff
@@ -20,40 +21,49 @@ TBLayout: axis: y, distribution: gravity, position: left
...
 	TBLayout: axis: y, position: left, distribution: gravity
 		TBTextField: text: "Android SDK Path:", skin: DarkGrayText
 		TBLayout: gravity: left right, distribution-position: right bottom
 			TBEditField: id: sdk_path, autofocus: 0
 				lp: min-width: 250
-			TBButton: text: "Choose" id: choose_sdk_path
+			TBButton: text: "Choose", id: choose_sdk_path
+				tooltip location of the Android SDK folder
```
JoshEngebretson: Nice on adding tool tips 👍

Script/AtomicEditor/ui/modal/build/platforms/AndroidSettingsWidget.ts
```diff
@@ -75,66 +83,47 @@ class AndroidSettingsWidget extends Atomic.UIWidget implements BuildSettingsWind
...
+                this.releaseNameEdit.text = path;                
+                return true;
 
-        }
+            }  else if (ev.target.id == "choose_icon") {
+                var fileUtils = new Editor.FileUtils();
+                var path = fileUtils.getIconPath("");
+                this.iconNameEdit.text = path;
+                this.iconImage.setImage(path + "/drawable-ldpi/icon.png");
+                return true;
```
JoshEngebretson: Cool :)

Source/AtomicEditor/Utils/FileUtils.cpp
```diff
@@ -239,5 +239,48 @@ void FileUtils::RevealInFinder(const String& fullpath)
...
+        iconPath = outPath;
+    }
+
+    if (outPath)
+        free(outPath);
+
+    GetSubsystem<Graphics>()->RaiseWindow();
+
+    return GetInternalPath(iconPath);
+}
```
JoshEngebretson: GetReleasePath and GetIconPath differ only in the message they display, maybe collapse into one method that takes a message string?  Do we have another method that these are based on not showing up in diff these are based on?

Source/ToolCore/Build/AndroidProjectGenerator.cpp
```diff
@@ -135,10 +140,35 @@ bool AndroidProjectGenerator::GenerateLocalProperties()
     File file(context_, buildPath_ + "/local.properties", FILE_WRITE);
 
     if (!file.IsOpen())
+    {
+        errorText_ = "Project generator unable to open file " + buildPath_ + "/local.properties ";
         return false;
-
```
JoshEngebretson: Better error messages are better!

Source/ToolCore/Build/AndroidProjectGenerator.cpp
```diff
@@ -281,6 +313,55 @@ bool AndroidProjectGenerator::GenerateAndroidManifest()
...
+    userIconDir = userIconPath + "/drawable-hdpi"; 
+    userIconFile = userIconDir + "/icon.png"; 
+    destDir = buildPath_ + "/res/drawable-hdpi";
+    if ( fileSystem->FileExists (userIconFile) )
+    {
+        Poco::File icon4 ( userIconFile.CString() );
+        icon4.copyTo( destDir.CString() );
+    }
+
+    return true;
```
JoshEngebretson: Custom icons, awesome :)

Source/ToolCore/Build/BuildAndroid.cpp
```diff
@@ -368,6 +383,13 @@ void BuildAndroid::Build(const String& buildPath)
     mfile->WriteString(manifest);
     mfile->Close();
 
+    //check for Deployment/Android/libs/armeabi-v7a/libAtomicPlayer.so
+    if ( !fileSystem->FileExists(androidProject + "/libs/armeabi-v7a/libAtomicPlayer.so")  )
+    {
+        FailBuild( " the file libAtomicPlayer.so is not found. This is required for APK generation." );
+        return;
+    }
```
JoshEngebretson: Nice catch

JoshEngebretson:
This is really solid, I have one note about collapsing FileUtil dialogs creation into a single method.  @LaraEngebretson is doing some testing with a packages build of the editor and will have some followups.  Thanks!

JimMarlowe:
I understand about the FileUtils concerns. If you want the two new methods compressed, there is no reason why most of the others methods dealing with paths couldn't be compressed in the same way. It would involve adding two new string args, and move the strings from FileUtils.cpp to wherever in the ts code they may be called. And it would make a better generic service for future needs in the editor. Give the command, and I'll make this change.  I'd like to mention that the defaultpath strings for these methods doesn't make it down from ts to c++ code, but I wasn't ready to tilt that windmill in this PR. That would make those path finders even friendlier. 
I'll look at the default icon button in a released setting, I suspected that it was wrong putting it in the tb.txt file, I'll see if I can move it into the BuildSettings windows constructor.
I'd be interested to hear the testing results, and what could be made better, like if the tooltips helped or hindered or could be platform specific, and if errors were encountered and it was obvious what to do about them.

LaraEngebretson:
Wow, this is a huge level up on Android deploy! 😄  Here are my notes from Windows/Mac testing:

1. Unable to deploy from Mac, cannot select API level and the Refresh button doesn't populate the dropdown menu
2. The project has to be deployed at least once in order for the drawable folder to exist and for the icons to show up in the Build Settings dialog.
appicon01

3. Opening the Build Settings dialog, after restarting the editor, resizes the window and cuts off some of the right-hand side 
appicon02

4. Input fields lose the current path name if you click the Choose button and then select Cancel

5. Using Search to browse folders opens multiple finder windows (OSX only), to replicate:
  * Click "choose"
  * Enter text in the Search field
  * Select a folder
  * Hit Ok to confirm, twice

6. On tooltips: first word of tooltip should be capitalized and no periods are necessary

JoshEngebretson:
I'll look into 1) at first opportunity, thanks for testing!

JoshEngebretson:
@JimMarlowe Anything that can be done to make FileUtils more friendly, please feel free :)  I can also look into 5) if the issue isn't obvious

JimMarlowe:
For 1, This sounds like an android installation issue, I looked in the file that the code lives and saw that it doesn't fall under the build processing, so no errors would show up anyway to tell you what's wrong.
I added a generic FindPath and FindFile functions to FileUtils, and they seem to work fine. I'll switch everything over that makes sense. 
5. is not obvious, The linux file finder does not have any editable fields, and the windows one does, but does not have that behavior. Did the OSX finder work correctly before my changes?
For the rest, I should be able to finish them off.

Source/ToolCore/Build/AndroidProjectGenerator.cpp
```diff
@@ -135,10 +140,35 @@ bool AndroidProjectGenerator::GenerateLocalProperties()
...
+        
+        String antname = Reldir + "/ant.properties";
+        if ( !fileSystem->FileExists ( antname) ) 
+        {
+            errorText_ = "The file ant.properties not found in " + Reldir + ", unable to generate Release APK.";
+            return false;
+        }
+
+        Poco::File antp ( antname.CString() ); // lookup ant.properties 
+        antp.copyTo( buildPath_.CString());  // copy offending file
```
JoshEngebretson: Another note, it is good to use the build system directory creation, file copies, etc and not Poco as the build system ones check for and propagate errors: https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/ToolCore/Build/BuildWindows.cpp#L132

Source/ToolCore/Build/AndroidProjectGenerator.cpp
```diff
@@ -281,6 +313,55 @@ bool AndroidProjectGenerator::GenerateAndroidManifest()
...
+    if (!fileSystem->DirExists(userIconPath))               // dont do anything if there is no path defined.
+        return true;
+            
+    String userIconDir = userIconPath + "/drawable";        // 1st target dir 
+    String userIconFile = userIconDir + "/logo_large.png";  // 1st target file
+    String destDir = buildPath_ + "/res/drawable";          // where it should be in the build
+    if ( fileSystem->FileExists (userIconFile) )            // is there a file there?
+    {
+        Poco::File icon1 ( userIconFile.CString() );        // declare the file
+        icon1.copyTo( destDir.CString() );                  // copy the file into the build
```
JoshEngebretson: Here too on the file copies

JoshEngebretson: There's a slug of stuff in BuildBase, so probably pass in the Android build instance for the project generator to use, alternatively could refactor this stuff out to BuildSystem, probably opt for the former as it touches less code :) https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/ToolCore/Build/BuildBase.cpp#L146

JimMarlowe: Oh my, BuildBase.cpp is quite the treasure trove.  This may have to be a refactor job. The BuildBase file operator methods are in the protected area of the class, so they will have to get moveed into the public area to be usable passed into the Android project generator. Not a lot of risk to do this. I'll audit the other builders also, and update them, I see some are not using the BuildBase file ops, if its not done now, it'll never be done. And it will provide a good examples for the BuildLinux module.
The call to fileSystem->RemoveDir(buildPath_, true); is recursive crash code on linux. This is called when you attempt to create a project where one already exists. Windows does not use this code. The Poco File::remove(true) does work. If RemoveDir wont work, I can ifdef it for linux, and test on OSX.

JoshEngebretson: Awesome, my C++ visibility knowledge is failing this morning, if we made the Android project generator a friend of the Android build class would it have access to the protected BuildBase methods?  If so, maybe opt for this... and sounds great on the audit.

JimMarlowe:
These are the things I fixed: If no iconpath  is set, show the icon in the deployment relative location. Tooltips fixed. Locked down icon buttonTBImageWidget size, it mostly works. Fix cancel in finder wiping out existing paths. Added generic FindPath and FindFile to FileUtils class, dog-fooded most methods. Added BuildCopyDir() to BuildBase, and dog-fooded the BuildPlatform files, friends and all.

JoshEngebretson:
@JimMarlowe Great work and thanks for the general BuildBase cleanups 👍   

It looks like we have some merge conflicts now, I'll rebase/squash the branch and fix whatever this and test out/address anything I run into with OSX deploy to Android.  @LaraEngebretson already tested Windows side and it worked great...

JoshEngebretson:
I rebased to current master, applied a couple minor cleanups, and tested on OSX...  will land from PR #899

Thanks! :)

PR #894, opened 2016-06-30
Improvements to CMake generation bash and bat scripts

JoshEngebretson:


JoshEngebretson:
Closing, will reopen once target build order is screwing up on Linux

PR #905, opened 2016-07-05
Refine .gitignore for Script/TpeScript to only exclude non-Work d.ts

mattbenic:
One of our branches ended up missing tsconfig.json after a git clean -xdf because of this


JoshEngebretson:
Great 👍

PR #909, opened 2016-07-06
Expose fontFamily to the monaco editor prefs

shaddockh:
@JoshEngebretson , I've quickly exposed ```fontFamily``` to the monaco prefs.  Also, the AtomicNET.d.ts automatically gets added back in by the Atom editor if it exists, so I added a negative glob so this doesn't happen anymore.


JoshEngebretson:
Ok, cool.  I don't *think* we'll be scripting C# from TS, though definitely the reverse will be possible.  Landing  :)

PR #910, opened 2016-07-06
Linux only ForceGL2 based in gfxlib version

JimMarlowe:
This PR fixes a common problem with the Linux Editor and player. Depending on a specific Mesa library version, a 3d scene, like ToonTown, becomes a flashing mess when viewed.  Urho3D has the same issue.  As a first impression, it's bad, real bad. The solution is to detect and compare mesa lib versions, and if the linux distro is afflicted, set a value in tool.prefs, to account for this problem.  There are complications, the detection code only runs after the graphics are initialized, and the flag you need to set to make the flashing stop can only be set before the graphics are initialized. So there is a one execution lag to starting the fix in the editor, this is reported in the log. The editor player is passed the flag thru the command line, so that is transparent to the user. 
   For the linux standalone player, the user will have to change the Settings/Engine.json file manually to set the ForceGL2 flag, there is no way to do that automatically, since the user has to know that they need the Settings directory.
   This processing is automatic, so the user does not have to do anything, whether they have the problem or not. If they upgrade their Mesa Libs to a version that does not have the problem, it will fix itself, with the one execution lag.

### DIFF
--- a/Source/Atomic/Graphics/OpenGL/OGLGraphics.h
+++ b/Source/Atomic/Graphics/OpenGL/OGLGraphics.h
@@ -252,6 +252,9 @@ public:
     /// Return graphics API name.
     const String& GetApiName() const { return apiName_; }
 
+    /// Return current driver version patch information 
+    const String& GetVersionPatch() const { return versionPatch_; }
+
     /// Return window position.
     IntVector2 GetWindowPosition() const;
 
@@ -723,6 +726,8 @@ private:
     String orientations_;
     /// Graphics API name.
     String apiName_;
+    /// String to report driver version problems
+    String versionPatch_;
 
     /// Pixel perfect UV offset.
     static const Vector2 pixelUVOffset;
--- a/Script/AtomicEditor/editor/Preferences.ts
+++ b/Script/AtomicEditor/editor/Preferences.ts
@@ -161,6 +161,10 @@ class Preferences {
         return this._prefs.editorBuildData;
     }
 
+    get editorFeature(): EditorFeature {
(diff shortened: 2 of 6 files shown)


Source/Atomic/Graphics/OpenGL/OGLGraphics.h
```diff
@@ -252,6 +252,9 @@ class ATOMIC_API Graphics : public Object
     /// Return graphics API name.
     const String& GetApiName() const { return apiName_; }
 
+    /// Return current driver version patch information 
+    const String& GetVersionPatch() const { return versionPatch_; }
+
```
JoshEngebretson: This is a good add

JoshEngebretson:
I'd like to see if can avoid having to touch preferences and multiple configuration spots, I *think* we might be able to get the driver version and change gl2/3 before the screen is opened, which would also solve the initial run issue and work with the deployed player... needed a reason to reinstall Linux VM :)

JimMarlowe:
Yes, this definitely falls into the barnacle code category.  Let me see if I can get the check operating in void Graphics::Restore() and make it fallback from GL3 to GL2, which it looks like it does for other reasons.

JoshEngebretson:
Thanks, I put this into chat, What do you think of: "we can probably default to GL3 initially, detect old mesa and recreate a GL2 with warning instructing to run with forcegl2 to avoid this warning"?

This would work first run and in the deployed player, I don't think releasing/recreating context should have a huge hit... though, that remains to be seen :)

JimMarlowe:
Closing this branch, most of these files are unneeded.

PR #911, opened 2016-07-07
Jm linux 863

JimMarlowe:
This PR fixes a common problem with the Linux Editor and player. Depending on the Mesa library version, a 3d scene, like ToonTown, becomes a flashing mess when viewed. Urho3D has the same issue. As a first impression, it's bad. The solution is to detect and compare mesa lib versions, and if it meets the error criteria, force the graphics to revert to GL2 mode.


JoshEngebretson:
Excellent :)  I squashed the commits and tweaked the log message a bit over in #912, will land from there.  Thanks!

PR #918, opened 2016-07-19
Add a parameter to genvs2015 task to enable/disable ATOMIC_DEV_BUILD

mattbenic:
This allows building the VS2015 solution without it having ATOMIC_DEV_BUILD set to 1. In our case this is useful for generating the solution for custom non-dev builds for artists. Will still default to 1.


JoshEngebretson:
Great, thanks

Issue #914, opened 2016-07-08
User Interface problems


JimMarlowe:
These are (subjectively) annoying all platform user interface problems. 

*Add tooltips to the top editor icons for translate, rotate, scale, and the mysterious LOCAL button.
*The edit menu has a scrollbar, get rid of it.
*Convert dark skin checkboxes images to light_skin style checkboxes.
*The window_close and window_close_pressed images need to be changed to something non-cartoony for both skin types.
*The attribute editor is using the grayed check button image, TBGreyCheckBox, it needs to use TBCheckBox. 
*There are embedded hex color codes in the source code and ui.tb.txt files, which are good for a dark backgrounds. We can create AccentColors1-n elements in the skin/skin.tb.txt, and program them with different values appropriate for the dark and light skins, and use them in the ui.tb.txt files. so each skin type will be able to look good.
*Need better icons for the player play and pause buttons, both use an arrow icon, which is confusing.
*The NFD_ChooseDirectory() does not receive the default directory string when it set in editor typescript code, so the user has to start looking for a directory starting at the current editor directory.

JimMarlowe:
Fixed in #920 and #951, except for Need better icons ... no better icons were found.

Issue #379, opened 2015-10-28
Implement a color picker


mattbenic:
Artist request:
*Add a color picker for editing colors

JustNukeIt:
In the Editor the color are usually display as 1 1 1 1 (R G B A), it will be nice to replace that box by a Square that is fill with the corresponding color (with the checker pattern to show the alpha). If a level designer click on the square it open a color picker with the different value (like in Photoshop for instance). see ref picture
![colorpickerref](https://cloud.githubusercontent.com/assets/11769017/12555176/8e9a90b8-c388-11e5-82c3-f6e427546065.jpg)

JoshEngebretson:
Thanks, we're planning on implementing a nice color picker with the new UIWebView component landing this week

raheelx:
Hi @JoshEngebretson :) 

did this end up getting landed?

JoshEngebretson:
Yup, the UIWebView component was landed that week.  I misread it this morning as the color picker was landing too, English is hard! :)

There currently isn't a color picker, I'd think there would be a nice web based one somewhere which would be relatively easy to grab and integrate using UIWebView.  Otherwise, maybe we can snag a color picker from somewhere and adapt it, like maybe MIT Torque3D?

brandonlamb:
@JoshEngebretson can this be implemented in TS? Looking through issues to see if there is some non-C++ thing I could attempt to contribute to. Implementing a color picker could be something, and I would greatly desire this feature myself

CTrauma:
is this available from the webview? https://samrueby.com/2014/10/10/did-you-know-chrome-has-its-own-color-picker/

shaddockh:
you would probably want a color picker in the inspector, right?  The web view currently is only used for the code editor.  Would it be possible to launch a small pop up web view from the inspector that exposes a color picker?  Or would that need to be rendered via turbo badger?

JimMarlowe:
There is a package, https://sourceforge.net/projects/tinyfiledialogs/, which is zlib licensed, and it provides a set of cross platform dialogs, one of which is a color picker. It is two files, and can be handled thru the FileUtils class like the nativefiledialogs. After that, adding it to the color inspector would be possible with a "..." button or other means.

JoshEngebretson:
[...]

PR #919, opened 2016-07-21
UI enhancements and color picker

JimMarlowe:
UI enhancements for #379 and #914

Added AccentColors elements to the light/dark skins, and used them to replace hex code colors in editor tb.txt files.
Replaced stock TB checkboxes in both skins with a blue square check box, fixed some instances when the UI files were referencing the grayed checkbox instead of the normal checkbox.
Replaced window and tab closer X in both skins with something better looking.
Added tooltips to the top editor icons for Translate, Rotate, Scale, and the Local/World button.
Created TBColorWidget and TBColorWheel widgets for use in the color picker
Added UISlider, UIColorWidget and UIColorWheel Atomic widgets, to make them scriptable.
Added a simple color picker to the color inspector.


JoshEngebretson:
Really great work, I went over the entire pull and looks good.  Nice job figuring out some of the more esoteric UI stuff, especially things like event blocking for changed notifications (so you don't end up filling up a ton of undo levels) and getting new widgets written and others wrapped :+1: 

My only comment is wondering if the dark theme should avoid using a bright-ish blue for checkboxes and sliders, I know there was some legacy usage of this from the original TB theme.  We can live in it for a bit and see what people think.  I like the changes to the close X :)  

I'll build out the branch here once I have a couple more things done for today, then land it.  THANKS!

JoshEngebretson:
A couple notes from testing:

I opened up a color chooser on global light color setting and chose red:

choosered

When I hit ok, the inline select fields of  the light color did update however the light component itself was not updated, this is probably due to blocking change events or an edit complete not getting caught, not sure.

notupdated

When a color chooser is open, clicking the '...' button will open additional choosers, it should maybe be disabled until the associated chooser closes.

multichoosers

JimMarlowe:
I'll make it so you cant bring up more than one color chooser at a time.
I was wondering if I should send an edit complete event, it looks like yes, I'll do that too.

JimMarlowe:
I still have to make it only bring up only one color chooser at a time, but I finally got the color to apply. Though it only took 4 additional lines, the path to get there was interesting to say the least. And I added HSL text, by request.

JoshEngebretson:
Great, I'll have a quick stab this morning to see if can get the edit event down to one, otherwise will be 4 levels of undo I believe.  It would also be good to get the '...' button justified to the right for consistency and to save vertical space

JimMarlowe:
These are skin changes for the checkbutton, radiobutton, and slider thumb. No more blue.

JoshEngebretson:
Very cool, if you merge https://github.com/JimMarlowe/AtomicGameEngine/pull/1 those changes will show up here too

JoshEngebretson:
Great work, made a PR with combined updates, will land from there :airplane:  : https://github.com/AtomicGameEngine/AtomicGameEngine/pull/920

Thanks!

PR #921, opened 2016-07-26
Introduces selective build configuration

mattbenic:
Allows a file called AssetBuildConfig.json to be added to settings, and tags within it used to specify what content to include when building. As an example:

{
  "AssetBuildConfig": [
    
   {
      "Tag": "FOO",
      "Resources": [
       "Components",
       "Scripts",
       "Scenes",
       "Sprites/Foo"
      ]
    },
   
   {
      "Tag": "BAR",
      "Resources": [
       "Components",
       "Scripts",
       "Scenes",
       "Sprites/Bar"
      ]
    }
  ]
}

This allows a build to be generated in three ways:
1) AtomicTool.exe build WINDOWS will generate a build as before, including all resources.
2) AtomicTool.exe build WINDOWS FOO will generate a build that includes only the sprites in Sprites/Foo.
3) AtomicTool.exe build WINDOWS BAR will generate a build that includes only the sprites in Sprites/BAR.

Any number of configurations can be described like this, and both complete filenames and directories are supported (both relative to /Resources


mattbenic:
@JoshEngebretson  Please let me know what the Mac/Linux fails are so we can sort them out.

JoshEngebretson:
It isn't getting to other platforms as the AtomicTool task is failing.  This task initially builds the tool so it doesn't have to be rebuilt for other platforms which use OSX to build (android, ios, webgl)

OSX and Linux are the same error:

```
Users/jenkins/Home/workspace/AtomicDS_BuildAtomicTool_Mac/Source/ToolCore/Build/BuildBase.cpp:518:21: error: non-const lvalue reference to type 'Atomic::String' cannot bind to a temporary of type 'Atomic::String'
            String &filename = GetFileName(filesInCacheFolder[j]);
```

This is easy to replicate and test locally if you have a Linux VM 

On Linux, the PR task just builds the editor, so could use the root shell script or:

```./Build/Linux/node/node ./Build/node_modules/jake/bin/cli.js -f ./Build/Scripts/Bootstrap.js build:atomiceditor --trace```

on Mac.

```./Build/Mac/node/node ./Build/node_modules/jake/bin/cli.js -f ./Build/CIScripts/Build.js build:atomictool```

JoshEngebretson:
Great work gents, will land in #922

PR #928, opened 2016-07-29
Added Max Size setting to Texture Importer

mattbenic:
This setting imposes the specified maximum dimensions on the compressed version of the texture in the cache.
Also fixes a case where the uncompressed file would have been loaded instead of the compressed version.

### DIFF
--- a/Source/Atomic/Resource/ResourceCache.cpp
+++ b/Source/Atomic/Resource/ResourceCache.cpp
@@ -371,8 +371,33 @@ bool ResourceCache::ReloadResource(Resource* resource)
 
     bool success = false;
     SharedPtr<File> file = GetFile(resource->GetName());
-    if (file)
-        success = resource->Load(*(file.Get()));
+	
+
+	if (file) 
+	{
+#ifdef ATOMIC_PLATFORM_DESKTOP
+		String ext = GetExtension(resource->GetName());
+		if (ext == ".jpg" || ext == ".png" || ext == ".tga")
+		{
+			String ddsName = "DDS/" + resource->GetName() + ".dds";
+			file = GetFile(ddsName, false);
+			if (file != NULL)
+				success = resource->Load(*(file.Get()));
+			else
+			{
+				file = GetFile(resource->GetName());
+				success = resource->Load(*(file.Get()));
+			}
+		}
+		else
+		{
+			success = resource->Load(*(file.Get()));
+		}
+#else
+		success = resource->Load(*(file.Get()));
+#endif
+	}
+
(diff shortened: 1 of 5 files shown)


Source/Atomic/Resource/ResourceCache.cpp
```diff
@@ -371,8 +371,33 @@ bool ResourceCache::ReloadResource(Resource* resource)
 
     bool success = false;
     SharedPtr<File> file = GetFile(resource->GetName());
-    if (file)
-        success = resource->Load(*(file.Get()));
+	
+
+	if (file) 
+	{
+#ifdef ATOMIC_PLATFORM_DESKTOP
```
JoshEngebretson: We'll refactor this, though good for now

Source/ToolCore/Assets/TextureImporter.cpp
```diff
@@ -71,7 +75,30 @@ bool TextureImporter::Import()
...
+			}
+
+			image->Resize(width*resizefactor, height*resizefactor);
+		}
+
+		if (image->SaveDDS(compressedPath))
+		{			
+			Renderer * renderer = GetSubsystem<Renderer>();
+			renderer->ReloadTextures();
+		}
```
JoshEngebretson: Just noting some whitespace issues, looking like tabs to me :)

JoshEngebretson:
Looks good, some of the code appears to have tab indents, I wonder if there is a CI plugin that will fail on mixed tabs/spaces in a source file :)  Until then, if it is the only issue in a PR, think landing trumps the turnaround unless it is a lot of code or a reoccurring theme.

Thanks, LANDING! :airplane:

mattbenic:
I'll refactor the ddr stuff in another PR, it bugs me having it spread out
like that as well, but that same change is in the other PR's code.

Will recheck whitespace settings as well on all devs, I think Johnny
recently had to reinstall. I see there's a VS plugin that supports that
project.js I pointed out earlier. Did you end up going for it?

JoshEngebretson:
Do you mean project.json?  NETCore is deprecating that in favor of cross platform msbuild solutions/projects, which is a good decision I think.  The AtomicEditor/ToolCore can also generate and build sln/csproj's now.

I could add a simple Jake task to check the sources in Source/ for mixed tabs/spaces, though no biggie, we've got bigger fish to fry :)

mattbenic:
Nope, I was referring to editorconfig. It has whitespace and EOL control
with plugins for all the major editors and IDEs:

http://editorconfig.org/

PR #935, opened 2016-08-01
NET Project Improvements (WIP)

JoshEngebretson:
This pull adds new project templates for pure C# applications, creates a default .user file for generated AtomicProject solution which allows immediately running project from VS, and some early NETProjectSystem hooks.  It also will open a C# file in project in VS if it is available on system


JoshEngebretson:
Oh, it also removes legacy JS editor stuff which has been replaced by Monaco

PR #948, opened 2016-08-10
fix setting PostUpdate event mask in JSComponent

EternalXY:
Fix setting PostUpdate event mask in JSComponent
Without this patch, disabling the PostUpdate event does not seem to work for JS components, and (according to the Chrome profiler for the Web version) this can add up to a significant % when there are a lot of nodes.


JoshEngebretson:
Thanks for the fix, can you add yourself to AUTHORS.md? 

https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/AUTHORS.md

@AtomicBuildBot Ok to test

JoshEngebretson:
Ok, see you added to authors in the other PR, I'll land this one. Thanks!

PR #953, opened 2016-08-11
provide some basic CEF cursor types in WebView; updated AUTHORS.md

EternalXY:
- Provide some mappings between SDL and CEF cursors over a WebView, so that the WebView feels more interactive/real.

- Update AUTHORS.md

### DIFF
--- a/Source/AtomicWebView/WebTexture2D.cpp
+++ b/Source/AtomicWebView/WebTexture2D.cpp
@@ -32,6 +32,7 @@
 #include <Atomic/Graphics/Graphics.h>
 #include <Atomic/Graphics/GraphicsImpl.h>
 #include <Atomic/Graphics/Technique.h>
+#include <SDL/include/SDL_mouse.h>
 
 #include "WebClient.h"
 #include "WebTexture2D.h"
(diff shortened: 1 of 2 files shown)


Source/AtomicWebView/WebTexture2D.cpp
```diff
@@ -74,6 +75,52 @@ class WebTexture2DPrivate : public CefRenderHandler
...
+       case CT_NOTALLOWED:
+           id = SDL_SYSTEM_CURSOR_NO;
+           break;
+       case CT_POINTER:
+           break;
+       default:
+           break;
+       }
+       c = SDL_CreateSystemCursor(id);
+       SDL_SetCursor(c);
```
JoshEngebretson: Thanks, I am sure this makes the integrated WebView feel far more natural.  A couple notes, I believe cursors will leak without corresponding SDL_FreeCursor calls (https://wiki.libsdl.org/SDL_FreeCursor).  The cursors should probably be pre-created, reused, and freed at shutdown.  It would also make some sense that this is an optional feature as may not be using the system cursor, for example in fullscreen mode.

EternalXY:
You are right, system cursors need to be freed in SDL, that would make this change a lot more intrustive than I had hoped.

Issue #356, opened 2015-10-21
Tabbing between inspector fields should highlight the entry


LaraEngebretson:
Currently when tabbing through inspector fields the mouse cursor gets placed in the left-most position. If the entire field is highlighted it would make it a lot faster to modify.

shaddockh:
Closed by #955

PR #955, opened 2016-08-12
UIEditFields auto-select contents on enter

### LINKED ISSUE #700: Editor functionality - Selection
When we select a value "box" (where we can changed the rotation of an object for example). If we want to enter a new value we have to select the current value then enter the new value, so if I want to set an object to the coordinate 0,0,0, I have to select the X "box" then Crtl+A enter the new value, then tab to go to the other box and have to Crtl+A then enter the value.

Could it be changed by default to select the current value in order to change it directly.

shaddockh:
Fixes #700 and #356 

This PR provides the following enhancements to the UIEditField
- If you tab into it from another field, the contents get auto selected so you can quickly type a new value
- If you click-select the field, the contents get auto selected so you can quickly type a new value
- If a field is already focused and auto selected and you click into it again, the auto-selection will disappear and the caret will appear where you clicked

Some issues I ran across:
It appears that the ```OnFocusedChange``` event in the UIEditField does not get called when the widget is getting created and when the ```autofocus:1``` value is specified in the ```tb.txt``` field.  The ```OnFocusedChange``` event **is** getting fired by the underlying TB widget, but may be getting triggered too early and the UIEditField doesn't capture it.  

To get around this, I made a modification to the new project dialog to remove the ```autofocus``` tag and programmatically set focus to the field in the code on creation.  This behaves as expected.  Let me know if this is an acceptable solution and I can look at other places where the auto-focus is being used or if there needs to be some low level research into why that event isn't getting bubbled up.

### DIFF
--- a/Resources/EditorData/AtomicEditor/editor/ui/renameasset.tb.txt
+++ b/Resources/EditorData/AtomicEditor/editor/ui/renameasset.tb.txt
@@ -6,10 +6,9 @@ TBLayout: axis: y, distribution: gravity, position: left
 	TBSeparator: gravity: left right, skin: AESeparator
 	TBLayout: gravity: left right, distribution-position: right bottom
 		TBTextField: text: "New Name:"
-		TBEditField: id: new_name, autofocus: 1
+		TBEditField: id: new_name
 			lp: min-width: 180
 	TBSeparator: gravity: left right, skin: AESeparator
-	TBLayout: 
+	TBLayout:
 		TBButton: text: Rename, id: rename
 		TBButton: text: Cancel, id: cancel
-		
--- a/Source/Atomic/UI/UIEditField.h
+++ b/Source/Atomic/UI/UIEditField.h
@@ -74,10 +74,9 @@ protected:
 
     virtual void OnFocusChanged(bool focused);
 
+private:
     // Used to keep track of if we have just been focused for the click select
     bool firstFocusFlag_;
-private:
-
 };
 
 }
--- a/Source/Atomic/UI/UIEditField.h
+++ b/Source/Atomic/UI/UIEditField.h
@@ -73,6 +73,9 @@ protected:
     virtual bool OnEvent(const tb::TBWidgetEvent &ev);
 
     virtual void OnFocusChanged(bool focused);
+
+    // Used to keep track of if we have just been focused for the click select
+    bool firstFocusFlag_;
 private:
 
 };
--- a/Source/Atomic/UI/UIEditField.h
+++ b/Source/Atomic/UI/UIEditField.h
@@ -72,6 +72,7 @@ protected:
 
     virtual bool OnEvent(const tb::TBWidgetEvent &ev);
 
+    virtual void OnFocusChanged(bool focused);
 private:
 
 };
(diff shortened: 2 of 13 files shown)


Source/Atomic/UI/UIEditField.h
```diff
@@ -72,6 +72,10 @@ class UIEditField : public UIWidget
 
     virtual bool OnEvent(const tb::TBWidgetEvent &ev);
 
+    virtual void OnFocusChanged(bool focused);
+
+    // Used to keep track of if we have just been focused for the click select
+    bool firstFocusFlag_;
```
JoshEngebretson: This could be a private member?   Also, I don't see that it is being initialized in the constructor so it will have an undefined value

JoshEngebretson:
Looks good!  I commented about the new member variable firstFocusFlag_ which needs to be initialized and can probably be private.  I think the focus change event workaround is ok, maybe file an issue for this though since TB is firing the event and our widget wrapper isn't catching it?

Source/Atomic/UI/UIEditField.h
shaddockh: Will do, I'll make it private and initialize it.  What's the protocol on bools in C though?  In JS and C# if a bool isn't set it defaults to false.

JoshEngebretson: C/C++ are low level and variables must be explicitly initialized otherwise they are whatever happens to be in memory, here is an example of doing so in the constructor (outside of the body which is preferred): https://github.com/AtomicGameEngine/AtomicGameEngine/blob/b3267990cc6724d60e5333f00e795e46533feede/Source/Atomic/UI/UI.cpp#L117

Resources/EditorData/AtomicEditor/editor/ui/renameasset.tb.txt
```diff
@@ -6,10 +6,9 @@ TBLayout: axis: y, distribution: gravity, position: left
 	TBSeparator: gravity: left right, skin: AESeparator
 	TBLayout: gravity: left right, distribution-position: right bottom
 		TBTextField: text: "New Name:"
-		TBEditField: id: new_name, autofocus: 1
+		TBEditField: id: new_name
```
JoshEngebretson: Darn, this commit adds a lot of these, it is really convenient to be able to set the autofocus flag in the declarative UI files and not have to script which to autofocus after the fact.  Going to need to look into this :)

JoshEngebretson: Thanks for filing https://github.com/AtomicGameEngine/AtomicGameEngine/issues/956

shaddockh: The only dialogs where this actually makes a difference is in snap settings, rename resource, and create project.  All the others don't populate the first field so auto select doesn't do anything.  If you want, I can back out the others and just leave these 3

JoshEngebretson: Yeah, let's do that since we'll be backing these out once we have a fix for the auto-focus widget event

shaddockh: Done.  Now only the Snap Settings, New Project, and Rename Resource dialogs are manually setting focus.

JoshEngebretson:
Awesome, thanks!  landing: :airplane:

PR #960, opened 2016-08-15
Generate TypeScript project files for external editors

Kobusvdwalt:
If you create a TypeScript project and you go to Developer>>Plugins>>TypeScript>>Generate External Editor Project the editor will create a tsconfig.json file in the root of the project and copy its Atomic.d.ts file into a folder "TypeScript" in the root of the project. This means that if you open the project in VS Code or Atom it will have autocomplete.

I am new to working on open source stuff so feel free to criticize as much as you want :)

### DIFF
--- a/Script/AtomicEditor/hostExtensions/languageExtensions/TypscriptLanguageExtension.ts
+++ b/Script/AtomicEditor/hostExtensions/languageExtensions/TypscriptLanguageExtension.ts
@@ -343,39 +343,22 @@ export default class TypescriptLanguageExtension implements Editor.HostExtension
 
     generateExternalProject ()
     {
-      var projectPath : string = ToolCore.toolSystem.project.projectPath;
+      var projectDir : string = ToolCore.toolSystem.project.projectPath;
 
       // Create TypeScript folder in project root
-      var tsPath : string = projectPath + "TypeScript/";
-      Atomic.getFileSystem().createDir(tsPath);
+      var projectDirTypeScript : string = Atomic.addTrailingSlash(projectDir + "TypeScript");
+      Atomic.getFileSystem().createDir(projectDirTypeScript);
 
       // Copy the Atomic.d.ts definition file to the TypeScript folder
       var toolDataDir : string = ToolCore.toolEnvironment.toolDataDir;
-      Atomic.getFileSystem().copy(toolDataDir + "/TypeScriptSupport/Atomic.d.ts", tsPath + "Atomic.d.ts");
+      var typescriptSupportDir : string = Atomic.addTrailingSlash(toolDataDir + "TypeScriptSupport");
+      Atomic.getFileSystem().copy(typescriptSupportDir + "Atomic.d.ts", projectDirTypeScript + "Atomic.d.ts");
 
       // Generate a tsconfig.json file
-      var config = new Atomic.File(projectPath + "tsconfig.json", Atomic.FILE_WRITE);
-      config.writeString
-(`
-{
-    "compilerOptions": {
-        "noEmitOnError": true,
-        "noImplicitAny": false,
-        "target": "ES5",
-        "module": "commonjs",
-        "outDir": "./Resources",
-        "rootDir": "./Resources",
-        "declaration": false,
-        "inlineSourceMap": false,
-        "removeComments": false,
-        "allowJs": true,
-        "noLib": true,
-        "allowNonTsExtensions": true,
-    }
-}
-`);
-
+      var config = new Atomic.File(projectDir + "tsconfig.json", Atomic.FILE_WRITE);
+      config.writeString(JSON.stringify(defaultCompilerOptions));
       config.close();
+
     }
     /**
      * Perform a full compile of the TypeScript
--- a/Script/AtomicEditor/hostExtensions/languageExtensions/TypscriptLanguageExtension.ts
+++ b/Script/AtomicEditor/hostExtensions/languageExtensions/TypscriptLanguageExtension.ts
@@ -345,19 +345,23 @@ export default class TypescriptLanguageExtension implements Editor.HostExtension
     {
       var projectDir : string = ToolCore.toolSystem.project.projectPath;
 
-      // Create TypeScript folder in project root
-      var projectDirTypeScript : string = Atomic.addTrailingSlash(projectDir + "TypeScript");
-      Atomic.getFileSystem().createDir(projectDirTypeScript);
+      // Create the typings folder in project root
+      var projectDirTypings : string = Atomic.addTrailingSlash(projectDir + "typings/main/ambient/atomicgameengine");
+      Atomic.getFileSystem().createDir(projectDirTypings);
 
-      // Copy the Atomic.d.ts definition file to the TypeScript folder
+      // Copy the Atomic.d.ts definition file to the typings folder
       var toolDataDir : string = ToolCore.toolEnvironment.toolDataDir;
       var typescriptSupportDir : string = Atomic.addTrailingSlash(toolDataDir + "TypeScriptSupport");
-      Atomic.getFileSystem().copy(typescriptSupportDir + "Atomic.d.ts", projectDirTypeScript + "Atomic.d.ts");
+      Atomic.getFileSystem().copy(typescriptSupportDir + "Atomic.d.ts", projectDirTypings + "Atomic.d.ts");
 
       // Generate a tsconfig.json file
-      var config = new Atomic.File(projectDir + "tsconfig.json", Atomic.FILE_WRITE);
-      config.writeString(JSON.stringify(defaultCompilerOptions));
-      config.close();
+      defaultCompilerOptions.noLib = false;
+      var tsconfigFile = new Atomic.File(projectDir + "tsconfig.json", Atomic.FILE_WRITE);
+      let tsconfig = {
+        compilerOptions: defaultCompilerOptions
+      };
+      tsconfigFile.writeString(JSON.stringify(tsconfig));
+      tsconfigFile.close();
 
     }
     /**
--- a/Script/AtomicEditor/hostExtensions/languageExtensions/TypscriptLanguageExtension.ts
+++ b/Script/AtomicEditor/hostExtensions/languageExtensions/TypscriptLanguageExtension.ts
@@ -165,6 +165,7 @@ export default class TypescriptLanguageExtension implements Editor.HostExtension
             this.compileOnSaveMenuItem = new Atomic.UIMenuItem(`Compile on Save: ${isCompileOnSave ? "On" : "Off"}`, `${this.name}.compileonsave`);
             menu.addItem(this.compileOnSaveMenuItem);
             menu.addItem(new Atomic.UIMenuItem("Compile Project", `${this.name}.compileproject`));
+            menu.addItem(new Atomic.UIMenuItem("Generate External Editor Project", `${this.name}.generateexternalproject`));
             this.menuCreated = true;
         }
     }
@@ -316,6 +317,9 @@ export default class TypescriptLanguageExtension implements Editor.HostExtension
                 case "compileproject":
                     this.doFullCompile();
                     return true;
+                case "generateexternalproject":
+                    this.generateExternalProject();
+                    return true;
             }
         }
     }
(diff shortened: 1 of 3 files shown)


Script/AtomicEditor/hostExtensions/languageExtensions/TypscriptLanguageExtension.ts
```diff
@@ -337,6 +341,42 @@ export default class TypescriptLanguageExtension implements Editor.HostExtension
         WebView.WebBrowserHost.setGlobalStringProperty("TypeScriptLanguageExtension", "tsConfig", JSON.stringify(tsConfig));
     }
 
+    generateExternalProject ()
+    {
+      var projectPath : string = ToolCore.toolSystem.project.projectPath;
+
+      // Create TypeScript folder in project root
+      var tsPath : string = projectPath + "TypeScript/";
```
shaddockh: You'll want to use ```Atomic.addTrailingSlash``` here so that it behaves on both all of the operating systems.

Script/AtomicEditor/hostExtensions/languageExtensions/TypscriptLanguageExtension.ts
```diff
@@ -337,6 +341,42 @@ export default class TypescriptLanguageExtension implements Editor.HostExtension
...
+      Atomic.getFileSystem().createDir(tsPath);
+
+      // Copy the Atomic.d.ts definition file to the TypeScript folder
+      var toolDataDir : string = ToolCore.toolEnvironment.toolDataDir;
+      Atomic.getFileSystem().copy(toolDataDir + "/TypeScriptSupport/Atomic.d.ts", tsPath + "Atomic.d.ts");
```
shaddockh: Another candidate for ```Atomic.addTrailingSlash```

Script/AtomicEditor/hostExtensions/languageExtensions/TypscriptLanguageExtension.ts
```diff
@@ -337,6 +341,42 @@ export default class TypescriptLanguageExtension implements Editor.HostExtension
...
+
+      // Generate a tsconfig.json file
+      var config = new Atomic.File(projectPath + "tsconfig.json", Atomic.FILE_WRITE);
+      config.writeString
+(`
+{
+    "compilerOptions": {
```
shaddockh: At the top of this module is an object called ```defaultCompilerOptions``` which is what is passed to the Monaco editor.  You should be able to use that and perform a ```JSON.stringify``` to write it to the file.

Script/tsconfig.json
```diff
@@ -109,15 +109,8 @@
...
-        "./TypeScript/AtomicApp.d.ts",
-        "./TypeScript/AtomicNETScript.d.ts",
-        "./TypeScript/AtomicPlayer.d.ts",
         "./TypeScript/AtomicWork.d.ts",
         "./TypeScript/duktape.d.ts",
-        "./TypeScript/Editor.d.ts",
-        "./TypeScript/EditorWork.d.ts",
-        "./TypeScript/ToolCore.d.ts",
-        "./TypeScript/WebView.d.ts"
+        "./TypeScript/EditorWork.d.ts"
```
shaddockh: It looks like some references got deleted here.

shaddockh:
Thanks! This looks pretty good.

I just scanned it over and left a few comments in the source.  In addition, one of the things that the ```TypeScriptLanguageExtension``` does when it loads up is it will look for a ```tsconfig.json``` file in the root of the project and use that instead of creating an in-memory one.  You will want to verify that the in-editor TypeScript completion and compilation still works when you add the new ```tsconfig.json``` file.

JoshEngebretson:

Script/AtomicEditor/hostExtensions/languageExtensions/TypscriptLanguageExtension.ts
JoshEngebretson: addTrailingSlash will add an internal slash, which is always "/", it will also normalize the path to internal representation:

https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/Atomic/IO/FileSystem.cpp#L1017

GetNativePath will convert a path string to use native "\" on Windows, though this is seldom necessary: https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/Atomic/IO/FileSystem.cpp#L1049

Script/AtomicEditor/hostExtensions/languageExtensions/TypscriptLanguageExtension.ts
JoshEngebretson: In practice "/" and "\" are pretty interchangeable on Windows, though if there is a problem we should modify methods like copy to use the native path representation, instead of making calling sites responsible for handling it

Script/tsconfig.json
JoshEngebretson: Yup, this is causing the CI build failure:

```
Script/AtomicEditor/hostExtensions/languageExtensions/TypscriptLanguageExtension.ts(341,9): error TS2304: Cannot find name 'WebView'.
Script/AtomicEditor/ui/frames/inspector/AssemblyInspector.ts(54,29): error TS2503: Cannot find namespace 'AtomicNETScript'.
Script/AtomicEditor/ui/frames/inspector/SelectionEditTypes.ts(64,21): error TS2503: Cannot find namespace 'AtomicNETScript'.
Script/AtomicEditor/ui/frames/inspector/SelectionEditTypes.ts(65,21): error TS2503: Cannot find namespace 'AtomicNETScript'.
Script/AtomicEditor/ui/frames/inspector/SelectionInspector.ts(181,20): error TS2503: Cannot find namespace 'AtomicNETScript'.
Script/AtomicEditor/ui/frames/inspector/SelectionInspector.ts(186,34): error TS2503: Cannot find namespace 'AtomicNETScript'.
Script/AtomicEditor/ui/frames/inspector/SelectionInspector.ts(198,20): error TS2503: Cannot find namespace 'AtomicNETScript'.
Script/AtomicEditor/editor/Editor.ts(163,9): error TS2304: Cannot find name 'WebView'.
Script/AtomicEditor/editor/Editor.ts(181,9): error TS2304: Cannot find name 'WebView'.
Script/AtomicEditor/editor/Editor.ts(204,13): error TS2304: Cannot find name 'WebView'.
Script/AtomicEditor/editor/Editor.ts(205,13): error TS2304: Cannot find name 'WebView'.
Script/AtomicEditor/ui/frames/inspector/CSComponentClassSelector.ts(27,59): error TS2503: Cannot find namespace 'AtomicNETScript'.
Script/TypeScript/Editor.d.ts(135,16): error TS2503: Cannot find namespace 'WebView'.
Script/TypeScript/Editor.d.ts(140,21): error TS2503: Cannot find namespace 'WebView'.
```

Kobusvdwalt:
Im not sure how to fix the refernces that were deleted. I cloned the original repo and it seems to be exactly the same.

shaddockh:
Thanks, this looks good. :+1:  I'll run it through a couple tests this evening and then it should be good to go.

@JoshEngebretson I'll create an issue to track and verify that all of the file calls can support native forward slashes on all platforms w/o the need for ```Atomic.addTrailingSlash```.  That would definitely clean up some code.

JoshEngebretson:
@Kobusvdwalt Good work, I would like to figure out a way to handle the tsconfig.json better, it is a source of shenanigans as editors like to change it, even if the contents are just reordered.  It can be a problem too if a script package isn't compiled on a specific platform, etc.   

@shaddockh Thanks for reviewing and testing :+1:

Script/AtomicEditor/hostExtensions/languageExtensions/TypscriptLanguageExtension.ts
```diff
@@ -337,6 +341,25 @@ export default class TypescriptLanguageExtension implements Editor.HostExtension
...
+      // Copy the Atomic.d.ts definition file to the TypeScript folder
+      var toolDataDir : string = ToolCore.toolEnvironment.toolDataDir;
+      var typescriptSupportDir : string = Atomic.addTrailingSlash(toolDataDir + "TypeScriptSupport");
+      Atomic.getFileSystem().copy(typescriptSupportDir + "Atomic.d.ts", projectDirTypeScript + "Atomic.d.ts");
+
+      // Generate a tsconfig.json file
+      var config = new Atomic.File(projectDir + "tsconfig.json", Atomic.FILE_WRITE);
+      config.writeString(JSON.stringify(defaultCompilerOptions));
+      config.close();
+
```
shaddockh: I took a look at the generated ```json``` file and the ```compiler options``` need to be under their own node.  So, something like:
```
let tsconfig = {
  compilerOptions: defaultCompilerOptions
};
config.writeString(JSON.stringify(tsconfig));
```
should work

shaddockh:
I tried this out with Atom and made sure that Atom with the TypeStrong TypeScript extension works as well as the built-in compiler.  In order for this to work, I made a note above about nesting the ```compilerOptions``` node under the root.  The other change that needs to be done to make this work is at the top of the file in the ```defaultCompilerOptions```, the property: ```noLib: true``` needs to be set to ```noLib: false```.  Doing this didn't affect the Atomic Game Engine, but w/o it, Atom complained about missing libraries.

The other note that I had was the location of the ```Atomic.d.ts```.  I'm seeing some movement in the TypeScript community to place these under a ```typings``` folder at the root of your project.  This is the location that https://github.com/typings/typings uses to store type definitions from various projects.

So maybe drop the ```Atomic.d.ts``` into ```typings/main/ambient/atomicgameengine/Atomic.d.ts```?  

This way the Atomic.d.ts file is in the right place and if you end up using the typings tool to bring in other libraries, they will all fall under the same folder structure.

Kobusvdwalt:
@shaddockh Cool thanks !!
To do : 
I will nest the compiler options under its own node.
I will change the noLib: true to noLib: false.
I will move the Atomic.d.ts into TypeScript/typings.

Its interesting that Atom complained. I also used TypeStrong and it works perfectly. I tested with VS Code too fwiw.

PS : Atom is so slow compared to VS Code but the overall functionality is much better.

shaddockh:
@Kobusvdwalt just a slight correct, I would suggest moving the Atomic.d.ts just under ```<project root>/typings``` to align with how standard TypeScript projects are structured.

JoshEngebretson:
<project root>/typings will make a nice sibling to <project root>/AtomicNET :+1:

Script/AtomicEditor/hostExtensions/languageExtensions/TypscriptLanguageExtension.ts
```diff
@@ -337,6 +341,29 @@ export default class TypescriptLanguageExtension implements Editor.HostExtension
         WebView.WebBrowserHost.setGlobalStringProperty("TypeScriptLanguageExtension", "tsConfig", JSON.stringify(tsConfig));
     }
 
+    generateExternalProject ()
+    {
```
shaddockh: Linter caught this.  The { needs to be on the same line as the function name.

Script/AtomicEditor/hostExtensions/languageExtensions/TypscriptLanguageExtension.ts
```diff
@@ -337,6 +341,29 @@ export default class TypescriptLanguageExtension implements Editor.HostExtension
...
+      var typescriptSupportDir : string = Atomic.addTrailingSlash(toolDataDir + "TypeScriptSupport");
+      Atomic.getFileSystem().copy(typescriptSupportDir + "Atomic.d.ts", projectDirTypings + "Atomic.d.ts");
+
+      // Generate a tsconfig.json file
+      defaultCompilerOptions.noLib = false;
+      var tsconfigFile = new Atomic.File(projectDir + "tsconfig.json", Atomic.FILE_WRITE);
+      let tsconfig = {
+        compilerOptions: defaultCompilerOptions
+      };
+      tsconfigFile.writeString(JSON.stringify(tsconfig));
```
shaddockh: Sorry didn't catch this before, but the JSON for this file should probably be pretty printed so it's easy to edit.  That can be done with:
```JSON.stringify(tsconfig, null, 4)```  which tells the JSON process to format each entry correctly using 4 space tabs.

shaddockh:
Just a couple final cosmetic things that need to be fixed and this is ready to land.  
- lint error
- pretty print the output of the JSON

I was able to load this up directly into Atom and compile w/o issue.  For VSCode, I needed to set up a VSCode task in order to do this.  That's all documented on other sites so that may just be an exercise for the user (or an entry in the Q&A site)

Good work!

Kobusvdwalt:
@shaddockh I am not sure what you meant by lint error. Could you elaborate ?
I am done with the pretty print.

shaddockh:
We use a linter to catch a handful of errors before you start running the code as well as to retain consistency in the format of the code.  You can see: https://palantir.github.io/tslint/ for the tool that is used during the build process.

The lint error I was referencing is on line 344/345 in reference to the curly brace not being on the same line as the function.  If you scroll up in this conversation, you'll see where I called it out in one of the code line comments.

shaddockh:
Excellent!  Looks good.

Landing.

PR #965, opened 2016-08-16
Joystick Rumble support 954

JimMarlowe:
This PR adds rumble support for joysticks and game controllers. It also enables user configuration of unhandled game controllers to make them behave in a standardized way.
The rumble support is self initializing, the user does not have to make any calls to use the rumble. There is one convenience call to ask if rumble is supported on the joystick/controller. The main call is :
void JoystickRumble(unsigned int id, float strength, unsigned int length); /// produce rumble
    The id argument is what gets passed into the event handlers as the JoystickId, not the index, though they are often the same.
    The strength argument is a float from 0.0 to 1.0 and is the magnitude of the rumble event.
    The length argument is the number of milliseconds to play the rumble event.

The test program and example usage are delayed until the joystick event handlers stabilize.

### DIFF
--- a/Source/Atomic/Input/Input.cpp
+++ b/Source/Atomic/Input/Input.cpp
@@ -338,7 +338,7 @@ bool JoystickState::StartRumble()
         haptic_ = SDL_HapticOpen(joystickID_);
         if (haptic_ == NULL)
         {
-            ATOMIC_LOGERRORF( "Unable to access haptic device[%d] %s", joystickID_, hname );
+            ATOMIC_LOGERRORF( "Unable to access haptic device[%d] %s", joystickID_, hname ? hname : "Unknown" );
             canRumble_ = false;
             return false;
         }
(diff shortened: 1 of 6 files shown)


Source/Atomic/Input/Input.cpp
```diff
@@ -1510,6 +1595,13 @@ void Input::Initialize()
 
     graphics_ = graphics;
 
+// ATOMIC BEGIN
+    //  get user gamecontroller configs for SDL
+    FileSystem* fs = GetSubsystem<FileSystem>();
+    const String configName = fs->GetUserDocumentsDir() + "/gamecontrollerdb.txt"; 
+    SDL_GameControllerAddMappingsFromFile(configName.CString());
+// ATOMIC END
```
JoshEngebretson: Should we file a followup issue to include the community controller database (https://raw.githubusercontent.com/gabomdq/SDL_GameControllerDB/master/gamecontrollerdb.txt) into the AtomicEditor data files, so it is guaranteed to be available and can be copied to deployed applications?

JoshEngebretson:
Great work, very clean 👍  

I have a note about a possible follow up, though LANDING! ✈️

PR #971, opened 2016-08-19
Update to latest version of TypeDoc

shaddockh:
The issue that wasn't allowing the TypeDoc documentation to be generated with a custom theme has been resolved ( https://github.com/TypeStrong/typedoc/issues/227 ) 

This change does the following
- Allows for the latest TypeDoc to be pulled which uses the latest TypeScript
- adds a fix for the latest NPM so that the node_modules get put in the right place


JoshEngebretson:
Great, would be good to have master docs available too for perusal online.  Just filing that in consciousness for now, LANDING ✈️

Issue #944, opened 2016-08-06
[Linux] Segmentation fault when running project


McFizh:
When I create new 2D project and try to run it, the app shows for a second, then atomic crashes.

Platform: Fedora Core 24 , 64-bit
Kernel: 4.6.4-301.fc24.x86_64
Gcc: gcc version 6.1.1 20160621 (Red Hat 6.1.1-3) (GCC)

Git HEAD hash: e7255df3ce18bb17bf1f805a83aa2383873e6df7

dmesg output:
  traps: AtomicEditor[29909] general protection ip:d926dc sp:7f39155faa80 error:0 in AtomicEditor[400000+1bf9000]
  AtomicEditor[29908]: segfault at 28 ip 0000000000c21940 sp 00007ffc9a8a4f98 error 4 in AtomicEditor[400000+1bf9000]

(gdb) where
#0  0x0000000000d926dc in Atomic::VectorBuffer::Write(void const*, unsigned int) ()
#1  0x0000000000e76be2 in Atomic::IPCChannel::Receive() ()
#2  0x0000000000e77988 in Atomic::IPCBroker::ThreadFunction() ()
#3  0x0000000000cf4a0a in Atomic::ThreadFunctionStatic(void*) ()
#4  0x00007ffff7bc45ca in start_thread (arg=0x7fff65ad2700) at pthread_create.c:333
#5  0x00007ffff072cead in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:109

And this is what is printed on console:

[Fri Aug  5 17:34:34 2016] INFO: Opened log file /home/user/.local/share/AtomicEditor/Logs/AtomicEditor.log
[Fri Aug  5 17:34:34 2016] INFO: Created 3 worker threads
[Fri Aug  5 17:34:34 2016] INFO: Added resource path /home/user/AtomicGameEngine/Artifacts/AtomicEditor/Resources/CoreData/
[Fri Aug  5 17:34:34 2016] INFO: Added resource path /home/user/AtomicGameEngine/Artifacts/AtomicEditor/Resources/EditorData/
[Fri Aug  5 17:34:34 2016] DEBUG: Skipped autoload path 'Autoload' as it does not exist, check the documentation on how to set the 'resource prefix path'
[Fri Aug  5 17:34:34 2016] INFO: Set screen mode 3640x1023 windowed resizable
[Fri Aug  5 17:34:34 2016] INFO: Initialized input
[...]

Issue #713, opened 2016-03-18
[LINUX] Segmentation fault when running some of the sample projects


guzzard:
Opening Basic2D and BunnyMark example projects works fine, but they crash on play with "Segmentation fault (core dumped)". Before the crash I'm able to see some moving graphics for half a second.

Running Arch Linux. Example projects cloned from https://github.com/AtomicGameEngine/AtomicExamples

guzzard:
I built wiuth "set(CMAKE_BUILD_TYPE Debug)" and attempted to get more information about the segfault with gdb. Strange thing is when running Basic2D or BunnyMark with the build debug flag set the editor behaves different. The segfault doesn't occur, but when stopping the player either by closing the window or using the stop button it seems like the thread handling the player doesn't exit properly, because the stop button doesn't change to a play button, and it's not possible to start the player again. When attempting to start the player again a SIGPIPE occurs as visible in the log below.

(gdb) exec-file /home/username/temp/AtomicGameEngine/Artifacts/AtomicEditor/AtomicEditor -log debug
(gdb) r
Starting program: /home/username/temp/AtomicGameEngine/Artifacts/AtomicEditor/AtomicEditor
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".
[Sat Mar 19 22:35:25 2016] INFO: Opened log file /home/username/.local/share/AtomicEditor/Logs/AtomicEditor.log
[New Thread 0x7fffe47b1700 (LWP 21713)]
[New Thread 0x7fffe3fb0700 (LWP 21714)]
[New Thread 0x7fffe37af700 (LWP 21715)]
[Sat Mar 19 22:35:25 2016] INFO: Created 3 worker threads
[Sat Mar 19 22:35:25 2016] INFO: Added resource path /home/username/temp/AtomicGameEngine/Artifacts/AtomicEditor/Resources/CoreData/
[Sat Mar 19 22:35:25 2016] INFO: Added resource path /home/username/temp/AtomicGameEngine/Artifacts/AtomicEditor/Resources/EditorData/
[Sat Mar 19 22:35:25 2016] DEBUG: Skipped autoload path 'Autoload' as it does not exist, check the documentation on how to set the 'resource prefix path'
[Sat Mar 19 22:35:25 2016] INFO: Set screen mode 2360x1475 windowed resizable
[Sat Mar 19 22:35:25 2016] INFO: Initialized input
[...]

PR #982, opened 2016-08-24
Fix linux crash on player exit

JimMarlowe:
This pr fixes the editor crash when the player is exited. It was tracked down to an optimization issue with g++ -O3 and VectorBuffer::Write() in a particular situation. This call was changed to VectorBuffer::SetData(), which will perform the same action for filling a new VectorBuffer, and does not incur the -O3 optimization problem.


JoshEngebretson:
Awesome, thanks for digging into this!  Closes #944, #713

Issue #480, opened 2015-12-01
Update camera rotation in scene view


LaraEngebretson:
* Allow the camera to rotate in the scene view infinitely. If you continue pushing the cursor along the border of the editor, while holding down the right mouse button, the camera should keep rotating.

* We may want to switch the mouse cursor to another icon

JoshEngebretson:
There is a bandaid for this in #987, closing issue as we don't currently have multiple/positonal edit views

PR #987, opened 2016-09-01
Fix issue #480 Allow the camera to rotate in the scene view infinitely

benwolf:
Right click on viewport will hide the cursor and allow the camera to continue to rotate. Fix issue #480


JoshEngebretson:
Easy peasy :)  This works for now as the scene view is always centered, though once we have multiple scene views and can position them, we'll have to change it up.  Thanks!

Issue #991, opened 2016-09-02
getJSComponent fails if called immediately after createChildPrefab


bitonator:
Creating this issue to submit a PR

I added a function called `callme` in `Coin.js` in the platformer example, and then added this in level.js

```
var node = self.scene.createChildPrefab("Coin", "Prefabs/Coin.prefab");
var script=node.getJSComponent("Coin");
script.callme();
```

This fails with a message that `callme` is undefined.

Looked through the code, looks like when a prefab is loaded `JSComponent::InitInstance` is called only in `JSComponent::Update`, so the instance isn't initialized immediately after `createChildPrefab`

JoshEngebretson:
Thanks, closing

PR #992, opened 2016-09-02
Issue #991 getJSComponent fails if called immediately after createChildPrefab

bitonator:
added an `JSComponent::IsInstanceInitialized` to return if an instance is initialized

added code in `JSScene::Node_getJSComponent` to check if `JSComponent` is initialized and if not, call `JSComponent::InitInstance`


JoshEngebretson:

JoshEngebretson:
Thanks for the pull request, landing :+1:

PR #1007, opened 2016-09-13
Fix components in separate dlls being available

mattbenic:
Also exposed skeleton nodes as previously discussed

### DIFF
--- a/Script/AtomicNET/AtomicNET/Scene/CSComponentCore.cs
+++ b/Script/AtomicNET/AtomicNET/Scene/CSComponentCore.cs
@@ -332,6 +332,21 @@ namespace AtomicEngine
 
         }
 
+        void HandleComponentAssemblyReference(uint eventType, ScriptVariantMap eventData)
+        {
+#if ATOMIC_DESKTOP || ATOMIC_ANDROID
+            string assemblyPath = eventData["AssemblyPath"];
+
+            string assemblyName = Path.GetFileNameWithoutExtension(assemblyPath);
+            if (componentCache.ContainsKey(assemblyName))
+                return;
+
+            Assembly assembly = Assembly.LoadFrom(assemblyPath);
+
+            ParseAssembly(assembly);
+#endif
+        }
+
         void ParseComponents()
         {
 #if ATOMIC_DESKTOP || ATOMIC_ANDROID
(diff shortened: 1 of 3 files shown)


Source/Atomic/Graphics/AnimatedModel.h
```diff
@@ -99,6 +99,8 @@ class ATOMIC_API AnimatedModel : public StaticModel
 
     /// Return skeleton.
     Skeleton& GetSkeleton() { return skeleton_; }
+    /// Return the node of a skeleton bone (for script access)
+    Node* GetSkeletonBoneNode(const String& boneName);
```
JoshEngebretson: When modifying core Urho sources, it is good to enclose new methods in a // ATOMIC BEGIN and // ATOMIC END block : https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/Atomic/Graphics/Animation.h#L181

I also have been doing this in the implementation side: https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/Atomic/Graphics/Animation.cpp#L398

... and trying to be consistent when modifying code, also trying to block/comment.  

This greatly helps when pulling in updates, I can get this in post

mattbenic: Right, will keep an eye out in future

Script/AtomicNET/AtomicNET/Scene/CSComponentCore.cs
```diff
@@ -332,6 +332,21 @@ void HandleComponentLoad(uint eventType, ScriptVariantMap eventData)
 
         }
 
+        void HandleComponentAssemblyReference(uint eventType, ScriptVariantMap eventData)
+        {
+#if ATOMIC_DESKTOP || ATOMIC_ANDROID
+            string assemblyPath = eventData["AssemblyPath"];
```
JoshEngebretson: Probably don't need this on Android?  Can leave it if it isn't hurting anything...

mattbenic: Kept it because the others were Android included. If there are any additional dlls in an Android project, would also be needed there?

JoshEngebretson: I am pretty sure we're going to need to make design/edit time systems for this... at runtime, we should ensure all the types are loaded, ahead of being referenced.  Probably not the last time this code will be touched :)

JoshEngebretson:
Looks good, landing 👍

PR #1015, opened 2016-09-16
Add --with-docs, --with-examples to build

JimMarlowe:
Add --with-docs, --with-examples to build, all three platforms

### DIFF
--- a/Build/Scripts/Bootstrap.js
+++ b/Build/Scripts/Bootstrap.js
@@ -26,7 +26,6 @@ function printHelp() {
     console.log("--debug         : Build debug version of the editor and associated platform runtimes")
     console.log("--noclean       : Do not clean before building, useful during development")
     console.log("--nonet         : Build without AtomicNET C# scripting support")
-    console.log("--with-docs     : Build and install API documents into the editor")
     console.log("--with-examples : Install examples into the editor")
     console.log("--------------------------")
 
--- a/Build/Scripts/Bootstrap.js
+++ b/Build/Scripts/Bootstrap.js
@@ -26,6 +26,7 @@ function printHelp() {
     console.log("--debug         : Build debug version of the editor and associated platform runtimes")
     console.log("--noclean       : Do not clean before building, useful during development")
     console.log("--nonet         : Build without AtomicNET C# scripting support")
+    console.log("--with-docs     : Build and install API documents into the editor")
     console.log("--with-examples : Install examples into the editor")
     console.log("--------------------------")
 
--- a/Build/Scripts/Bootstrap.js
+++ b/Build/Scripts/Bootstrap.js
@@ -20,12 +20,14 @@ function printHelp() {
 
     console.log("\nAtomic Editor Build Script")
     console.log("--------------------------")
-    console.log("--help         : This help text")
-    console.log("--with-android : Build with Android platform support");
-    console.log("--with-ios     : Build with iOS platform support");
-    console.log("--debug        : Build debug version of the editor and associated platform runtimes")
-    console.log("--noclean      : Do not clean before building, useful during development")
-    console.log("--nonet        : Build without AtomicNET C# scripting support")
+    console.log("--help          : This help text")
+    console.log("--with-android  : Build with Android platform support");
+    console.log("--with-ios      : Build with iOS platform support");
+    console.log("--debug         : Build debug version of the editor and associated platform runtimes")
+    console.log("--noclean       : Do not clean before building, useful during development")
+    console.log("--nonet         : Build without AtomicNET C# scripting support")
+    console.log("--with-docs     : Build and install API documents into the editor")
+    console.log("--with-examples : Install examples into the editor")
     console.log("--------------------------")
 
     process.exit(0);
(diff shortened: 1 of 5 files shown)


Build/Scripts/Bootstrap.js
```diff
@@ -20,12 +20,14 @@ function printHelp() {
...
-    console.log("--nonet        : Build without AtomicNET C# scripting support")
+    console.log("--help          : This help text")
+    console.log("--with-android  : Build with Android platform support");
+    console.log("--with-ios      : Build with iOS platform support");
+    console.log("--debug         : Build debug version of the editor and associated platform runtimes")
+    console.log("--noclean       : Do not clean before building, useful during development")
+    console.log("--nonet         : Build without AtomicNET C# scripting support")
+    console.log("--with-docs     : Build and install API documents into the editor")
+    console.log("--with-examples : Install examples into the editor")
     console.log("--------------------------")
```
JoshEngebretson: Awesome :)

Build/Scripts/BuildCommon.js
```diff
@@ -207,5 +213,61 @@ namespace('build', function() {
...
+ 
+  task('gendocs', {
+    async: true
+    }, function() {
+
+    fs.copySync(atomicRoot + "Build/Docs/Readme.md", jsDocFolder + "Readme.md");
+    fs.copySync(atomicRoot + "Build/Docs/atomic-theme", jsDocFolder + "atomic-theme");
+
+    cmds = [
+      "cd " + jsDocFolder + " && echo {} > package.json", // newer versions of npm require package.json to be in the folder or else it searches up the heirarchy
```
JoshEngebretson: Does this work on Windows?

Build/Scripts/BuildCommon.js
```diff
@@ -207,5 +213,61 @@ namespace('build', function() {
...
+      "cd " + jsDocFolder + " && npm install typedoc",
```
JoshEngebretson: We don't have a local npm so would be adding a dependency to build, not sure on typedoc's size, maybe it can go in: https://github.com/AtomicGameEngine/AtomicGameEngine/tree/master/Build/node_modules with other node build modules?

Build/Scripts/BuildCommon.js
```diff
@@ -207,5 +213,61 @@ namespace('build', function() {
...
+  
+  task('genexamples', {
+    async: true
+    }, function() {
+    
+    common.cleanCreateDir( toolDataFolder + "AtomicExamples");
+
+    cmds = [
+      "git clone https://github.com/AtomicGameEngine/AtomicExamples " + toolDataFolder + "AtomicExamples && rm -rf " + toolDataFolder + "AtomicExamples/.git",
+    ];
```
JoshEngebretson: I don't think rm -rf will fly on Windows :)

JoshEngebretson:
Nice and clean PR :) 

Needs some tweaks for windows, though that is ok for now.  The main thing is introducing npm as a dependency, which would only be for generating docs.  I think this is better than introducing typedoc to node_modules... we can see what kind of mileage we get out of it.  

I'll land when green!  Thanks! 🐫

Build/Scripts/BuildCommon.js
JimMarlowe: I'll factor the rm out, it can be done with fs calls instead.

Build/Scripts/BuildCommon.js
JimMarlowe: I'll look at that.

Build/Scripts/BuildCommon.js
JimMarlowe: I dont know, but if there is a node available there is no reason it shouldn't. If not, it can removed from the windows build.

JimMarlowe:
I've backed out the --with-doc option for now, I believe its possible to get typedoc into node_modules, but it will take more investigation on how to do it without npm.

JoshEngebretson:
I think having a npm dependency for docs generation is fine for now, iterative :)  

I want to generate some site binaries soonish which include docs, CI boxes have npm... so will reintroduce --with-docs using npm anyway and probably better than adding more to node_modules, unless typedoc (and whatever dependencies it has) is pretty small.

JimMarlowe:
The --with-docs is back in, and the --with examples is fixed.

JoshEngebretson:
The CI failure is due to a code signing issue on the build box.  

Thanks!  LANDING! ✈️

JimMarlowe:
Is good. What I found with typedoc, is that when npm installs it, it is 25Meg, although 18Meg of it is a copy of typescript. I downloaded typedoc master, and it is only 1Meg+, to move it into node_modules may require a couple more packages like handlebars, but the big item, typescript, is already in there. For someone who know what they are doing, like @shaddockh it's probably an "easy" task to move typedoc in to node_modules, but for me it'll take a little longer.

JoshEngebretson:
Great, thanks for R&D'ing it :)  

npm is a sane thing to require on box to support generation of documentation, for C#/C++ docs we will have other requirements, which likely won't be included in repo either.  So, let's stick with requiring npm I'd say

PR #1022, opened 2016-09-19
Basket of Unamables

JimMarlowe:
Sorry, but I removed "Early Access", it was time, unless you are under a contractual agreement to show it? I can add a banner into the generic image, if you like.  
I had a pile of other details, cosmetic and code-wise that would be good to do before a release.


JoshEngebretson:
Nice!  That is just enough lensflare and thanks for reintroducing  the pause/step functionality I clobbered :)  

One thing on the examples, some of these should be hidden from the example browser as they are tests or outdates, probably a good idea to move them around in the repo itself so that is clear.  Also, need to make sure the length of desc's isn't causing a formatting issue.

This reminds me, I want to get this into NFD in hopes it fixes the dialog focus issue once and for all: https://github.com/mlabbe/nativefiledialog/commit/6527087b1e6a37b7ea228bf79ec0f319c731b4aa 

I have never personally seen it :/

JoshEngebretson:
BTW, I'll comb through the example json, also need to detect if C# is supported on system and not show those if it isn't.  This comes down to having Visual Studio/Xamarin Studio installed, I think we're just going to require those for C#... it is possible to compile C# with vanilla mono and the integrated Monaco code editor, though that is just getting to be too many paths for us to maintain (and you really do want an IDE for C#)

JimMarlowe:
Push to C# - Does that mean the Atomic editor wouldn't/shouldn't play a c# project? Then now is the time for a web page to explain the Atomic C# tool chain, how to setup, install, code, run and deploy.

JoshEngebretson:
Yes, VS/XS are required for C#, in a future build we can probably include the necessary compilation bits, as long as NET framework/Mono on system.  In general though, really want to be using VS/XS when working with C# anyway for IDE, Debugging, and both are required for deployment to Android/iOS and when the C# WebAssembly stuff comes together, the web.

The editor does need to detect this and we can pop up a help/info window, a Q & A link would be good too, and we need to address docs in general.  

Pushing for a "release", we need to have a couple smoke tests first :)  I still have a couple loops to close on the C# side for new projects/examples, wrt mobile support mostly

JoshEngebretson:
... aaaand LANDING! Thanks! ✈️

PR #1031, opened 2016-09-24
Update Typescript to 2.0.3

### LINKED ISSUE #1030: Upgrade TypeScript support to 2.0
Both the editor and the built-in TypeScript compiler should be updated to the newly released 2.0.  Currently there are some dependency issues with the  dts-generator library that are blocking the upgrade.

Research needs to be done to determine if the functions of dts-generator can be replicated in vanilla TypeScript during the build process, or wait until dts-generator supports TS2.0.  Also need to tune the compiler settings for the Editor as well as the built-in compiler to take advantage of the new TS typechecking and code analysis features.

shaddockh:
Fixes #1030

This updates the editor to be compiled with TypeScript 2.0.3
Other changes:
- Built-in web view intellisense now uses TS 2.0.3
- Built-in TS project compiler uses TS 2.0.3
- Modified TS Projects to look for *.d.ts files anywhere in the active project structure, not just under Resources
- Modified TS Projects to enforce consistent filename casing when importing (requiring) other modules

Open issue:
```dts-generator``` complains about an unmet peer dependency on an old version of typescript but it doesn't appear to break anything.  Once ```dts-generator``` gets updated, this should go away.


JoshEngebretson:
Awesome, LANDING! Thanks :)

PR #1054, opened 2016-09-28
Sound.setData in JavaScript

Type1J:
First off: Sorry! I know that I fell off the face of the planet. I'm hoping to start working more on Atomic soon. Here's something that I needed for one of my projects. I have an example for it that I'll PR once this goes through.


JoshEngebretson:
No problem, welcome back :)

JoshEngebretson:
Great, one thing that helps when adding custom native JS methods is to add a TypeScript decl for the method, this way it will show up in the tsdocs (which are useful for vanilla JS as well), for example: https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Script/Packages/Atomic/Graphics.json#L40

The decl would go in Audio.json in this case

PR #1062, opened 2016-09-30
Linux package generation

JimMarlowe:
This will create a linux debian package from the ./Build_AtomicEditor.sh  --package command.


Build/Scripts/BuildLinux.js
```diff
@@ -23,6 +23,9 @@ function copyAtomicNET() {
 
 function copyAtomicEditor() {
 
+    if ( config["noclean"])
+        host.cleanCreateDir(editorAppFolder);
```
JoshEngebretson: The noclean option is primarily intended for building the binaries, in general probably want a clean folder for the app copy.  In either case, should this be ```if ( !config["noclean"] )``` ?

Build/Scripts/PackageEditor.js
```diff
@@ -96,5 +97,68 @@ namespace('package', function() {
...
+        //copy in menu pixmap
+        fs.copySync(config.atomicRoot + "Build/Linux/atomic_menu.xpm", dstDir + "usr/share/AtomicGameEngine/atomic_menu.xpm" );
+        
+        // get rid of some lintian errors
+        fs.removeSync( dstDir + "usr/share/AtomicGameEngine/Resources/ToolData/.gitignore");
+        fs.removeSync( dstDir + "usr/share/AtomicGameEngine/Resources/ToolData/CodeEditor/.gitignore");
+        fs.removeSync( dstDir + "usr/share/AtomicGameEngine/Resources/ToolData/Deployment/Android/assets/.gitignore");
+        fs.removeSync( dstDir + "usr/share/AtomicGameEngine/Resources/ToolData/ProjectTemplates/.gitignore");
+        fs.removeSync( dstDir + "usr/share/AtomicGameEngine/Resources/ToolData/TypeScriptSupport/.gitignore");
+
```
JoshEngebretson: Does removeSync or some other variation do globbing? So we could do something like **/.gitignore instead of specifying files?

Build/Linux/DEBIAN/control
```diff
@@ -0,0 +1,14 @@
+Package: AtomicGameEngine
+Version: 0.0.8
+Installed-Size: 393064
+Section: graphics
+Maintainer: pincushion <info@atomicgameengine.com>
+Homepage: http://atomicgameengine.com
```
JoshEngebretson: Who is pincushion? :)

Build/Linux/DEBIAN/control
```diff
@@ -0,0 +1,14 @@
+Package: AtomicGameEngine
+Version: 0.0.8
+Installed-Size: 393064
+Section: graphics
```
JoshEngebretson: I don't suppose there is a game engine or editor section? :)

Build/Linux/atomic_menu.xpm
```diff
@@ -0,0 +1,123 @@
...
+"   ...<[}|1234+&&+4321|}[<...   ",
+"   ....~{]^/(_:..:_(/^]{~....   ",
+"    ........!)'&&')!........    ",
+"    ........&;=--=;&........    ",
+"     ........#$%%$#........     ",
+"      ........+@@+........      ",
+"        ................        ",
+"          ............          ",
+"            ........            ",
+"                                "};
```
JoshEngebretson: xpm for  the win

Build/Scripts/PackageEditor.js
```diff
@@ -96,5 +97,68 @@ namespace('package', function() {
...
+        cmds = [];
+        
+        // go to staging root directory
+        cmds.push("cd " + config.artifactsRoot + " ;" );
+        
+        // get rid of some more lintian errors
+        cmds.push("/usr/bin/strip --strip-unneeded " + dstDir + "usr/share/AtomicGameEngine/AtomicEditor ;" );
+        cmds.push("/usr/bin/strip --strip-unneeded " + dstDir + "usr/share/AtomicGameEngine/Resources/ToolData/Deployment/Linux/AtomicPlayer ;" );
+        cmds.push("/bin/chmod oug-wx " + dstDir + "usr/share/AtomicGameEngine/libcef.so ;");
+
```
JoshEngebretson: oug-wx! Good jump massaging the binaries 👍

Build/Scripts/BuildLinux.js
JimMarlowe: when I tried the --noclean option, it faulted because the fs.copySync won't overwrite an existing item, Do the other platform --noclean options work this way?

Build/Scripts/PackageEditor.js
JimMarlowe: I dont know if there is a globbing, this is part of making fewer lintian errors and warnings. Calling out specific files in another repo is fragile at best, I have a plan to write a program that will take the output of lintian and fix the files in the staging area, I don't know why one doesn't already exists, a new opportunity I guess.

Build/Linux/DEBIAN/control
JimMarlowe: Its a toss up between Programming or Graphics. I picked Graphics because that's where blender and other 3d modelling and art programs ended up. Its a one word change, if you want it moved.

Build/Linux/DEBIAN/control
JimMarlowe: pincushion is a name we used to give to novice quake players, if you'd like a real maintainer name that easily fixed too. Call it plausible denial-ability ... "blame pincushion, he b0rked it all up."

JimMarlowe:
The intention is to allow you to provide a .deb on the download page in the place of the grayed out Linux Editor, in addition to users making their own packages.

To install via command line:

sudo dpkg -i AtomicGameEngine_0.0.8_amd64.deb
sudo apt-get -f install

Build/Scripts/BuildLinux.js
JoshEngebretson: Ok, I see,  fs,copySync should take a clobber option which will overwrite: https://github.com/jprichardson/node-fs-extra#copy though it would be nice to not have to specify that every call :/

I think we should change all the builds to clean the editor app folder (and I think some checks needs to be put in for noclean + doc/example generation), so let's move this folder creation down to where the host.setupDirs is called, removing it from the provided lists to that and do a clean create regardless... will follow up with other builds

Build/Linux/DEBIAN/control
JoshEngebretson: Let's stick with Graphics, programming is boring! ;)

Build/Linux/DEBIAN/control
JoshEngebretson: Ok, as long as there is rationale for the history books, pincushion is great

JoshEngebretson:
That's cool, we'll stick the deb install command docs alongside the download, I might have to update the linux build box, I guess we'll see... is code signing a deb a common practice?

JimMarlowe:
I don't think this will need any additional software to be installed on the linux build box, if anything, it would be 'dpkg', it's installed on all my linux boxes, and I don't remember putting it there. Do a 'which dpkg-deb' on the linux box, if its there, it's all good.
There are a couple of ways to install the atomicXXX.deb, the command line way, or most distros will bring a a gui installer when you double-click on it in a file browser, which guides you along and prompts for your password.
There is no code signing for linux packages or executables.

JimMarlowe:
I added the AtomicEditor lib dependencies, so if someone loads this on a system without graphics drivers installed, it will bark.
I also changed the Category to Programming, @CTrauma is right, those other IDEs are gonna be our new best friends. Better to be boring than forgotten, or something.
Pincushion didn't need any more emails, so I changed the maintainer to info.

JoshEngebretson:
Excellent!

Maybe we can find some kind of deb wrapper which will make an installer (basically wraps the command line dpkg stuff), so it can be installed simply?

In the meantime, LANDING! ✈️

PR #1066, opened 2016-10-02
linux package cleanup

JimMarlowe:
The initial package was good, but was not detecting if a version was already installed, which was an annoyance.  In tracking down more lintian errors, I found the I wasn't following enough rules to make the package manager happy. With these changes, now it is, and the number of remaining lintian errors, though valid, is small and manageable.
Change the package install directory name to (all lowercase) atomicgameengine.
All files do not have the execute bit set, except for AtomicEditor and AtomicPlayer, which are also stripped.
Remove all embedded .gitignore files.
Removed Depend package version number checks, they were not needed.
Fixed warning about using /tmp directory usage during installation.
Change the deb package filename to match the naming format of the other platform packages.


Build/Linux/usr/share/doc/atomicgameengine/copyright
```diff
@@ -0,0 +1,1176 @@
+Copyright (c) 2014-2016 THUNDERBEAST GAMES LLC  
+https://github.com/AtomicGameEngine/AtomicGameEngine
```
JoshEngebretson: Can this be copied in from the root license file instead of being duplicated, this way it will always be up to date and won't need to be maintained in multiple locations

JimMarlowe: Oh. Will do.

JoshEngebretson: Awesome, thanks 👍

JimMarlowe:
ok, NOW it's perfect.

PR #1069, opened 2016-10-04
Deployment for linux platform

JimMarlowe:
This PR adds Linux to the list of platform deployables. This is a deployment is in the same vein as the legacy deployments, using the linux built AtomicPlayer and adding the AtomicResource.pak to make a complete program. It does not do linux C# programs (yet).
Note: There is a small issue with execution bits on the AtomicPlayer program, when deployed, the AtomicPlayer is copied from one file to another, which removes the linux execution bits, because a new file is created. I fixed this linux on linux, but if a linux program deployed on windows or osx, and the program is moved back to linux, it will not execute. The user can restore the execute function, but it's still a little disconcerting, and no cross-platform solution (other than doing an actual os copy).


Source/ToolCore/Build/BuildLinux.cpp
```diff
@@ -0,0 +1,182 @@
...
+// the BuildCopyFile does not copy a file from one location to another,
+// instead it opens a new file, and copies the contents. This results
+// in the loss of file execution bits/acls. There is no cross platform
+// way to transferr or restore these on the copied file.
+#ifdef ATOMIC_PLATFORM_LINUX 
+    // linux-to-linux restore the execute permissions lost in copy
+    FileSystem* fileSystem = GetSubsystem<FileSystem>();
+    if (fileSystem->SystemCommand ( "chmod +x " + buildPath_ + "/AtomicPlayer") ) 
+        return;
+#endif
```
JoshEngebretson: Would a chmod call be possible here instead of shelling out to system chmod command? http://man7.org/linux/man-pages/man2/chmod.2.html, I like the idea of explicitly setting the permission, even if we were doing a file copy, permissions can still get stripped pretty easily by packaginging, etc

JoshEngebretson: We do have Poco::File which has a setExecutable call, using this wouldn't need to add any platform preprocessor platform guarded code or includes (on Windows it is a NOP) https://pocoproject.org/docs/Poco.File.html (the implementation is pretty reasonable too: https://github.com/pocoproject/poco/blob/13a999de4238e8097ac73a71de87e2fb54ec1647/Foundation/src/File_UNIX.cpp#L324)

JoshEngebretson:
Thanks for all the great work into the Linux build and deployment, LANDING! ✈️

PR #1070, opened 2016-10-04
Minor enhancements and a try/catch to prevent dll parsing from failing

mattbenic:
The first three commits are harmless/simple

The last one (wrap InlineField handling..) is related to the issue I was having last week. The parser is stumbling on components that have no constructor, and no explicit inline assigned values. This is a complete showstopper, so until a better solution can be found I'm wrapping the source of the exception in a try/catch.

The loadedValues still gets cleared, so parsing of the rest of the component (and other components in the assembly) still succeeds. Should .ctor processing perhaps just be skipped entirely?

### DIFF
--- a/Script/AtomicNET/AtomicNETService/CSComponentInspector.cs
+++ b/Script/AtomicNET/AtomicNETService/CSComponentInspector.cs
@@ -441,21 +441,28 @@ namespace AtomicTools
 
                             var fieldDef = metaReader.GetFieldDefinition(fieldHandle);
 
-                            var fieldName = metaReader.GetString(fieldDef.Name);
-
-                            if (opCode.ToString() == "stfld")
+                            // No way to predetermine if fieldDef.Name is valid
+                            try
                             {
+                                var fieldName = metaReader.GetString(fieldDef.Name);
 
-                                InspectorField inspectorField;
-
-                                if (_inspectorComponent.Fields.TryGetValue(fieldName, out inspectorField))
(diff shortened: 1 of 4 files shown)


Script/AtomicNET/AtomicNETService/CSComponentInspector.cs
```diff
@@ -441,21 +441,28 @@ internal static bool CrackStringInAttributeValue(out string value, ref BlobReade
...
-                            loadedValues.Clear();
+                                    if (_inspectorComponent.Fields.TryGetValue(fieldName, out inspectorField))
+                                    {
+                                        inspectorField.DefaultValue = String.Join(" ", loadedValues.ToArray());
+                                    }
+
+                                }
+                            }
+                            catch (Exception) { break; }
+                            finally
```
JoshEngebretson: Any value in outputting the exception?

mattbenic: I don't know. The only semi-useful info would be the class name, which could be passed down from the calling method, but even that isn't really useful since it happens on classes that have no constructor or inline assignments. If there was something we could actually report saying "fix this to make the message go away", then fine, but that doesn't seem to be the case.

JoshEngebretson:
I don't think .ctor processing can be skipped, as there is generated ctor IL code which handles the assigning of initial values (which is where we pull defaults from), etc.

Looks good, thanks!

PR #1075, opened 2016-10-08
Upgrade Monaco to 0.7

shaddockh:
Monaco 0.7 just released and this updates to that version.  

Monaco changes can be found at https://github.com/Microsoft/monaco-editor/blob/master/CHANGELOG.md#070

Additionally, the following were fixed in Atomic:
- changed the whitespace setting to match the new whitespace setting in Monaco ("none","all","border") instead of true/false.  An adapter was put in so that existing true/false entries still work.
- discovered that the monaco web workers weren't being utilized, so fixed that so that monaco can run some stuff in a background thread
- discovered that comments weren't being displayed in hovers/intellisense correctly, so that was also added


JoshEngebretson:
Excellent! Does enabling monaco web workers result in any noticeable speed improvements?

LANDING! ✈️

shaddockh:
I didn't notice anything significant.  The typescript completions were already using a custom web worker, so it would be anything else the editor might be using web workers for.

PR #1076, opened 2016-10-10
Fixes #1074

### LINKED ISSUE #1074: Editor crashes when laptop wakes
On my laptop, the editor crashes when it wakes from a sleep.

It seems to be related to trying CEF and graphics objects.

I don't have much information, but it's pretty easy to reproduce.

Type1J:
I kept running into this one since I've been on the move lately. It was simpler to fix than I thought it might be.


JoshEngebretson:
Great, thanks for the quick fix! 👍

PR #1079, opened 2016-10-10
Animation Transition Fix

GarethNN:
Resetting the skeleton of a mesh each frame causes animation blending to not work as hoped/expected when transitioning from one animation to another, involving a regression toward the base/T pose. This reset doesn't seem to really be necessary, and it looks like it would be fine to simply lerp from the last position of the skeleton into the new animation frames.

This change removes the skeleton reset, and appears to work well for the cases tested of blending between animations.


JoshEngebretson:

JoshEngebretson:
I have not noticed this, it would be good to have a reproduction case.  It seems like it is an issue which could be related to a bad FBX import, bone animation, or something else, not sure of the side effects on not reseting the master skeleton, which there may be some...

In the meantime, let's get the change up will land in #1089

PR #1080, opened 2016-10-10
Reparent multiple nodes

GarethNN:
If you have multiple nodes selected you can drag all of them and reparent them to a single parent


JoshEngebretson:
Very nice! Thanks :)

PR #1082, opened 2016-10-10
Search Bar Update

GarethNN:
Moved the search implementation to where the other widget events are eing handled.


JoshEngebretson:
Thanks @raheelx , nice improvements :)

PR #1083, opened 2016-10-10
AnimationController - Bug fix

GarethNN:
Added a null check to the AnimationController's animations array.


JoshEngebretson:

JoshEngebretson:
In debug builds, should this maybe be a warning?  This would mean an animation resource failed to load, right?

JoshEngebretson:
The logging would probably get spammy, this is probably something to handle further up the chain... there shouldn't be null resources (this is an error condition), however guarding against crashes is good :)

Will land in #1088

PR #1084, opened 2016-10-10
Clamped Lerp

GarethNN:
Added a Clamped Lerp to Vector3


JoshEngebretson:
Great thanks, will land in #1087

PR #1096, opened 2016-10-12
Adding the Web subsystem back to Engine.cpp

Type1J:
Merging the current Urho3D was really good, but it did have some casualties. The Web subsystem was giving a null because it was never registered. This PR fixes that problem.


JoshEngebretson:
Great thanks, Engine.cpp had a few dropouts

PR #1097, opened 2016-10-13
Expose Renderer2D to script

mattbenic:


JoshEngebretson:
I guess Renderer2D was the unknown component :)  Renderer2D is an internal class (self creates when needed and has no useful user API), thinking should use this as an opportunity to improve the error handling to report the unexposed class and not throw.  We should filter in this case, as we don't want to require all native components be registered with script bindings.

I'll work on this today.

mattbenic:
Doh, yeah sorry should have had the Renderer2D one in a separate feature branch. I'll get those into separate feature branches for you tomorrow.

JoshEngebretson:
That's ok, see my note on Renderer2D above, I'll address this today.

mattbenic:
Saw that, thanks. I'll replace our band-aid with the proper fix

PR #1098, opened 2016-10-13
Profiler Update

weinandvv:
Divided the triangles into categories "All pass", "Single pass" and "Editor"

### DIFF
--- a/Source/Atomic/Graphics/Direct3D9/D3D9Graphics.cpp
+++ b/Source/Atomic/Graphics/Direct3D9/D3D9Graphics.cpp
@@ -268,6 +268,10 @@ Graphics::Graphics(Context* context) :
     instancingSupport_(false),
     sRGBSupport_(false),
     sRGBWriteSupport_(false),
+    // ATOMIC BEGIN
+    numPasses_(0),
+    numSinglePassPrimitives_(0),
+    // ATOMIC END
     numPrimitives_(0),
     numBatches_(0),
     maxScratchBufferRequest_(0),
@@ -736,6 +740,9 @@ bool Graphics::BeginFrame()
     for (unsigned i = 0; i < MAX_TEXTURE_UNITS; ++i)
         SetTexture(i, 0);
 
+    // ATOMIC BEGIN
+    numSinglePassPrimitives_ = 0;
+    // ATOMIC END
     numPrimitives_ = 0;
     numBatches_ = 0;
 
@@ -847,6 +854,12 @@ void Graphics::Draw(PrimitiveType type, unsigned vertexStart, unsigned vertexCou
     impl_->device_->DrawPrimitive(d3dPrimitiveType, vertexStart, primitiveCount);
 
     numPrimitives_ += primitiveCount;
+
+    // ATOMIC BEGIN
+    if (GetNumPasses() == 1)
+        numSinglePassPrimitives_ += primitiveCount;
+    // ATOMIC END
+
     ++numBatches_;
 }
 
@@ -864,6 +877,12 @@ void Graphics::Draw(PrimitiveType type, unsigned indexStart, unsigned indexCount
     impl_->device_->DrawIndexedPrimitive(d3dPrimitiveType, 0, minVertex, vertexCount, indexStart, primitiveCount);
 
     numPrimitives_ += primitiveCount;
+
+    // ATOMIC BEGIN
+    if (GetNumPasses() == 1)
+        numSinglePassPrimitives_ += primitiveCount;
+    // ATOMIC END
+
     ++numBatches_;
 }
 
@@ -881,6 +900,12 @@ void Graphics::Draw(PrimitiveType type, unsigned indexStart, unsigned indexCount
     impl_->device_->DrawIndexedPrimitive(d3dPrimitiveType, baseVertexIndex, minVertex, vertexCount, indexStart, primitiveCount);
 
     numPrimitives_ += primitiveCount;
+
+    // ATOMIC BEGIN
+    if (GetNumPasses() == 1)
+        numSinglePassPrimitives_ += primitiveCount;
+    // ATOMIC END
+
     ++numBatches_;
 }
 
@@ -911,6 +936,11 @@ void Graphics::DrawInstanced(PrimitiveType type, unsigned indexStart, unsigned i
     impl_->device_->DrawIndexedPrimitive(d3dPrimitiveType, 0, minVertex, vertexCount, indexStart, primitiveCount);
 
     numPrimitives_ += instanceCount * primitiveCount;
+
+    // ATOMIC BEGIN
+    numSinglePassPrimitives_ += instanceCount * primitiveCount;
+    // ATOMIC END
+
     ++numBatches_;
 }
 
@@ -941,6 +971,11 @@ void Graphics::DrawInstanced(PrimitiveType type, unsigned indexStart, unsigned i
     impl_->device_->DrawIndexedPrimitive(d3dPrimitiveType, baseVertexIndex, minVertex, vertexCount, indexStart, primitiveCount);
 
     numPrimitives_ += instanceCount * primitiveCount;
+
+    // ATOMIC BEGIN
+    numSinglePassPrimitives_ += instanceCount * primitiveCount;
+    // ATOMIC END
+
     ++numBatches_;
 }
 
--- a/Source/Atomic/Graphics/Graphics.h
+++ b/Source/Atomic/Graphics/Graphics.h
@@ -537,6 +537,9 @@ public:
 
     // ATOMIC BEGIN
 
+    /// Return number of single render pass primitives drawn this frame.
+    unsigned GetSinglePassPrimitives() const { return numSinglePassPrimitives_; }
+
     /// Get the SDL_Window as a void* to avoid having to include the graphics implementation
     void* GetSDLWindow() { return window_; }
 
@@ -545,6 +548,10 @@ public:
     bool GetMaximized();
     IntVector2 GetMonitorResolution(int monitorId) const;
     void RaiseWindow();
+
+    int GetNumPasses() { return numPasses_; }
+    void ResetNumPasses() { numPasses_ = 0; }
+    void IncNumPasses() { numPasses_++; }
 
     // ATOMIC END
 
@@ -767,6 +774,15 @@ private:
     static const Vector2 pixelUVOffset;
     /// OpenGL3 support flag.
     static bool gl3Support;
+
+    // ATOMIC BEGIN
+
+    /// Number of render passes
+    int numPasses_;
+    /// Number of single render pass primitives
+    unsigned numSinglePassPrimitives_;
+
+    // ATOMIC END
 };
 
 /// Register Graphics library objects.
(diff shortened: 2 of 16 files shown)


Source/Atomic/Graphics/Direct3D9/D3D9Graphics.cpp
```diff
@@ -941,6 +946,7 @@ void Graphics::DrawInstanced(PrimitiveType type, unsigned indexStart, unsigned i
     impl_->device_->DrawIndexedPrimitive(d3dPrimitiveType, baseVertexIndex, minVertex, vertexCount, indexStart, primitiveCount);
 
     numPrimitives_ += instanceCount * primitiveCount;
+    numSinglePassPrimitives_ += instanceCount * primitiveCount;
     ++numBatches_;
```
JoshEngebretson: Ok for D3D9, in order to use with D3D11/OpenGL the metrics gathering needs to be added to D3D11 and OpenGL systems

Source/Atomic/Graphics/Graphics.h
```diff
@@ -537,6 +537,9 @@ class ATOMIC_API Graphics : public Object
 
     // ATOMIC BEGIN
 
+    /// Return number of True primitives drawn this frame.
+    unsigned GetSinglePassPrimitives() const { return numSinglePassPrimitives_; }
```
JoshEngebretson: I am not sure the word "True" describes this best, it leaves me guessing at least, can this be summarized better?

Source/Atomic/Graphics/Graphics.h
```diff
@@ -767,6 +774,14 @@ class ATOMIC_API Graphics : public Object
     static const Vector2 pixelUVOffset;
     /// OpenGL3 support flag.
     static bool gl3Support;
+
+    // ATOMIC BEGIN
+
+    int numPasses_;
```
JoshEngebretson: A comment on numPasses_ describing it?

Source/Atomic/Graphics/Renderer.cpp
```diff
@@ -757,6 +757,7 @@ void Renderer::Render()
     }
 
     // Copy the number of batches & primitives from Graphics so that we can account for 3D geometry only
+    numSinglePassPrimitives_ = graphics_->GetSinglePassPrimitives();
```
JoshEngebretson: // ATOMIC BEGIN, // ATOMIC END block here

Source/Atomic/Graphics/View.cpp
```diff
@@ -1589,8 +1589,12 @@ void View::ExecuteRenderPathCommands()
...
 
+                    graphics_->ResetNumPasses();
+
                     for (Vector<LightBatchQueue>::Iterator i = actualView->lightQueues_.Begin(); i != actualView->lightQueues_.End(); ++i)
                     {
+                        graphics_->IncNumPasses();
+
                         // If reusing shadowmaps, render each of them before the lit batches
                         if (renderer_->GetReuseShadowMaps() && NeedRenderShadowMap(*i))
                         {
```
JoshEngebretson: // ATOMIC BEGIN, // ATOMIC END block here, please note that you don't need to enclose each line separately, find the best spots for BEGIN and END that encapsulate the changes, in this case left aligning the comments would help as across indent blocks

Source/Atomic/Graphics/View.cpp
```diff
@@ -3041,6 +3045,7 @@ void View::RenderShadowMap(const LightBatchQueue& queue)
 
         if (!shadowQueue.shadowBatches_.IsEmpty())
         {
+            graphics_->IncNumPasses();
             graphics_->SetViewport(shadowQueue.shadowViewport_);
```
JoshEngebretson: // ATOMIC BEGIN, // ATOMIC END block here

Source/Atomic/UI/SystemUI/DebugHud.h
```diff
@@ -106,6 +106,21 @@ class ATOMIC_API DebugHud : public Object
...
     void ClearAppStats();
 
+    //Atomic Begin
+
+    bool GetSceneOpen() { return isSceneOpen_; }
+    void SetSceneOpen(bool isOpen) { isSceneOpen_ = isOpen; }
+
+    String GetStatsString() { return statsString_; }
+    String GetModeString() { return modeString_; }
+    String GetProfilerString() { return profilerString_; }
```
JoshEngebretson: These call all return const String& otherwise duplicating strings on return, the members can also be marked const

Source/ThirdParty/TurboBadger/tb_widgets_reader.cpp
```diff
@@ -42,6 +42,8 @@ void TBWidget::OnInflate(const INFLATE_INFO &info)
 
     SetOpacity(info.node->GetValueFloat("opacity", GetOpacity()));
 
+    SetDisabledOpacity(info.node->GetValueFloat("disabledOpacity", GetDisabledOpacity()));
```
JoshEngebretson: Nice add, one note "disabled-opacity" would be the TB convention for the attr name

Source/Atomic/Graphics/Direct3D9/D3D9Graphics.cpp
```diff
@@ -268,6 +268,7 @@ Graphics::Graphics(Context* context) :
     instancingSupport_(false),
     sRGBSupport_(false),
     sRGBWriteSupport_(false),
+    numSinglePassPrimitives_(0),
```
JoshEngebretson: numPasses_ should be initialized here as well, even if it is first Reset before used

Source/Atomic/Graphics/Direct3D9/D3D9Graphics.cpp
```diff
@@ -268,6 +268,7 @@ Graphics::Graphics(Context* context) :
     instancingSupport_(false),
     sRGBSupport_(false),
     sRGBWriteSupport_(false),
+    numSinglePassPrimitives_(0),
     numPrimitives_(0),
```
JoshEngebretson: Ok for D3D9, metric gathering would also need to be added to D3D11 and OpenGL systems.  These changes are also missing ATOMIC BEGIN, ATOMIC END

Source/ThirdParty/TurboBadger/tb_widgets.h
```diff
@@ -461,6 +461,12 @@ class TBWidget : public TBTypedObject, public TBLinkOf<TBWidget>
     void SetOpacity(float opacity);
     float GetOpacity() const { return m_opacity; }
 
+    /** Set opacity for this widget and its children from 0.0 - 1.0 when in WIDGET_STATE_DISABLED.
+    If disabled opacity is < 0, default skin opacity will be used.
+    If opacity is 0 (invisible), the widget won't receive any input. */
+    void SetDisabledOpacity(float opacity);
+    float GetDisabledOpacity() const { return m_disabledOpacity; }
```
JoshEngebretson: This is a really nice addition

Source/AtomicEditor/Editors/SceneEditor3D/SceneEditor3D.cpp
```diff
@@ -217,6 +226,20 @@ void SceneEditor3D::HandleUpdate(StringHash eventType, VariantMap& eventData)
...
+        debugHud_->SetSceneOpen(true);
+        statsText_->SetText(debugHud_->GetStatsString());
+        modeText_->SetText(debugHud_->GetModeString());
+        profilerText_->SetText(debugHud_->GetProfilerString());
+    }
+    else
+    {
+        statsText_->SetText("");
+        modeText_->SetText("");
+        profilerText_->SetText("");
```
JoshEngebretson: Instead of using "" which constructs a new String, you can use String::EMPTY

Source/AtomicEditor/Editors/SceneEditor3D/SceneEditor3D.cpp
```diff
@@ -217,6 +226,20 @@ void SceneEditor3D::HandleUpdate(StringHash eventType, VariantMap& eventData)
 {
     if (!cubemapRenderCount_)
         gizmo3D_->Update();
+
```
JoshEngebretson: Does this need to be run every frame?  Setting up text on widget controls is seldom a cheap operation

Source/AtomicEditor/Editors/SceneEditor3D/SceneEditor3D.h
```diff
@@ -22,6 +22,9 @@
 
 #pragma once
 
+#include <Atomic/UI/SystemUI/SystemUI.h>
+#include <Atomic/UI/SystemUI/DebugHud.h>
+
```
JoshEngebretson: Including SystemUI and DebugHud in to the SceneEditor header, can these be brought into the .cpp file instead, always opt for including in the source file as this helps keep the build fast and helps prevent breakage... additional includes also erode encapsulation and are often a sign the API can be better

Source/AtomicEditor/Editors/SceneEditor3D/SceneEditor3D.h
```diff
@@ -117,6 +120,8 @@ class SceneEditor3D: public ResourceEditor
 
     void UpdateGizmoSnapSettings();
 
+    SharedPtr<SystemUI::DebugHud> debugHud_;
+
```
JoshEngebretson: Hm, the SceneEditor holding a strong ref to the debugHud, let's make this a WeakRef, and would rather not require the member at all

Source/AtomicEditor/Editors/SceneEditor3D/SceneEditor3D.cpp
```diff
@@ -259,6 +282,8 @@ void SceneEditor3D::Close(bool navigateToAvailableResource)
         sceneImporter_ = nullptr;
     }
 
+    debugHud_->SetSceneOpen(false);
+
```
JoshEngebretson: This is the only time SetSceneOpen is called with false?  Though the scene is still open?  How does this play with multiple scenes being opened at same time?

Source/Atomic/Graphics/Renderer.h
```diff
@@ -598,6 +602,12 @@ class ATOMIC_API Renderer : public Object
     bool initialized_;
     /// Flag for views needing reset.
     bool resetViews_;
+    // ATOMIC BEGIN
+
+    /// Number of True primitives (3D geometry only.)
+    unsigned numSinglePassPrimitives_;
```
JoshEngebretson: I am not seeing this initialized?  Also, why is it duplicated on Renderer and the Graphics subsystems?

Source/Atomic/UI/SystemUI/DebugHud.h
JoshEngebretson: Another note here, do the backing stat widgets hold this text, otherwise we're doing unnecessary string copying

weinandvv: I was having trouble setting them to const since the String values constantly keep changing.
Sending them as String& only return 0 and the reason I ended up using Strings and not just use the variables that already existed is because the widgets can only use String values when setting their text.

JoshEngebretson:
Thanks, I'll go over the new improved PR soon, closing this PR.

PR #1101, opened 2016-10-13
linux xbuild config test

JimMarlowe:
Test for the existence of xbuild by looking for it as an installed program, instead of using a node command.
I didn't know if you wanted the OSX to do the same thing, so I left it as is, since it was working.


JoshEngebretson:
Great, landing ✈️ 
Thanks!

PR #1102, opened 2016-10-14
Profiler improvements

mattbenic:
Divided the triangles into categories "All pass", "Single pass" and "Editor"


mattbenic:
Closing, alternate incoming with review corrections rebased

PR #1105, opened 2016-10-17
Allow animation generation for skeleton hierarchies without skinned meshes

JohnnyWahib:
This change allows users to import animations into the engine without having to include a model's skinned mesh in each export. Facilitates a common work flow used in other game engines.


JoshEngebretson:
Great, thanks for the PR

PR #1112, opened 2016-10-19
Divided the triangles into categories "All pass", "Single pass" and "…

weinandvv:
…Editor"


JoshEngebretson:
Great, thanks for the updated PR, I'll have a look soon, was the intent to make this PR from ```LumaDigital:vse``` btw and not a feature branch?

mattbenic:
vse and the feature branch are currently on the same commit

mattbenic:
let me replace it

PR #1113, opened 2016-10-19
Divided the triangles into categories "All pass", "Single pass" and "…

mattbenic:
…Editor"


JoshEngebretson:
Thanks for the updated PR, will go over this as soon as I can 👍

JoshEngebretson:
I tweaked this in #1136, the per frame update was quite expensive and reduced performance by 4-6 frames a second.  I also reduced some complexity and maintained the system ui usage for the debug hud

PR #1117, opened 2016-10-19
webgl upgrade

JimMarlowe:
This allows for a more refined WebGL experience, with some different themes for the web page, an icon and the ability to change the game frame size. I was trying to make the game iframe responsive, but ran into possible cross domain complications, so fell back to setting the width and height.
I also whacked the splash screen icons, since they are not implemented.


JimMarlowe:
I'll fix the initialization for the theme and game width, height.

JoshEngebretson:
Great, thanks! 👍 

I like the photography btw 📷 

LANDING

PR #1126, opened 2016-10-24
fix linux child processes not dying

### LINKED ISSUE #1125: [Linux]  AtomicEditor leaves running process AtomicNETService.exe
When you start the AtomicEditor, a mono + AtomicNETService.exe is also started, this is expected. On linux, when you close the editor, the AtomicNETService.exe is not stopped, so they hang around. Subsequent AtomicEditors start another AtomicNETService.exe and so on. 
OSX and Windows do not appear to have this problem. Im hoping @CTrauma can test this on linux distros other than Mint, to see how big the problem really is. 
   In diagnosing the problem, I'm finding that AtomicNETService.exe process is never told to send the event to engine it is running, to exit. It has a case in it's command loop to do this. Perhaps the process that is suppose to send this dies quicker on linux than other platforms, or they are suppose to be stopped some other way.

JimMarlowe:
This is to fix the AtomicNETService.exe not stopping when the AtomicEditor is closed #1125 
Using the StackOverflow most popular answer, it appears to work without (bad) side effects.

### DIFF
--- a/Source/Atomic/Engine/Application.cpp
+++ b/Source/Atomic/Engine/Application.cpp
@@ -33,6 +33,13 @@
 // ATOMIC END
 #endif
 
+// ATOMIC BEGIN
+#if defined(__linux__)
+#include <sys/prctl.h>
+#include <signal.h>
+#endif
+// ATOMIC END
+
 #include "../DebugNew.h"
 
 namespace Atomic
@@ -53,6 +60,12 @@ Application::Application(Context* context) :
     Object(context),
     exitCode_(EXIT_SUCCESS)
 {
+// ATOMIC BEGIN
+#if defined(__linux__)
+    prctl(PR_SET_PDEATHSIG, SIGHUP);
+#endif
+// ATOMIC END
+
     engineParameters_ = Engine::ParseParameters(GetArguments());
 
     // Create the Engine, but do not initialize it yet. Subsystems except Graphics & Renderer are registered at this point


JoshEngebretson:
Does this logic need to be in the parent application or the child, both?  If the child, maybe a better place would be in the IPCClientApp: https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/AtomicApp/IPCClientApp.cpp

I'd also like to look into why the managed player exits with the editor fine, while the managed service doesn't

JoshEngebretson:
Thanks for the leg work :) 

I am looking into this, need to see why the managed player doesn't have the issue. If possible would like a clean exit from the NET service (and other processes) as well.

From what I am reading ```prctl(PR_SET_PDEATHSIG, SIGHUP);``` is called in the child process so if necessary will move to the IPC subsystem.

JimMarlowe:
Sorry I had to fight with windows last night.
prctl is a kernel service, and it appears to save the child pids when the parent forks off a new sub process. When the parent dies the kernel then takes the list of child pids and issues whatever signal you specified to be issued to them. This would enforce the policy that any process would clean up it's children, which appears to be the case with OSX and Windows, and avoid zombie processes that the user must hunt down and kill, re #1125. 
If you intend to have a linux child process stay running after the parent exits, you do a double fork. I'm guessing OSX and windows have the same behaviour and launching VS/XS/MD is using Poco::Process::launch()  achieves this in a platform independent way.

JoshEngebretson:
np, I filed #1128 as a follow up

PR #1132, opened 2016-10-31
graphics FPS api

JimMarlowe:
This pr adds a graphics frames per second api to Atomic.
The api is:   float Graphics::GetGraphicsFps() for C++,
                    Atomic.graphics.getGraphicsFps()  for JS


JoshEngebretson:
Cool, I tweaked this in #1133 to add the current fps to the engine subsystem (which is consistent with the other fps functionality (get/set min/max fps, etc).  It also avoids needing to touch multiple graphics subsystems as they (unfortunately) don't have a common base constructor.

PR #1134, opened 2016-11-03
Copy libAtomicPlayer.so to build dir

JimMarlowe:
The --with-android task correctly builds the libAtomicPlayer.so, and it copies it into AtomicEditor.app/Contents/Resources/ToolData/Deployment/Android/libs/armeabi-v7a, for managed C# deployment. This leaves the in-editor Build Android option unable to create an apk, as Dirk found in a Q&A entry.  This additional copy will restore the editor Build android functionality.


JoshEngebretson:
Thanks for the PR, good catch 👍

PR #1150, opened 2016-11-14
A number of small issue fixes

mattbenic:


Source/Atomic/Graphics/AnimatedModel.cpp
```diff
@@ -1044,7 +1044,7 @@ void AnimatedModel::OnWorldBoundingBoxUpdate()
     if (isMaster_)
     {
         // Note: do not update bone bounding box here, instead do it in either of the threaded updates
-        worldBoundingBox_ = boneBoundingBox_.Transformed(node_->GetWorldTransform());
+        worldBoundingBox_ = boundingBox_.Transformed(node_->GetWorldTransform());
     }
```
JoshEngebretson: This is interesting, can the issue be described, seems like something that should be PR'd to Urho project

mattbenic: @weinandvv please expand on this for Josh

weinandvv: At first any object, excluding prefabs, that had an AnimatedModel component
attached to it could'nt be focused on by pressing the F-key. The issue was
getting the BoundingBox to be drawn again as the F-Key focus and BoundingBox
seems to be related to each other.

Each model's BoundingBox gets created based upon the Center point of the model.
Using the BoneBoundingBox no Center point value could be found. I figured it's 
because the bones don't get created in the scene. So the BoundingBox was never 
drawn.

Using the BoundingBox of the model instead of the BoundingBox of the Bones
a Center point value could be found (probably the center of the mesh?) and the box was drawn and updating properly again.

JoshEngebretson: Ok, I understand the issue, unfortunately this will break the runtime bounding box, potentially causing false clipping, etc.

Please check for the editor context:  ```context_->GetEditorContext();``` before using the boundingBox instead, also place in an // ATOMIC BEGIN and // ATOMIC END block, with either a filed issue linked or as best and concise of a description as you can muster.  Otherwise, when pulling upstream, these changes become very easy to lose/tell if they are still needed.  

This issue will go away once prefabs can be revisited.

mattbenic: @weinandvv please address this in the LumaSmallPRs branch

JoshEngebretson:
Thanks for the PR, just the one issue and can land 👍

JoshEngebretson:
Ok, I fixed this up in #1179 and will land from there, I also added #1166 to the Build 1 milestone for tracking purposes, and filed #1178, whew :)

PR #1158, opened 2016-11-16
Fixes #177 - Display name of JS file in inspector when a JS component is selected

### LINKED ISSUE #177: JSComponent in inspector should display component title
Instead of the drawer in the inspector being titled "JSComponent", it seems like it should display either the full path or just the bare filename of the component file it references.  Maybe: "Enemy (js)" or "JSComponent - Enemy". If you get multiple JSComponents attached to a node, you don't know which is which when the drawers are closed.

shaddockh:
I opted for just displaying the bare .js filename with the .js file extension here.  

So,
```Components/Spinner.js``` would be titled ```Spinner.js```.  That should give enough information to know that it is a JSComponent and that it resides in Spinner.js and doesn't take up extra screen real-estate on the title bar.

I didn't touch the way the C# components display (the class name).


JoshEngebretson:
Great!  For posterity, this is the very first issue closed for an Atomic Build.  Congrats and sorry if this makes more than one thrill for the day ;)

Issue #1153, opened 2016-11-14
[UI] - Fix dark theme checkmark graphic


JoshEngebretson:
The checkmark graphic has some issues, both in not filling the space very well and the checkmark itself is a bit crunchy

![checkcapture](https://cloud.githubusercontent.com/assets/376203/20287297/7dfc5db0-aa80-11e6-8928-542f530d3883.PNG)

PR #1164, opened 2016-11-17
improve dark skin checkbox

JimMarlowe:
This improves the the outline coverage of the image, and smooths the check outline, as mush as possile with so few pixels. From #1153


JoshEngebretson:
The check looks fantastic :)  Thanks, LANDING! ✈️ 

newchecks

PR #1165, opened 2016-11-17
Adding UIOffscreenView

Type1J:
I need UI as part of the 3D scene. This allows Turbo Badger to render additional root widgets to offscreen render targets.

Input mapping is also included, and must be opted-in, and can be disabled and re-enabled at will.


Type1J:
I can't see the details on that build: http://atomicbuilds.mooo.com:9765/job/AtomicPR_MacEditor_iOS_Android/132/

I don't have DNS info for that domain. Is there another way to see the log for that build?

JoshEngebretson:
Windows decided to update and so the build timed out as the machine wasn't available.  

The build errors used to forward from mooo.com to our dynamic IP, though this meant keeping a port open and the jenkins + that box live on the internet, which isn't really a great plan these days :)

So, need to come up with better system for reporting build errors, maybe moving CI from build boxes into the cloud (only thing there is full platform coverage and possible hosting expenses)

I'll kick off the build again if it is jammed, now that Windows build box is back up.

Type1J:
Awesome. Thanks! :-)

JoshEngebretson:

JoshEngebretson:
I'll go over the PR in more detail soon, going to be **awesome** to have 3D UI widgets :)  

We can probably update the 3D WebView example to use it and provide additional use/test coverage: https://github.com/AtomicGameEngine/AtomicExamples/blob/master/WebView3D/JavaScript/Resources/Components/WebTexture.js

There's a build error with super being a keyword in TypeScript, maybe ```superDown``` would be better, this is an inherited API issue as super really should just be coded into qualifiers.

```typescript
invokeRightPointerDown(x: number, y: number, click_count: number, qualifiers: number, super?: boolean): void;
```

Type1J:
Did the Linux build fail?

Type1J:
... or did the machine timeout?

JoshEngebretson:
I think the build workspace was dirty as failed building script bindings where other host platforms passed, will retest

JoshEngebretson:
Compile update the Linux build box doesn't like:

```
UIOffscreenView.h:41:12: error: ‘UI’ does not name a type
friend UI;
```

Type1J:
I think I hit all of your points except for 1:
`HashSet<UIOffscreenView*>* GetOffscreenViews()`
was not changed changed to
`const HashSet<UIOffscreenView*>& GetOffscreenViews()`
, and instead was changed to
`HashSet<UIOffscreenView*>& GetOffscreenViews()` (no `const`)
because it's primary use is for the UIOffscreenView constructor and destructor to register and unregister with the UI subsystem by modifying that HashSet. I could just make it a friend, and hide that process, or add public functions for registration and deregistration, but they would have no practical use except for the UIOffscreenView constructor and destructor.

JoshEngebretson:
Thanks! 👍 

I am going to spend a little time with the changes before landing, great work :)

Source/Atomic/UI/UI.cpp
```diff
@@ -316,23 +316,38 @@ void UI::AddFont(const String& fontFile, const String& name)
...
+    // On OpenGL, flip the projection if rendering to a texture so that the texture can be addressed in the same way
+    // as a render texture produced on Direct3D
+#ifdef ATOMIC_OPENGL
+    if (renderSurface)
+    {
+        offset.y_ = -offset.y_;
+        scale.y_  = -scale.y_;
+    }
+#endif
+
```
JoshEngebretson: :+1:

JoshEngebretson:
Whew, think we set a record here 🏅  :)

Refactored and landed in https://github.com/AtomicGameEngine/AtomicGameEngine/pull/1582

PR #1172, opened 2016-11-18
Implements support for using engine through `add_subdirectory()`.

rokups:
`CMAKE_SOURCE_DIR` variable replaced with `ATOMIC_SOURCE_DIR` (which is alias to `Atomic_SOURCE_DIR` provided by cmake). `ATOMIC_SOURCE_DIR` points to root directory of Atomic project just like `CMAKE_SOURCE_DIR` does.

Introduced three internal cache variables that provide information to c++ projects that use engine through `add_subdirectory()`:
  * ATOMIC_INCLUDE_DIRS - list of Atomic include directories.
  * ATOMIC_DEFINITIONS - list of Atomic build definitions.
  * ATOMIC_LINK_LIBRARIES - list of Atomic link libraries.

Introduced three macros:
  * include_directories_exported() - behaves like `include_directories()`, also appends included directories to `ATOMIC_INCLUDE_DIRS` cache variable.
  * add_definitions_exported() - behaves like `add_definitions()`, also appends added definitions to `ATOMIC_DEFINITIONS` cache variable.
  * add_link_libraries_exported() - appends libraries to `ATOMIC_LINK_LIBRARIES` cache variable. Targets still have to manually link to libraries specified in `ATOMIC_LINK_LIBRARIES` variable.

Macros are to be used when includes/definitions/libraries should be exposed to game projects linking to engine.

Sample CMakeLists.txt of project that embeds engine through `add_subdirectory()`:

    cmake_minimum_required(VERSION 3.6)
    project(AtomicTest)
    
    add_subdirectory(AtomicGameEngine)
    include_directories(${ATOMIC_INCLUDE_DIRS})
    add_definitions(${ATOMIC_DEFINITIONS})
    
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
    
    set(SOURCE_FILES main.cpp)
    add_executable(AtomicTest ${SOURCE_FILES})
    target_link_libraries (AtomicTest ${ATOMIC_LINK_LIBRARIES})

P.S. Turns out i slipped in silencing `find_package()` warnings because they were annoying to me. I think these changes are worth having because some of packages are mutually exclusive and project would generate projects no matter configuration. If you think these changes have no place in this PR then let me know and i will fix this commit.


JoshEngebretson:

JoshEngebretson:
Thanks for the PR, this is fantastic and would have thought it required loads more changes :)

We'll get CI churning on it and I'll give it a good looking over.

Build/CMake/Modules/AtomicLinux.cmake
```diff
@@ -3,19 +3,15 @@ set (ATOMIC_NODE_JAKE Build/Linux/node/node Build/node_modules/jake/bin/cli.js
 
 include(AtomicDesktop)
 
-add_definitions(-DATOMIC_PLATFORM_LINUX -DATOMIC_OPENGL -DKNET_UNIX -DHAVE_INT64_T)
+add_definitions_exported(-DATOMIC_PLATFORM_LINUX -DATOMIC_OPENGL -DATOMIC_TBUI -DKNET_UNIX -DHAVE_INT64_T)
 
-set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fPIC")
-set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-invalid-offsetof -std=gnu++0x -fPIC")
+set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-warn-absolute-paths -fPIC")
+set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-warn-absolute-paths -Wno-invalid-offsetof -std=gnu++0x -fPIC")
```
JoshEngebretson: Good attention to detail...

Build/CMake/Modules/AtomicCommon.cmake
```diff
@@ -203,3 +203,19 @@ macro (setup_executable)
     setup_target ()
 
 endmacro ()
+
+macro (include_directories_exported)
+    include_directories(${ARGV})
```
JoshEngebretson: I thought about these names a bit, I think "_exported" is pretty accurate, it would be good to add some comments about what these macros do here

CMakeLists.txt
```diff
@@ -3,12 +3,14 @@ project (Atomic)
 
 cmake_minimum_required(VERSION 2.8.12.1)
```
JoshEngebretson: Nice we don't need to bunch CMake to achieve this...

Source/Atomic/CMakeLists.txt
```diff
@@ -1,13 +1,6 @@
...
-                    ${CMAKE_SOURCE_DIR}/Source/ThirdParty
-                    ${CMAKE_SOURCE_DIR}/Source/ThirdParty/WebSocketPP/include
-                    ${CMAKE_SOURCE_DIR}/Source/ThirdParty/rapidjson/include
-                    ${CMAKE_SOURCE_DIR}/Source/ThirdParty/libcurl/include
-                    ${CMAKE_SOURCE_DIR}/Source/ThirdParty/kNet/include
-                    ${CMAKE_SOURCE_DIR}/Source/ThirdParty/FreeType/include
-                    ${CMAKE_SOURCE_DIR}/Source/ThirdParty/Box2D
-                    ${CMAKE_SOURCE_DIR}/Source/ThirdParty/ASIO/include)
+include_directories(${ATOMIC_INCLUDE_DIRS})
+include_directories_exported(${CMAKE_CURRENT_SOURCE_DIR})
```
JoshEngebretson: Good cleanup

Source/CMakeLists.txt
```diff
@@ -1,4 +1,6 @@
 
+include_directories_exported(${CMAKE_CURRENT_SOURCE_DIR})
```
JoshEngebretson: It may be time to clean this up: https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/CMakeLists.txt#L27 which is effectively adding same folder (though not exported to parent projects)

Source/ThirdParty/SDL/cmake/sdlchecks.cmake
```diff
@@ -104,7 +104,7 @@ endmacro()
 macro(CheckOSS)
   if(OSS)
     # Urho3D - bug fix - should use different variables for different checks, however, we replace the whole checks with find_package() approach for consistency sake
-    find_package (OSS)
+    find_package (OSS QUIET)
     if(OSS_FOUND)
```
JoshEngebretson: These QUIET's are good, please note changes to third party sources/scripts with an ATOMIC BEGIN/ATOMIC END comment block and description of change.  Otherwise, easy to lose changes when updating from upstream.

JoshEngebretson:
Ok, this checks out in testing here

Type1J:
This will be awesome! It will simplify our repo management after a refactor. Wide  ease of use benefits here.

Source/CMakeLists.txt
rokups: I noticed that, but since comment said it is for QtCreator i figured maybe it is just some dumb include to make IDE happy therefore i did not touch it. If someone could verify if IDE autocompletion without those lines still works it would be great. For some reason QtCreator crashes on my system. :|

rokups:
I added requested changes. Sorry for being a bit late - emails of this issue got lost in sea of other emails from AtomicGameEngine org :)

Build/CMake/Modules/AtomicCommon.cmake
```diff
@@ -204,16 +204,32 @@ macro (setup_executable)
...
 macro (add_definitions_exported)
     add_definitions(${ARGV})
     set (ATOMIC_DEFINITIONS ${ATOMIC_DEFINITIONS} ${ARGV} CACHE INTERNAL "Atomic game engine definitions" FORCE)
 endmacro ()
 
+# Macro for exporting game engine link libraries. Adds specified targets/libraries to ATOMIC_LINK_LIBRARIES cache
+# variable. This macro does not implicitly add link libraries to any target.
+# Macro arguments:
+#  List of definitions
 macro (add_link_libraries_exported)
```
JoshEngebretson: Great comments, thanks for adding 👍

JoshEngebretson:
This is really fantastic and far less sweeping changes than I would have expected.  We can continue to tweak, though this PR is well along the path.  Thanks! LANDING ✈️

PR #1183, opened 2016-11-19
Add audio strings into bindings for 1152

JimMarlowe:
This will allow const String constants in general to be bound into both JS and C#.  I thought that this might match other instances in the Atomic source code, but the Audio strings were the only ones.
I freely admit to repurposing the char primitive type to represent a string for the module constants, for two reasons, only numeric constants were being registered, and changing the primitives type enum could have compatibility issues.

### DIFF
--- a/Source/ToolCore/JSBind/CSharp/CSModuleWriter.cpp
+++ b/Source/ToolCore/JSBind/CSharp/CSModuleWriter.cpp
@@ -203,7 +203,9 @@ String CSModuleWriter::GetManagedPrimitiveType(JSBPrimitiveType* ptype)
         return "int";
     if (ptype->kind_ == JSBPrimitiveType::Float)
         return "float";
-
+    if (ptype->kind_ == JSBPrimitiveType::Char)
+        return "string";  // it technically should return "char", but the
+                          // intent is really to support a string constant
     return "int";
 }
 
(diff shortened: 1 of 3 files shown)


Source/ToolCore/JSBind/CSharp/CSModuleWriter.cpp
```diff
@@ -203,7 +203,9 @@ String CSModuleWriter::GetManagedPrimitiveType(JSBPrimitiveType* ptype)
         return "int";
     if (ptype->kind_ == JSBPrimitiveType::Float)
         return "float";
-
+    if (ptype->kind_ == JSBPrimitiveType::Char)
+        return "string";  // it technically should return "char", but the
+                          // intent is really to support a string constant
     return "int";
```
JoshEngebretson: Good comment here

Source/ToolCore/JSBind/JSBHeaderVisitor.h
```diff
@@ -568,6 +568,24 @@ class JSBHeaderVisitor : public SymbolVisitor
...
+            if ( dtype.isConst() && classname == "String" )  // find a const string
+            {
+                const StringLiteral* sinit = decl->getInitializer();
+                if (sinit)
+                {
+                    if (sinit->chars())
+                    {
+                        String svalue = sinit->chars();
+                        module_->RegisterConstant(getNameString(decl->name()).CString(), svalue, JSBPrimitiveType::Char );
+                        return true;  // tag this constant as a Char type, because there is no String Primitive type
```
JoshEngebretson: Ok, great solution, we can "special case" the char const for bindings

PR #1198, opened 2016-11-24
Enable linux accelerator keys

JimMarlowe:
The way the control/Command key was being handled wrong for linux. It was falling into the OSX behavior with a Command key, when it actually follows the Windows Control key behavior.


JoshEngebretson:
Fantastic, great catch! ⚾️ 

LANDING ✔️

PR #1201, opened 2016-11-24
Added animation preview tool bar

JohnnyWahib:
Created an animation preview toolbar to test out animations at custom playback speeds. It also allows the user to blend between 2 specified animations using custom blend speeds.

Open the window by clicking on the Preview Animation Button that now appears in the inspector when the user has a model selected.


JoshEngebretson:
Woah, cool, will look over in detail though at first glance excellent work on a tough (though temporary) editor limitation 👍

Source/ThirdParty/TurboBadger/tb_widgets.cpp
```diff
@@ -359,7 +359,8 @@ void TBWidget::AddChildRelative(TBWidget *child, WIDGET_Z_REL z, TBWidget *refer
 
 void TBWidget::RemoveChild(TBWidget *child, WIDGET_INVOKE_INFO info)
 {
-    assert(child->m_parent);
+    if (!child->m_parent)
+        return;
```
JoshEngebretson: I think this is a good change, as long as we are tripping the assert by explicitly calling RemoveChild.  If we're hitting the assert as a side effect of something, that should be looked into

JohnnyWahib: As it stands I would say it is a side effect of another call, I will need to look into it. I suspect it has something to do with the order in which I close the scene editor and remove the Animation previewer widget...

JoshEngebretson: Ok, if we do keep the guard in, we should check that the parent == this, otherwise things can go sideways later in the call.  I am also seeing a crash when clicking the preview button on macOS, can provide a callstack tomorrow.

JoshEngebretson: Ok, here's the crash I am seeing on macOS in release mode, I tried a few times in debug and didn't trip it:

https://github.com/AtomicGameEngine/AtomicGameEngine/issues/1216

I haven't verified tried to reproduce on Windows yet

JoshEngebretson: Ok, I fixed the crash in https://github.com/AtomicGameEngine/AtomicGameEngine/pull/1218

Kind of esoteric, though when a widget is deleted in its own callback, it needs to return true to let the underlying system know the event was handled.  I also changed the remove child guard to ensure ```m_parent == this``` and added // ATOMIC BEGIN // ATOMIC END block with a note on change, whew :)

PR #1196, opened 2016-11-23
[Build] Support excluding C++ examples from parent builds

rokups:
Allow to exclude examples from build by setting ATOMIC_CPLUSPLUS_EXAMPLES in parent project.

Export ATOMIC_CXX11, ATOMIC_64BIT, ATOMIC_BUILD_DIST (allows parent projects to use Atomic's c++11 support).

PR #1208, opened 2016-11-27
Updating examples commit for usage of Atomic through add_subdirectory().

Type1J:
Since #1196 the AtomicExamples sub module has not been advanced to see AtomicGameEngine/AtomicExamples#95.


JoshEngebretson:
Thanks for the PR 👍

PR #1213, opened 2016-11-28
Add compression command to AtomicTool

mattbenic:
This PR introduces a command to AtomicTool that allows a directory of images to be compressed for later use by the engine (eg by adding it as a Resource directory)

usage: AtomicTool compress "Path/To/Directory"

@JoshEngebretson I'm in two minds on this. We definitely need it for external resource directories as I've mentioned before. However given that it also generates a cache directory (useful for the same reason), just calling it "compress" seems incorrect. It _could_ just be part of the project command, but it doesn't actually require a project to be loaded. It kind of feels like something we should just keep to our side, then we don't need to worry about issues like that. I PR'd it for your ease of comparison.

### DIFF
--- a/Source/ToolCore/Command/Command.cpp
+++ b/Source/ToolCore/Command/Command.cpp
@@ -29,7 +29,8 @@ namespace ToolCore
 {
 
 Command::Command(Context* context) : Object(context),
-    timeOut_(0.0f)
+    timeOut_(0.0f),
+    requiresProjectLoad_(false)
 {
 
 }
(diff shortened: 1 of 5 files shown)


mattbenic:
And it looks like the entire command needs to be guarded for non-windows platforms. Before doing that though, let's establish if it's something you want to include :)

JoshEngebretson:
This looks solid, though I think we can hold back for now from including in stock AtomicTool.  So, I'll close the issue and we can revisit.  

On mobile side want other compressors, etc.  It is possible we will want to bundle some image utils, or maybe an AtomicImage to keep AtomicTool somewhat lean (for example, AtomicGlow will have it's own executable I am thinking, so we're not linking embree in AtomicTool)

PR #1220, opened 2016-11-30
AtomicTool improvements

mattbenic:
General AtomicTool improvements:
-Support for project paths being specified either as the project directory or the full .atomic file path
-Support for specifying log level with -loglevel argument

Project command specifically
-Fixed project<path> cache -clean reporting an error when the cache directory doesn't yet exist


JoshEngebretson:
"Support for project paths being specified either as the project directory or the full .atomic file path", good one, this has been a source of head scratching on the command line 👍 

Thanks, landing!

PR #1238, opened 2016-12-04
[WINDOWS] Fix for blurred text in editor under OpenGL on Windows.

darrylryan:
This was due to scaling logic in TurboBadger/renderers/tb_renderer_batcher.cpp not accounting for the OpenGL case.


JoshEngebretson:
Great, thanks for the PR 👍

PR #1239, opened 2016-12-05
Add a flag to set whether non-skinned bones should be included

mattbenic:
In some cases it's necessary to import non-skinned bones on a skinned model. OpenAssetImporter ignores them by default,


JoshEngebretson:
Thanks for the PR, for tags and things this seems like something we may want to default on and then have option to turn off per model in its import settings?  If so, can file an issue for the change.  In the meantime, this looks great, landing

PR #1241, opened 2016-12-05
Added 2D particle emission control

Alan-FGR:
Just a quick fix, dunno if acceptable

### DIFF
--- a/Source/Atomic/Atomic2D/ParticleEmitter2D.cpp
+++ b/Source/Atomic/Atomic2D/ParticleEmitter2D.cpp
@@ -326,7 +326,9 @@ void ParticleEmitter2D::Update(float timeStep)
 
 bool ParticleEmitter2D::EmitParticle(const Vector3& worldPosition, float worldAngle, float worldScale)
 {
+    // ATOMIC BEGIN
     if (!emitting_ || numParticles_ >= (unsigned)effect_->GetMaxParticles() || numParticles_ >= particles_.Size())
+    // ATOMIC END
         return false;
 
     float lifespan = effect_->GetParticleLifeSpan() + effect_->GetParticleLifespanVariance() * Random(-1.0f, 1.0f);
--- a/Source/Atomic/Atomic2D/ParticleEmitter2D.cpp
+++ b/Source/Atomic/Atomic2D/ParticleEmitter2D.cpp
@@ -106,7 +106,7 @@ void ParticleEmitter2D::SetEffect(ParticleEffect2D* model)
 
 void ParticleEmitter2D::SetEmissionEnabled(bool enable)
 {
-	emissionEnabled_ = enable;
+    emissionEnabled_ = enable;
 }
 
 void ParticleEmitter2D::SetSprite(Sprite2D* sprite)
@@ -143,7 +143,7 @@ void ParticleEmitter2D::SetMaxParticles(unsigned maxParticles)
 
 bool ParticleEmitter2D::GetEmissionEnabled() const
 {
-	return emissionEnabled_;
+    return emissionEnabled_;
 }
 
 ParticleEffect2D* ParticleEmitter2D::GetEffect() const
--- a/Source/Atomic/Atomic2D/ParticleEmitter2D.cpp
+++ b/Source/Atomic/Atomic2D/ParticleEmitter2D.cpp
@@ -104,6 +104,11 @@ void ParticleEmitter2D::SetEffect(ParticleEffect2D* model)
     emissionTime_ = effect_->GetDuration();
 }
 
+void ParticleEmitter2D::SetEmissionEnabled(bool enable)
+{
+	emissionEnabled_ = enable;
+}
+
 void ParticleEmitter2D::SetSprite(Sprite2D* sprite)
 {
     if (sprite == sprite_)
@@ -134,6 +139,11 @@ void ParticleEmitter2D::SetMaxParticles(unsigned maxParticles)
     sourceBatches_[0].vertices_.Reserve(maxParticles * 4);
 
     numParticles_ = Min(maxParticles, numParticles_);
+}
+
+bool ParticleEmitter2D::GetEmissionEnabled() const
+{
+	return emissionEnabled_;
 }
 
 ParticleEffect2D* ParticleEmitter2D::GetEffect() const
@@ -323,7 +333,7 @@ void ParticleEmitter2D::Update(float timeStep)
 
 bool ParticleEmitter2D::EmitParticle(const Vector3& worldPosition, float worldAngle, float worldScale)
 {
-    if (numParticles_ >= (unsigned)effect_->GetMaxParticles() || numParticles_ >= particles_.Size())
+    if (!emissionEnabled_ || numParticles_ >= (unsigned)effect_->GetMaxParticles() || numParticles_ >= particles_.Size())
         return false;
 
     float lifespan = effect_->GetParticleLifeSpan() + effect_->GetParticleLifespanVariance() * Random(-1.0f, 1.0f);
--- a/Source/Atomic/Atomic2D/ParticleEmitter2D.cpp
+++ b/Source/Atomic/Atomic2D/ParticleEmitter2D.cpp
@@ -48,7 +48,10 @@ ParticleEmitter2D::ParticleEmitter2D(Context* context) :
     emissionTime_(0.0f),
     emitParticleTime_(0.0f),
     boundingBoxMinPoint_(Vector3::ZERO),
-    boundingBoxMaxPoint_(Vector3::ZERO)
+    boundingBoxMaxPoint_(Vector3::ZERO),
+    // ATOMIC BEGIN
+    emitting_(true)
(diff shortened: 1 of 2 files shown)


Source/Atomic/Atomic2D/ParticleEmitter2D.h
```diff
@@ -88,6 +88,8 @@ class ATOMIC_API ParticleEmitter2D : public Drawable2D
     /// Handle enabled/disabled state change.
     virtual void OnSetEnabled();
 
+	/// Set emission enabled.
+	void SetEmissionEnabled(bool enable);
     /// Set particle effect.
```
JoshEngebretson: Look great, and thanks for adding /// comments, these will automatically show up in C# and TS.

Alan-FGR:
I guess this isn't the most efficient way to do that, maybe the check could happen here:
https://github.com/AtomicGameEngine/AtomicGameEngine/blob/d3b5069adedd8c4b78dae0949c1cdfe8e046c73e/Source/Atomic/Atomic2D/ParticleEmitter2D.cpp#L310

JoshEngebretson:
Ok, just update from master first, as I believe that line was updated recently: https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/Atomic/Atomic2D/ParticleEmitter2D.cpp#L300

Alan-FGR:
Yes, it was updated to fix a bug :). I'm not 100% sure we could place the logic in that line because I can't test it extensively right now and my brain doesn't have the 32kb of RAM necessary to run that code :P.

Alan-FGR:
On a more detailed analysis, by skipping that code block emitParticleTime_ won't be increased... but there's something strange in that code, how it comes that emissionTime_ is tested again on 316 if it's not modified in between the tests? (both EmitParticle and UpdateParticle don't seem to change that)

JoshEngebretson:
This is looking good, can we get the modifications between ATOMIC BEGIN/END blocks?  The new API, local variable, and implementation additions should be localized within these blocks.  There have been problems with changes being lost when pulling in updates and this helps mitigate that, thanks! 

For examples please see:

```New API Example```
https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/Atomic/Atomic2D/RigidBody2D.h#L158 

```New Variable Example```
https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/Atomic/Atomic2D/RigidBody2D.h#L188

```New Variable Initialization Example```
https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/Atomic/Atomic2D/RigidBody2D.cpp#L54

```Localized New API Implementation Example```
https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/Atomic/Atomic2D/TileMapDefs2D.cpp#L242

**Comment by Alan-FGR** [comment]
OK, I gotta pull an updated version with the Atomic safeguards... however, any ideas on the emissionTime_ check on line 316 on the current master? Its inside a code block that performs the same check and there's nothing in there that modifies that var unless I'm missing something... I mean, in which situation that will be false?

**Comment by JimMarlowe** [comment]
With the ATOMIC code guards, please remember to initialize the boolean that is being added, on different platforms, the initial value is not always set.
The emmisionTime_ check came in with PR I added, in order to support the duration spec in the pex file. The additional check may be of limited value, removing it would be more maintenance, and a change for change sake.

**Comment by JoshEngebretson** [comment]
Thanks guys! :)

I had a look as well and everything looks good,  am pretty sure that check is extraneous, so it could be removed.  I'm a fan of deleting lines when possible, red on a PR is nearly as good as green, and sometimes better :)  

C++ doesn't automatically initialize default member variable values, so it should be initialized in the constructor, in an ATOMIC BEGIN/END block:  https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/Atomic/Atomic2D/ParticleEmitter2D.cpp#L51

**Comment by Alan-FGR** [comment]
OK, as discussed with @JimMarlowe we decided it's better to use SetEmitting() and IsEmitting() to keep it consistent with the 3D counterpart. I think GetEmitting() could be implemented so the binding parser will parse into properties and stuff like that. That being said I'm having minor difficulties on technical lands, for some reason after rebasing and adding the guards the bindings aren't 100% yet, I'm trying to figure what's the problem but I'm going to update the PR if I can't find it out tomorrow since the code is simple enough. I've made some mistakes and forgot to rebranch after cp so it's not a matter of running the command 😝... noobs these days 🤣

File: Source/Atomic/Atomic2D/ParticleEmitter2D.cpp [hunk 2]
```diff
@@ -48,7 +48,10 @@ ParticleEmitter2D::ParticleEmitter2D(Context* context) :
     emissionTime_(0.0f),
     emitParticleTime_(0.0f),
     boundingBoxMinPoint_(Vector3::ZERO),
-    boundingBoxMaxPoint_(Vector3::ZERO)
+    boundingBoxMaxPoint_(Vector3::ZERO),
+    // ATOMIC BEGIN
+    emitting_(true)
+    // ATOMIC END
```
JimMarlowe: Using the same variable name as the 3d emitter is good.

Source/Atomic/Atomic2D/ParticleEmitter2D.cpp
```diff
@@ -312,8 +316,7 @@ void ParticleEmitter2D::Update(float timeStep)
             emitParticleTime_ -= timeBetweenParticles;
         }
 
-        if (emissionTime_ > 0.0f)
-            emissionTime_ = Max(0.0f, emissionTime_ - timeStep);
+        emissionTime_ = Max(0.0f, emissionTime_ - timeStep);
```
JimMarlowe: Sorry, I'm still torn about the removal of this line, since there are no guards, and the next rebase or merge from Urho3D can easily bring it back.

Source/Atomic/Atomic2D/ParticleEmitter2D.cpp
```diff
@@ -323,7 +326,7 @@ void ParticleEmitter2D::Update(float timeStep)
 
 bool ParticleEmitter2D::EmitParticle(const Vector3& worldPosition, float worldAngle, float worldScale)
 {
-    if (numParticles_ >= (unsigned)effect_->GetMaxParticles() || numParticles_ >= particles_.Size())
+    if (!emitting_ || numParticles_ >= (unsigned)effect_->GetMaxParticles() || numParticles_ >= particles_.Size())
```
JimMarlowe: This will need Atomic guards, around the entire line is ok.

Source/Atomic/Atomic2D/ParticleEmitter2D.h
```diff
@@ -117,6 +117,15 @@ class ATOMIC_API ParticleEmitter2D : public Drawable2D
     /// Return sprite attribute.
     ResourceRef GetSpriteAttr() const;
 
+    // ATOMIC BEGIN
+
+    /// Set emission state.
+    void SetEmitting(bool enable);
+    /// Return emission state.
+    bool IsEmitting() const;
+
```
JimMarlowe: Its good to be consistent with the other emitter, even if the names are less than optimal.

Source/Atomic/Atomic2D/ParticleEmitter2D.cpp
JoshEngebretson: I misunderstood that this is a "bug" from upstream, sorry, yeah, leaving this would be fine in this case, otherwise, please guard it, or even better make a PR upstream with its removal, which would solve both problems :)

Alan-FGR: Well, that line makes no practical difference whatsoever, and you never know if compilers will be smart enough to remove that redundancy, probably not since they'd have to crawl throw other functions to check if they change that var between the two checks. @JoshEngebretson can I guard exclusions? If by 'upstream' you're talking about Urho, it's no redundant in the code there because we're the ones who modified the line 303 which rendered that check I removed redundant. It was changed from >= to > so I wonder why it's like that in the Urho code, this line: https://github.com/urho3d/Urho3D/blob/master/Source/Urho3D/Urho2D/ParticleEmitter2D.cpp#L315 will only step in when emissionTime_ is exactly zero what's kinda weird because counting on floats comparisons to exact values isn't exactly good practice but in this case since the value probably will never be a result of complex arithmetic I guess it's acceptable.
@JimMarlowe even if that is lost upon a rebase/merge at least we'll see the benefits of less redundancy in the mean time 🐫🐫.

Source/Atomic/Atomic2D/ParticleEmitter2D.cpp
Alan-FGR: Yea, missed that... next time I'm gonna run a comparer to make sure I'm not forgetting guards.

JoshEngebretson:
Awesome :)  Great add @Alan-FGR and thanks for the assist @JimMarlowe , LANDING! ✈️

Issue #993, opened 2016-09-02
Potential array out of bounds access in ParticleEffect2D


samoatesgames:
In ParticleEffect2D line 187, access is made to srcBlendFuncs and destBlendFuncs using the an index of 0 -> MAX_BLENDMODES.

MAX_BLENDMODES is 9, srcBlendFuncs and destBlendFuncs are arrays with 7 elements.
So potentially array element 8 could be accessed.

It looks like the srcBlendFuncs and destBlendFuncs are missing an entry, however I'm not sure what the value should be, else i'd make a PR.

Code in question:

```
static const int srcBlendFuncs[] =
{
    1,      // GL_ONE
    1,      // GL_ONE
    0x0306, // GL_DST_COLOR
    0x0302, // GL_SRC_ALPHA
    0x0302, // GL_SRC_ALPHA
    1,      // GL_ONE
    0x0305  // GL_ONE_MINUS_DST_ALPHA
};

static const int destBlendFuncs[] =
{
    0,      // GL_ZERO
    1,      // GL_ONE
    0,      // GL_ZERO
    0x0303, // GL_ONE_MINUS_SRC_ALPHA
    1,      // GL_ONE
    0x0303, // GL_ONE_MINUS_SRC_ALPHA
    0x0304  // GL_DST_ALPHA
};

enum BlendMode
{
    BLEND_REPLACE = 0,
    BLEND_ADD,
    BLEND_MULTIPLY,
    BLEND_ALPHA,
    BLEND_ADDALPHA,
    BLEND_PREMULALPHA,
    BLEND_INVDESTALPHA,
    BLEND_SUBTRACT,
    BLEND_SUBTRACTALPHA,
    MAX_BLENDMODES
};

for (int i = 0; i < MAX_BLENDMODES; ++i)
{
    if (blendFuncSource == srcBlendFuncs[i] && blendFuncDestination == destBlendFuncs[i])
    {
        blendMode_ = (BlendMode)i;
        break;
    }
}
```

Sam.

JimMarlowe:
[...]

PR #1243, opened 2016-12-05
Potential array out of bounds access #993

JimMarlowe:
This PR plugs a hole for a potential crash from calling ParticleEffect2D API, and fixes the ability for a ParticleEmitter2D to use the Duration setting in a pex file.

### DIFF
--- a/Source/Atomic/Atomic2D/ParticleEmitter2D.cpp
+++ b/Source/Atomic/Atomic2D/ParticleEmitter2D.cpp
@@ -296,8 +296,9 @@ void ParticleEmitter2D::Update(float timeStep)
             --numParticles_;
         }
     }
-
-    if (emissionTime_ >= 0.0f)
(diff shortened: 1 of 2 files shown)


Source/Atomic/Atomic2D/ParticleEmitter2D.cpp
```diff
@@ -296,8 +296,9 @@ void ParticleEmitter2D::Update(float timeStep)
             --numParticles_;
         }
     }
-
-    if (emissionTime_ >= 0.0f)
+    // ATOMIC BEGIN
+    if (emissionTime_ > 0.0f)
+    // ATOMIC END
```
JoshEngebretson: Good catch ⚾️

JoshEngebretson:
Thanks for the PR and getting Build 2 brewing :)

Issue #1205, opened 2016-11-26
[UI] Capture TurboBadger hashtable fix from repo


JoshEngebretson:
There may be other updates in the TB repo that can be rolled in, this hashtable one would be good:

https://github.com/fruxo/turbobadger/commit/898cc72c1c3e771b62aed887a7ee53a6ffb421dd

JimMarlowe:
I'll apply all PRs up to the current point in time. It looks like most of the PRs are from one guy, and there is not a lot of new functionality.

JimMarlowe:
Doing a clean reboot of TB is a difficult task, because of the changes required for scripting support. It just means new features will need to be hand merged. The TB rate of change has been small, so this hasn't been an issue.

JimMarlowe:
Upgraded turbobadger some, it includes commits/prs:
ee05368 API for multitouch
898cc72 Added some unit tests & fixed bug in TBHashTable.
TBWindow should have a virtual destructor 102
Fix TBImage::IsEmpty 93

I had to do some additional fixes for ee05368, I saw during compile that InvokePointerDown() and InvokePointerUp() were expecting to return bool, but they were still returning void, so I fixed them to the spirit in which they were intended.  This means that both touch systems will be installed, and It will be possible to transition from one to the other.

Tested on Android as best as I could, I tried the Roboman example, but it is crashing out on all my devices with 
```
E/Urho3D  ( 2721): JSErrorEvent:  : Line 90
E/Urho3D  ( 2721):  Name: TypeError
E/Urho3D  ( 2721):  Message: undefined not callable
E/Urho3D  ( 2721):  Stack:TypeError: undefined not callable
E/Urho3D  ( 2721): 	duk_js_call.c:776
E/Urho3D  ( 2721): 	anon DPad:90
E/Urho3D  ( 2721): 	anon DPad:58
E/Urho3D  ( 2721): 	anon Scripts/main:9 preventsyield
E/Urho3D  ( 2721): 	require  native strict preventsyield
D/dalvikvm( 2721): threadid=15: thread exiting, not yet detached (count=0)
```
I tried it also with an earlier Atomic version, and I get the same crash, so it's not these changes that makes it happen. I tried some of my own programs that use TB+touch for UI on android, an they worked correctly.

JoshEngebretson:
[...]

PR #1246, opened 2016-12-06
Capture TurboBadger hashtable fix and more #1205

JimMarlowe:
Upgraded turbobadger some, it includes commits/prs:
ee05368  API for multitouch
898cc72  Added some unit tests & fixed bug in TBHashTable. 
TBWindow should have a virtual destructor #102
Fix TBImage::IsEmpty #93

I had to do some additional fixes for ee05368, I saw during compile that InvokePointerDown() and InvokePointerUp() were expecting to return bool, but they were still returning void, so I fixed them to the spirit in which they were intended.

Tested on Android as best as I could, see #1205 for details.


JoshEngebretson:
Excellent, lots of good stuff pulled over 👍 I'll give this a once over and do some UI testing tomorrow on it too before landing. Thanks!

Source/ThirdParty/TurboBadger/tb_widgets.cpp
```diff
@@ -1593,6 +1635,98 @@ void TBWidget::HandlePanningOnMove(int x, int y)
...
+        {
+            TBWidgetEvent ev(EVENT_TYPE_TOUCH_CANCEL, 0, 0, true);
+            ev.ref_id = id;
+            ti->captured_widget->InvokeEvent(ev);
+        }
+        DeleteTouchInfo(id);
+    }
+}
+
+
```
JoshEngebretson: Wow, this touch stuff is going to come in handy on mobile controls! 👍

Type1J:
Touch is good for some laptops (Windows/Linux) as well. I wonder if a GeoSynth-like example with a dynamically generated saw, square or sine wave would be easy with Atomic after this PR.

JimMarlowe:
I can reproduce it, will fix.

JoshEngebretson:
@Type1J The new touch stuff needs to be wrapped in the UI subsystem and I think there are some TB cleanups (removals) that need to happen from some initial multitouch support added to Atomic back in 2015

Type1J:
I assumed that we'd go through a clean up period. It's unfortunate that Atomic has some duplicated work, but it's good to see it on a path to consolidation. :smile:

JimMarlowe:
I found the problem, I inserted the new touch events into the middle of the TB event enum, since that's where TB put them. Not knowing there is a parallel array for the UI, which is initialized with numbers, not the TB enum values, and this caused some odd behaviors, more would have popped up. I moved the new enums to the end of both lists, and all is well.

JoshEngebretson:
I had to switch from the tb enums to use raw numbers for C# constants binding purposes, definitely preferred the matching enum values.  Probably worth a big WARNING comment in the TB enum, that it must be synced with the enum in UI subsystem... and  I think I have used up my allocated use of the keyword ```enum``` in this comment :)   I'll do some further testing before landing, thanks!

JoshEngebretson:
Awesome, tests out, landing ✈️ , thanks! 🐫 ☀️ 😄

PR #1250, opened 2016-12-07
JS- inspector plugin support

JohnnyWahib:
This will now allow the user to pass customized inspectors through the TS plugin system ( I will PR an example of how it gets used in the existing Atomic plugin example project shortly)

It comes with a change to recognize files with unknown extensions as .assets by using the default AssetImporter types.( I have left this change in a separate commit to be safe)

What this will allow us to do is fire up custom inspectors for currently unsupported file formats by checking the file's extension.


JoshEngebretson:
Looks great and only 55 lines of TypeScript!  Thanks, landing! 👍

PR #1256, opened 2016-12-07
Virtual DPad JS script is throwing exception

JimMarlowe:
Tracked it down to Input::BindButton() being evaporated from the source, this restores it.

Tested successfully on Android and on Desktop with Atomic.input.touchEmulation = true;


JoshEngebretson:
Awesome, thanks for tracking this down, will update site download with hotfix once clear and can build

PR #1259, opened 2016-12-07
Update TS to v 2.1 - fixes #1254

### LINKED ISSUE #1254: [TypeScript] Upgrade TypeScript to v2.1
Update the internal TypeScript transpiler to v2.1

See: https://blogs.msdn.microsoft.com/typescript/2016/12/07/announcing-typescript-2-1/

shaddockh:


JoshEngebretson:
Great, thanks for the PR, landing!

rosshadden:
I wasn't sure if it made sense to say this here, or in a new issue, but I found a minimally-reproducible issue when using TS 2.1.  This issue prevents Atomic games from working, including the base 3d boilerplate project.

- If you create a brand new 3d Typescript project, it runs as expected.
- If, with `typescript@2.1.1` or later (the current version at time of this writing is `v2.1.4`), you run `tsc` to compile the project, it no longer runs.  This includes the absolute latest version, `typescript@next`.
- If, with `typescript@2.0.10` or earlier, you run `tsc` to compile the project, it runs as expected.

I believe through code diffs that the code Atomic's base 3d project has generated on creation was done with typescript `v2.0.x`, which would explain why it works.

Speaking of diffs, this is the diff between the `Spinner.js` file generated with `v2.0.10` (or in this case the one that came directly from creating the base 3d project) (left) and `v2.1.1` (right):



As you can see, there are four changes.  Of those, the bottom one (`Object.defineProperty`) is insignificant.  The other three changes, where it messes with references to `this` and `super()`, cause the game to fail.

JoshEngebretson:
@rosshadden Thanks for the report, I believe this fixes the issue though I have not updated the site binaries quite yet with it: https://github.com/AtomicGameEngine/AtomicGameEngine/pull/1303

rosshadden:
Ah, yeah it looks like that should do the trick.  Thanks!

rosshadden:
Just for some closure, I built the latest version of Atomic from source, and this is indeed fixed.  Thanks again to you and @shaddockh.

PR #1263, opened 2016-12-08
Default AssetImporter.requiresCacheFile to false

mattbenic:
Following @JohnnyWahib's change to allow non-recognized assets to be imported, we ran into an issue with our modified (async) cache code. RequiresCache defaults to true, which is inappropriate for these unrecognized assets, which hits an assert when we collect the details of expected cache files per asset.

The fix for this (defaulting requiresCacheFile to false and setting it to true in appropriate cases) isn't strictly necessary for default Atomic, but it does seem to be the correct approach, so PRing it as well.

### DIFF
--- a/Source/ToolCore/Assets/AssetImporter.cpp
+++ b/Source/ToolCore/Assets/AssetImporter.cpp
@@ -31,7 +31,7 @@ namespace ToolCore
 
 AssetImporter::AssetImporter(Context* context, Asset *asset) : Object(context),
     asset_(asset),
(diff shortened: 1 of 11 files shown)


JoshEngebretson:
Great, it looks like nuget had an outage on the CI, so will kick it off again ```The remote name could not be resolved: 'api.nuget.org'```

JoshEngebretson:
:bomb: Boom! Landing, thanks!

PR #1266, opened 2016-12-08
Support for constructors in the Web subsystem

Type1J:
This patch adds support for constructors instead of forcing use of the "make" functions of the Web subsystem. When the Web subsystem was initially added constructors were not yet bound automatically. Now that they are, it would be surprising if using the WebRequest and WebSocket didn't work, if they were directly constructed. They probably should have had private constructors, but this patch removes the need for private constructors.


JoshEngebretson:
Great, thanks for the PR, landing! ✈️

Issue #1269, opened 2016-12-08
[Network] Resurrect additions for master server/client


JoshEngebretson:
Some additions in Connection.h and perhaps elsewhere for master server were lost in update, resurrect them. 

Commit which loses additions: https://github.com/AtomicGameEngine/AtomicGameEngine/commit/47fd8f9abbd821e8be5c347d705f6cea83d511f9

JoshEngebretson:
Closed via #1272

PR #1272, opened 2016-12-09
[Network] Resurrect additions for master server/client #1269

JimMarlowe:
Restored functionality, all wrapped in // ATOMIC guards
Tested on linux with the multiplayer spacegame example. 
A note about the master server, the list of included servers is getting pretty large, it needs some aging or unregistering functions.


JimMarlowe:
I also tried having a (built) MP Spacegame server on linux, and a Android client, and was able to connect the two! Sadly, the spacegame has no mobile controls, but very nice to see the nat punchthru in action.

JoshEngebretson:
Excellent work, it will help having the code blocks marked moving forward, I thought I had gone back and reapplied these.  I was wrong, thanks for the pick up and nat testing!  👍 LANDING!

PR #1274, opened 2016-12-10
Ct vcolor import

CTrauma:
A way to set the correct technique for a imported model with vertex colors.


JoshEngebretson:
Thanks for the squeaky clean PR, vertex colors are a great add 👍

PR #1280, opened 2016-12-14
Updates

JoshEngebretson:
- Web subsystem typed events, web request post data setter and convenience response getter (which also caches the response, otherwise you would get one shot at reading the BufferQueue, oops), some misc cleanups as well (also adjusted the AtomicExamples for the new event names)
- Adds Contributors and Vendor information to About window in editor


JoshEngebretson:
@Type1J I added info for WebRequest events, which changes the string hash used for the event in JS code (having events like "completed" are ambiguous and will confuse the event system).  Just a heads up if you need to adjust the string for subscribers in JS code, in C++ the E_EVENTNAME from the event declaration is used.

Type1J:
Looks good. Should something similar be done to WebSocket?

Source/Atomic/Web/WebRequest.cpp
```diff
@@ -401,11 +411,15 @@ void WebRequest::Send()
 {
     if (!is_->isAddedToMulti && !is_->isAborted)
     {
-        is_->es_hold = this;
+        // Add a reference to ourselves during the Send, this is released
+        // in notifyInternal,  if we're leaking WebRequests check that
+        // this is being called
+        AddRef();
+
```
Type1J: @JoshEngebretson Is there a reason to count manually? I'm just curious.

JoshEngebretson:
Currently on day 2.5 of no power, 15,000 households still without, so not looking good.  Everything is on electric, so no heat, hot water, etc :/

Yes, websocket events should be updated. 

The shared ptr in the internal rep had lifetime issues and was a bit funky, internally it was just adding ref and this is clearer.

We could hold inflight requests in the web subsystem which would provide better querying and lifetime control... 

Sending from phone will see if it comes through in issue

Issue #1236, opened 2016-12-04
[AtomicNET] C# code changes require manual recompilation in AtomicEditor


toddgibson:
Posted in Gitter chat:

@toddgibson 00:23
I'm not sure if I'm doing something wrong or it's a bug, but when working on a C# script on PC or Mac (Visual Studio 2015 CE, Xamarin Studio 6.x) if I make a change to the script in VS/XS and then go back and run the game from the AtomicEditor it doesn't see those script changes. Running it from either VS/XS works, but I'm not sure why they aren't syncing up. Is this known or am I missing a step?

@toddgibson 00:37
oh, I had to manually go into AtomicEditor menu and click Developer->Plugins->AtomicNet->Compile Project and then it synced up again... seems like it should watch those files and auto-compile or something.
Even with "Compile on Save" turned on I get the same result.

Alan-FGR:
I can confirm this on Windows... I guess "Compile on Save" works only for the built-in editor but I'm not sure, in any case it doesn't work when files are changed on disk and that's definitely a major when you play from the editor. Imho this should be either hotfixed or at very least should be fixed for build2.

JoshEngebretson:
Compile on Save was initially hooked up for the embedded Monaco editor, though there has since been Visual Studio/Xamarin Studio/MonoDevelop integration added, and the feature was disabled.  Clicking any C# source in the project should automatically open the file in the IDE where you can edit, compile, debug, deploy, etc.  Compiling in the IDE should also update the play mode in the editor.

I am thinking if a C# file under resources has been modified, bring up a simple dialog on play which asks whether to recompile.  It is possible this can be done automatically if no IDE is attached to the editor process, what we want to avoid dueling builds between the IDE and the Atomic Editor.

Of note, Atomic also supports C# sources, assemblies, etc outside of the resource tree for external projects.  Where there is no file notification. https://github.com/AtomicGameEngine/AtomicTests/tree/master/CSharpTests/ExternalDllTest

I will add this to Build 2, which is going to need some combing for priorities: https://github.com/AtomicGameEngine/AtomicGameEngine/milestone/1

Alan-FGR:
[...]

PR #1281, opened 2016-12-14
C# code changes require manual recompilation in AtomicEditor #1236

JimMarlowe:
This PR fixes the Play button for a C# project to always be up to date with the latest saved source files.  There is a check added to the play event handler, which looks to see if the current project is c#, then does a check for out of date sources. If the sources are ok, then no further action is needed, it simply plays the project. If the sources are out of date, then the BuildAndPlay processing starts and it returns out of the play event handler.  The BuildAndPlay function then compiles the project, and in the compile completed event handler, if successful, issues a Play event and the project is played after a short pause. If the compilation fails, only an error is put in the console, and the project is not played. Because of the fine work someone did when c# is compiling, the play button greys out while this compile happens also, giving the user an indication something is happening.

### DIFF
--- a/Source/ToolCore/NETTools/NETProjectSystem.cpp
+++ b/Source/ToolCore/NETTools/NETProjectSystem.cpp
@@ -29,6 +29,9 @@
 #include <Atomic/IO/FileSystem.h>
 #include <Atomic/Resource/ResourceEvents.h>
 
+#include <Atomic/UI/UIEvents.h>
+#include <Atomic/Input/InputEvents.h>
+
 #include "../ToolSystem.h"
 #include "../ToolEnvironment.h"
 #include "../Assets/AssetEvents.h"
@@ -214,6 +217,7 @@ namespace ToolCore
 
             if (build)
             {
+                build->UnsubscribeFromEvent(E_NETBUILDRESULT); //unsubscribe in case it has been subscribed before
                 build->SubscribeToEvent(E_NETBUILDRESULT, ATOMIC_HANDLER(NETProjectSystem, HandleNETBuildResult));
             }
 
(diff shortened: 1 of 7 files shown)


Source/ToolCore/NETTools/NETProjectSystem.cpp
```diff
@@ -214,6 +217,7 @@ namespace ToolCore
 
             if (build)
             {
+                build->UnsubscribeFromEvent(E_NETBUILDRESULT); //unsubscribe in case it has been subscribed before
                 build->SubscribeToEvent(E_NETBUILDRESULT, ATOMIC_HANDLER(NETProjectSystem, HandleNETBuildResult));
```
JoshEngebretson: It is only possible to have one subscription to an event, so this is unnecessary as the previous one will be overwritten

Source/ToolCore/NETTools/NETProjectSystem.cpp
```diff
@@ -421,6 +425,96 @@ namespace ToolCore
...
+
+            Project* project = GetSubsystem<ToolSystem>()->GetProject();
+            NETBuildSystem* buildSystem = GetSubsystem<NETBuildSystem>();
+            if (buildSystem)
+            {
+                NETBuild* build = buildSystem->BuildAtomicProject(project);
+                if (build)
+                {
+                    build->UnsubscribeFromEvent(E_NETBUILDRESULT); //unsubscribe in case it has been subscribed before
+                    build->SubscribeToEvent(E_NETBUILDRESULT, ATOMIC_HANDLER(NETProjectSystem, HandleNETBuildPlay));
```
JoshEngebretson: Here again on the no need to unsubscribe

Source/ToolCore/NETTools/NETProjectSystem.cpp
```diff
@@ -421,6 +425,96 @@ namespace ToolCore
...
+        if (myass)
+            dllTimestamp = myass->GetFileTimestamp();  
+        else 
+            return true; // dll not there, needs to be built, or the sources dont compile, or some other error
+        
+        int ii = 0, nn = 0;
+        Asset* filex = NULL;
+        StringVector results;
+        const String searchdir[3] = { "Scripts/", "Modules/", "Components/" }; // only search sanctioned project dirs
+        for ( ii=0; ii<3; ii++)
```
JoshEngebretson: I think we need check under all of resources for cs, I think the project generator globs for all .cs (recursively), need to verify

Source/ToolCore/NETTools/NETProjectSystem.cpp
```diff
@@ -421,6 +425,96 @@ namespace ToolCore
...
+    {
+        using namespace NETBuildResult;
+        
+        if (eventData[P_SUCCESS].GetBool())
+        {
+            ATOMIC_LOGINFOF("NETBuild Success for project, now playing.");
+            VariantMap shortcutData;  // encode a shortcut even to send a play
+            shortcutData[UIShortcut::P_KEY] = KEY_P;
+            shortcutData[UIShortcut::P_QUALIFIERS] = QUAL_CTRL;
+            SendEvent(E_UISHORTCUT, shortcutData);
```
JoshEngebretson: Artful workaround for the mess in "invokePlayOrStopPlayer", good job and nice catch on the programmatic qualifier in the event handler 👍

Source/ToolCore/NETTools/NETProjectSystem.cpp
```diff
@@ -421,6 +425,96 @@ namespace ToolCore
...
+        }
+        else
+        {
+            const String& errorText = eventData[P_ERRORTEXT].GetString();
+
+            ATOMIC_LOGERRORF("\n%s\n", errorText.CString());
+            ATOMIC_LOGERRORF("NETBuild Error for project, will not play.");
```
JoshEngebretson: It would be good to output an error modal with this info, should be pretty easy

Source/ToolCore/NETTools/NETProjectSystem.cpp
```diff
@@ -421,6 +425,96 @@ namespace ToolCore
...
+                }
+            }       
+        }
+        
+        return shouldRebuild;
+    }
+    
+    /// interrogate sources to see if any are dirty compared to the editor project.dll
+    bool NETProjectSystem::CheckForRebuild() 
+    {
```
JoshEngebretson: This method is feeling a bit fragile, which makes sense as the project dll handling needs tightening and we need standard ways to query project bits like the project assembly

Source/ToolCore/NETTools/NETProjectSystem.cpp
```diff
@@ -421,6 +425,96 @@ namespace ToolCore
...
+        FileSystem* fileSystem = GetSubsystem<FileSystem>();
+        AssetDatabase* db = GetSubsystem<AssetDatabase>();
+        Project* project = GetSubsystem<ToolSystem>()->GetProject();
+        unsigned dllTimestamp = 0;
+
+        String mydll = project->GetResourcePath() + project->GetProjectSettings()->GetName() + ".dll";
+        Asset* myass = db->GetAssetByPath(mydll);
+        if (myass)
+            dllTimestamp = myass->GetFileTimestamp();
```
JoshEngebretson: *smirk*

Source/ToolCore/NETTools/NETProjectSystem.cpp
```diff
@@ -421,6 +425,96 @@ namespace ToolCore
 
     }
 
+    /// handle the results of a recoompile, and auto play the project if successful
```
JoshEngebretson: Type on ```recoompile```

Source/ToolCore/NETTools/NETProjectSystem.cpp
```diff
@@ -421,6 +425,96 @@ namespace ToolCore
...
+    void NETProjectSystem::HandleNETBuildPlay(StringHash eventType, VariantMap& eventData)
+    {
+        using namespace NETBuildResult;
+        
+        if (eventData[P_SUCCESS].GetBool())
+        {
+            ATOMIC_LOGINFOF("NETBuild Success for project, now playing.");
+            VariantMap shortcutData;  // encode a shortcut even to send a play
```
JoshEngebretson: even -> event

Source/ToolCore/NETTools/NETProjectSystem.cpp
JimMarlowe: ok, so this isn't wxWidgets, where you can hang multiple event handlers on an event.

Source/ToolCore/NETTools/NETProjectSystem.cpp
JimMarlowe: ok, changed, you never know if kids are going to be reading the source code.

Source/ToolCore/NETTools/NETProjectSystem.cpp
JimMarlowe: It does depend on the existing project structure, you need to be able to trust the assetdatabase. There is always room for improvement.

Source/ToolCore/NETTools/NETProjectSystem.cpp
JimMarlowe: I'll search the entire project tree.

Source/ToolCore/NETTools/NETProjectSystem.cpp
JimMarlowe: > should be pretty easy
hah, I found the perfect vehicle with the E_EDITORMODAL event! Coded it up and ... there is no event handler for it. I'll add something.

JimMarlowe:
All issues fixed.

Alan-FGR:
Isn't this a dupe of another (possibly 2) other issues? We should link the duplicates imho.

Alan-FGR:
#1236?

Alan-FGR:
:facepalm: Holy geez... please forgive me, I really confused some stuff here.

JoshEngebretson:
Thanks, looking forward to having a look (and power, heat, lights, hot water) soon :)

JoshEngebretson:
@AtomicBuildBot Retest this please

JoshEngebretson:

JoshEngebretson:
I've been doing some testing and it is working great, especially like that we get to see C# compile errors in the editor now.  I'll verify that VS also has similar error text to truncate all the build spam 👍 

One note, when there is a build required, the play button switches state 2x. Here's a gif of what I am seeing:

compilebeforeplay

Do you think this is easy to fix?

JimMarlowe:
I dont know, I'll look.

JimMarlowe:
Well, I can't reproduce the double greying on Linux. There is only one place that sends the E_NETBUILDBEGIN event and that is in NETBuildSystem::HandleToolUpdate(), so that means it's going thru here twice on OSX, Win.  My computer must be good (or poor) at finishing a build and bails out of the next HandleToolUpdate() before emitting another E_NETBUILDBEGIN.  If the timing of your gif is accurate, then OSX and Win may start another build, but decides down the line it is up to date and finishes without recompiling.

JoshEngebretson:
Ok, building locally and tracking it down, I'll land in a new PR once sorted, thanks for the leads 🔍

PR #1285, opened 2016-12-18
Lint ts and fix ATOMIC_PARAMs

JimMarlowe:
Fixes the Typescript lint warnings, so the genscripts works better. Also fixes ATOMIC_PARAM syntax, since the js binding generator actually parses the comment after the definition for the arg type, otherwise it emits an odd warning.


JoshEngebretson:

JoshEngebretson:
Thanks for sweeping up around the place :)  LANDING!

PR #1288, opened 2016-12-20
[C#] On demand editor solution compilation, including play mode

### LINKED ISSUE #1236: [AtomicNET] C# code changes require manual recompilation in AtomicEditor

JoshEngebretson:
Closes #1236 

@JimMarlowe I ended up refactoring this a bit, one thing to watch is that the ToolCore module doesn't become dependent on the AtomicEditor.  It is core tool functionality shared by AtomicTool, AtomicEditor, etc.  I missed this on the initial review, easy to do :)

Another note, If you find yourself adding UI stuff, like events, into ToolCore, that is also a sign that they need to be hooked up in the editor instead.   

Whew, we have way more solid C# project workflow, so sayeth the wise 🐫  😄


JimMarlowe:
Is good, I will make notes.  I attempted to keep like things together... but there are times when the architecture is not obvious and then you've got this :hammer: and an error message.

Issue #1277, opened 2016-12-12
[TypeScript] import statements w/o path specified don't resolve


shaddockh:
if you import another module from a component and don't provide a path, it should base the path on the root of the Resources directory, ie:
```typescript
import * as myModule from "Scripts/mymodule";
```

should find that file at ```<game project>/Resources/mymodule.ts```
However, the TypeScript compiler complains that the file cannot be found.

If you do the same in vanilla js, it works
```javascript
var mymodule = require("Scripts/mymodule");
```

The routines that setup and invoke the TypeScript compiler need to be able to handle this situation.  The workaround is to specify relative paths, such as:
```typescript
import * as myModule from "./mymodule";
//or
import * as myModule from "../Modules/mymodule";
```

JoshEngebretson:
Closed via #1293

PR #1293, opened 2016-12-21
Fix for #1277 - Set the base URL for typescript files

shaddockh:
to be the Resources directory so that you can import non-relative modules.  

Should be able to ```import Foo from “Scripts/Foo”``` in addition to using ```import Foo from "./Foo"```


JoshEngebretson:
Gotta love the one liners :) Thanks! 👍

PR #1294, opened 2016-12-21
Fixes #1219 - improvements for working with vscode + some cleanup

### LINKED ISSUE #1219: Typescript improvements for integration with vscode
There needs to be some improvements to the TypeScript integration to make working with vscode a little easier
* The ```allowNonTsExtensions``` setting is not compatible with the command line tsc, so need to either determine if this is no longer needed for the in-editor compiler, and don't emit this when generating an external ```tsconfig.json```
* when generating the ```tsconfig.json```, don't specifically use the ```files``` block.  Instead, use the ```include``` block.  When consuming the ```tsconfig``` in the editor, ignore that block and re-generate the ```files``` block.  This will keep from needing to manually add files to the tsconfig in vscode.
* possibly give the option to build out a tasks.json that will allow the TypeScript compilation step from vscode
* relook at the ```typings``` folder.  Maybe move ```ambient``` up right below the ```typings``` level.

shaddockh:
This PR does the following:
- remove the deprecated ```allowNonTsExtensions```
- When copying the ```Atomic.d.ts``` file into the project, put it under ```typings/ambient/atomicgameengine``` instead of ```typings/main/ambient/atomicgameengine```
- Auto-generate a ```tasks.json``` file in the ```.vscode``` directory so that VSCode can compile the TS w/o any additional setup
- Updated the ```tsconfig``` file with the ```rootUrl``` properly so non-relative paths work
- clean up some ```var``` vs ```const``` issues

### DIFF
--- a/Script/AtomicEditor/hostExtensions/languageExtensions/TypscriptLanguageExtension.ts
+++ b/Script/AtomicEditor/hostExtensions/languageExtensions/TypscriptLanguageExtension.ts
@@ -37,7 +37,6 @@ const defaultCompilerOptions = {
     inlineSourceMap: false,
     removeComments: false,
     noLib: false,
-    allowNonTsExtensions: true,
     forceConsistentCasingInFileNames: true,
     allowJs: true,
     lib: ["es5"]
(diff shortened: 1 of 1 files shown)


JoshEngebretson:
Thanks for the awesome PR 👍  :)

Issue #1301, opened 2016-12-23
[TypeScript] Embedding InspectorFields inside component no longer works


shaddockh:
It looks like a new version of the TS compiler changed the way it handles internal object literals.  Previously:

```typescript
class Foo extends Atomic.JSComponent {
   inspectorFields = {
      speed: 1.0
  };
...
}
```
would work and would get converted in the JS to ```this.inspectorFields = ...```.

Now, it gets converted to ```_this.inspectorFields = ...``` which means it's no longer seen by the inspector.

There are some options:
1. Do not allow the ```inspectorFields``` to reside in the component and instead define it outside the component, but within the same module.  Need to update all the examples and project templates to handle it this way.
2. Specifically look for ```_this.inspectorFields``` in addition to ```this.inspectorFields```
3. Make the process that looks for the inspector fields actually instantiate the component and use normal JS methods to grab the contents of the inspector fields 

I'm opting for either 2 or 3.  3 would be more robust to future changes, but may require more work.  The quick fix would be 2 and then follow up with 3 for the long term solution.

JoshEngebretson:
Closed via #1303

JoshEngebretson:
Reopened as not solved via #1303 I don't think...

shaddockh:
From what I can tell, this is working correctly.  What are you seeing?

rosshadden:
It's working fine for me, too.

JoshEngebretson:
I read "quick fix" on #1303 and thought there was still some consideration being put to options 1/3, so didn't want to close it out.  If this is working, all for the close option :)

shaddockh:
ok, yeah that makes sense.  I'll break out option 3 as a new issue, but that is going to take some thinking about.  There may be dragons.

PR #1303, opened 2016-12-27
Quick fix for #1301 - TS generated inspector fields blocks

shaddockh:
TS now generates the inspector field blocks as ```_this.inspectorFields``` instead of ```this.inspectorFields```.  This looks for this additional pattern when trying to find inspector fields.


JoshEngebretson:
Good catch, we'll release a hot fix with this update 👍

PR #1304, opened 2016-12-27
Convert to TS style enums

shaddockh:
This adds in TS style enums for all of the enumerated types.  It also retains the original JS style constants, but comments them to indicate that they should only be used with vanilla JS.  The Editor code has been updated to take advantage of the TS style enums everywhere.

### DIFF
--- a/Script/AtomicEditor/editor/Preferences.ts
+++ b/Script/AtomicEditor/editor/Preferences.ts
@@ -67,7 +67,7 @@ class Preferences {
 
     addColorHistory(path: string): void {
         var index = this._prefs.colorHistory.indexOf(path);   // search array for entry
-        if (index >= 0) {                                     // if its in there, 
+        if (index >= 0) {                                     // if its in there,
             this._prefs.colorHistory.splice(index, 1);        // REMOVE it.
         }
         this._prefs.colorHistory.unshift(path);               // now add it to beginning of array
@@ -103,7 +103,7 @@ class Preferences {
         }
 
         //Read file
-        jsonFile = new Atomic.File(filePath, Atomic.FILE_READ);
+        jsonFile = new Atomic.File(filePath, Atomic.FileMode.FILE_READ);
 
         var prefs = null;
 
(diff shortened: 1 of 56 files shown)


JoshEngebretson:
The legacy JS enums are good to keep for compatibility with current JS examples.  Do you think we should deprecate the legacy enums (I guess saying they are "legacy" implies we should) and move the JS examples over to the new enum naming, so consistent between TS/JS, or is the "legacy" style a better fit for JS code?  I lean towards the consistency being preferable...

Excellent, thanks for the PR, LANDING :) ✈️

shaddockh:
Unfortunately, the TS style enums are not usable in JS, so the JS examples are going to need to stay as they are.  The TS transpiler, when it sees a TS style enum, will actually replace it with the numeric constant associated with that enum in the generated JS.

JoshEngebretson:
Can we declare native enums as ambient so they are computed instead of using the numerical constant value provided by the compiler.  This is a win with native enum values as they may not be initialized by numeric symbols in C++ headers (something the C# bindings have to deal with)

If we can do this, then we can generate objects/properties for consistent JS/TS resolution, right?  For example: ```Atomic.FileMode``` would be an object with a ```FILE_WRITE``` property in JS:

```Atomic.FileMode.FILE_WRITE```

I am hopeful that ambient enum declarations are what we need: https://www.typescriptlang.org/docs/handbook/enums.html

shaddockh:
Yes, I think that would be possible, so instead of defining them as ```const enum```, they would just be defined as ```enum```.  Then on the C++ side, there would need to be a mapping between the JS object literal property and the internal C++ enum as you described.

How is the best way to convert all those enums to object literals on the native side instead of just constants hanging off ```Atomic```?

JoshEngebretson:
We need to populate module enum objects with properties from native code, a nice thing is that we can use the native symbolic values and will be kept in sync, so the transpiled source isn't tied quite so closely to the binary. 

I think this is the right move, though it is a breaking change, so we should leave the existing enum values in JS in place finding some way to deprecate them.  There is a similar issue in C# with the Constants.* needing to be broken up into discrete enums

I filed #1305 for JS enum objects, do you have any thoughts on how to deprecate JS API, has to be at runtime right? Hmmmmm :)

shaddockh:
TS doesn't have any kind of way to deprecate something currently - see: https://github.com/Microsoft/TypeScript/issues/390

I think the best bet at this point is to just emit ```/** @deprecated */``` as the comment above these items so they come out in the docs and the intellisense.

PR #1309, opened 2017-01-02
Added `include` block to tsconfig.

rosshadden:
This makes typescript only scan the relevant directories, which should make it faster to build.


JoshEngebretson:
Thanks for the PR 👍 

rosshadden:
In retrospect, I am going to remove this.  I just set up a project such that the typescript source files were in a different directory than `Resources`, and it took me a bit to realize the reason it wasn't working was the include block not whitelisting my new directory...  Due to the lack of helpful errors or output when trying to debug something like that, it should remain a power-user feature that we add ourselves.

PR #1312, opened 2017-01-03
Expose PackageFile to scripting

JimMarlowe:
This PR exposes the PackageFile object so the scripting languages can load and unload them, and examine the contents. 
I had to create a new accessor, PackageFile::GetEntryName(unsigned) because the existing PackageFile::GetEntryNames() did not want to be bound.


Source/Atomic/IO/PackageFile.h
```diff
@@ -86,7 +86,20 @@ class ATOMIC_API PackageFile : public Object
     const Vector<String> GetEntryNames() const { return entries_.Keys(); }
 
     // ATOMIC BEGIN
-    
+
+    /// Return a file name in the package, for scripting use ( GetEntryNames is not bindable? )
+    const String& GetEntryName(unsigned index) const 
+    {
+        unsigned nn = 0;  // .Begin() + index syntax not supported on HashMap
+        for (HashMap<String, PackageEntry>::ConstIterator j = entries_.Begin(); j != entries_.End(); ++j)
```
JoshEngebretson: ```Vector<String>``` won't be picked up currently by bindings (and in general marshaling by an index accessor method can have some benefits over copying arrays/Vectors in terms of speed and dealing with outdated data).    

How about "Return a file name in the package at the specified index" as comments do show up in documentation?

I had a bit of a think on this too, as I am not sure we can count on the HashMap iterator being ordered, though if the package file is being modified while you are iterating, you probably have bigger problems :)

Script/Packages/Atomic/Scene.json
```diff
@@ -8,7 +8,8 @@
 	"excludes" : {
 		"Scene" : {
 			"GetComponent" : ["unsigned"],
-			"MarkReplicationDirty" : ["Node"]
+			"MarkReplicationDirty" : ["Node"],
+			"GetRequiredPackageFiles" : []
 		}
```
JoshEngebretson: What was the issue with GetRequiredPackageFiles?  I believe this signature should be picked up by the C# bindings (JS/TS don't have Vector instance support yet)

Script/Packages/Atomic/Resource.json
```diff
@@ -1,11 +1,16 @@
...
 	"classes" : ["Resource", "ResourceCache", "XMLFile", "PListFile", "JSONFile", "Image", "ResourceNameIterator"],
 	"overloads": {
 		"Image": {
 			"GetPixel": ["int", "int"],
 			"SetSize": ["int", "int", "int", "unsigned"]
+		},
+		"ResourceCache": {
+			"AddPackageFile": ["String", "unsigned"],
+			"RemovePackageFile": ["String", "bool", "bool"]
 		}
```
JoshEngebretson: Nice 👍

Source/Atomic/IO/PackageFile.h
JimMarlowe: Comment changed.  The HashMap has an order-ish and the const iterating should provide valid values for each index for the current state of the PackageFile. The object could change, and this method will not guarantee an item will remain at a specific index. GetEntryNames() only gives you a pile of names without index context, so maybe index is a bad term.

Script/Packages/Atomic/Scene.json
JimMarlowe: GetRequiredPackageFiles() was getting an error trying to convert a PackageFile pointer to an AObject something, which won't happen. So I excluded it, the loss from not having it scripted will be minimal. It is only used in the client-server scene+packagefile replication internals, which are also not exposed to scripting.

JoshEngebretson:
Ok great, thanks for the updates, LANDING! ✈️

PR #1327, opened 2017-01-07
Scene view 3d move camera with the middle mouse move

balmacefa:
use the same funcionality as the Q,E,A,D Keys to move the camera. @JoshEngebretson @rosshadden

### DIFF
--- a/Source/AtomicEditor/Editors/SceneEditor3D/SceneView3D.cpp
+++ b/Source/AtomicEditor/Editors/SceneEditor3D/SceneView3D.cpp
@@ -182,6 +182,11 @@ bool SceneView3D::GetChangingCameraSpeed()
     return MouseInView() && input->GetMouseMoveWheel() && input->GetMouseButtonDown(MOUSEB_RIGHT);
 }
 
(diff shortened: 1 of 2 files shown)


balmacefa:
@JoshEngebretson,The lines are indented

JoshEngebretson:
@balmacefa I'll give this a try a bit later and see if we can sneak it in for Build 3, looks cool

JoshEngebretson:

JimMarlowe:
The functionality for this PR is ok. The only problem I encountered in it's use, was from the camera speed.  It has no visible setting to show the user what it is, and is set by the right mouse button + mousewheel, according to https://github.com/AtomicGameEngine/AtomicGameEngine/wiki/Scene-Editor-Hotkeys. With no feedback that it is mouse speed is changing, you can do so, and then moving with camera with the middle mouse button can change dramatically, with no obvious way to recover the old setting.

balmacefa:
I propose a UI indicator of the status of the view https://github.com/AtomicGameEngine/AtomicGameEngine/issues/1330 , maybe we can include a speed indicator

PR #1329, opened 2017-01-07
Toggle to Play camera

balmacefa:
press C key to change between the __atomic_sceneview3d_camera and scene camera. Select a camera and press C to chage to that camera, if you dont select a camere, we change to the first camera compoment in the scene, if there is no camera switch to default (__atomic_sceneview3d_camera), lock the camera pos, rot with L key. Every time it is passed to a user's camera, it is blocked

### DIFF
--- a/Source/AtomicEditor/Editors/SceneEditor3D/SceneView3D.cpp
+++ b/Source/AtomicEditor/Editors/SceneEditor3D/SceneView3D.cpp
@@ -86,7 +86,9 @@ SceneView3D ::SceneView3D(Context* context, SceneEditor3D *sceneEditor) :
     perspectCamPosition_(0, 0, 0),
     perspectiveYaw_(0),
     perspectivePitch_(0),
-    fromOrthographic_(false)
+    fromOrthographic_(false),
+    isOnPlayCamera_(false),
+    lookCamera_(false)
 {
 
     sceneEditor_ = sceneEditor;
@@ -121,6 +123,11 @@ SceneView3D ::SceneView3D(Context* context, SceneEditor3D *sceneEditor) :
     assert(octree_.NotNull());
 
     cameraNode_->SetPosition(Vector3(0, 0, -10));
+
+    //save the default camera Ptr
+    defaultCameraNode_ = cameraNode_;
+
+    activeCameraNode_ = defaultCameraNode_;
 
     SetView(scene_, camera_);
     SetAutoUpdate(false);
@@ -240,7 +247,7 @@ void SceneView3D::MoveCamera(float timeStep)
     Quaternion q(pitch_, yaw_, 0.0f);
 
     if (!zooming || !changingCameraSpeed)
-        cameraNode_->SetRotation(q);
+        activeCameraNode_->SetRotation(q);
 
     if (orbitting)
     {
@@ -250,8 +257,8 @@ void SceneView3D::MoveCamera(float timeStep)
         if (bbox.Defined())
         {
             Vector3 centerPoint = bbox.Center();
-            Vector3 d = cameraNode_->GetWorldPosition() - centerPoint;
-            cameraNode_->SetWorldPosition(centerPoint - q * Vector3(0.0, 0.0, d.Length()));
+            Vector3 d = activeCameraNode_->GetWorldPosition() - centerPoint;
+            activeCameraNode_->SetWorldPosition(centerPoint - q * Vector3(0.0, 0.0, d.Length()));
 
         }
     }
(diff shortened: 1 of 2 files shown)


JoshEngebretson:
Thanks for the pull requests, will have a look at them soon

Alan-FGR:
@JoshEngebretson may need some refactor for capitalization and also the tabs :P. Not too sure on naming too, maybe could be more descriptive. Also, I don't see the need to maintain a bool separately for that @balmacefa. Isn't it better to just check if the editor camera is being used every time since you have actualCameraNode_? Just some random thoughts, otherwise good work!

Alan-FGR:
I also don't see the need for the lookCamera_... can't we just check that in L297? I mean, just check if the actualCameraNode_ is the defaultCameraNode_ and you don't need those bools at all... btw, I recommend changing those var names to something like 'activeCameraNode' or 'currentCameraNode' and 'sceneCameraNode' or something else more descriptive. I really couldn't tell right away what those are for.

Alan-FGR:
Also, we got hardcoded keys all over the place :P.

balmacefa:
Sorry for the naming of the variables, my English is bad :(, the reason I use the lockcamera boolean is if you want to move the camera (user once) for better framing, but I think is usedful to move it just we you wanna to. The actualCamera var is a dirty solution, to avoid saving the pos rot of the user's camera, if you change the cameraNode_ and save the scene in a user camera, it take like is the default camera

balmacefa:
Maybe if you received a event went the scene is saved, we can change the cameraNode_ ref to the default

Alan-FGR:
@balmacefa No need to apologize man! It's just constructive criticism :). Let's wait for further input from @JoshEngebretson then.

JoshEngebretson:
@balmacefa This looks like really interesting functionality, I hadn't thought of using the scene camera like this, is this inspired by an existing 3D package?  

Before reviewing in earnest, would you mind having a look over the [C++ code conventions](https://github.com/AtomicGameEngine/AtomicGameEngine/wiki/Coding-Conventions-(CPlusPlus)) and make sure that the contributions follow the brief guidelines.  For example, one thing I notice is that it looks like the code has soft tabs (this is a common issue if you haven't run into it before).

Thanks.

balmacefa:
Sure I can change that and amend to the last commit,  I suggest you add a link of c++ convention to CONTRIBUTING. This functionality is present in blender and in deferent way in unity, they use another tab.

JoshEngebretson:
@balmacefa Thanks, I just posted the C++ conventions and that is a very good suggestion on the ```CONTRIBUTING``` doc. I'll move the information to the wiki and then link that in the CONTRIBUTING file, as it is easy to forget to update that one.  

Maybe someone will solve whitespace in our time? :)

balmacefa:
@JoshEngebretson The lines are indented, i make cleared implementation https://github.com/balmacefa/AtomicGameEngine/commit/022bfc961b541b3e47308889f951e686d6f959c4, but i do not know what you think, maybe insted of change the method to virtual (https://github.com/balmacefa/AtomicGameEngine/commit/022bfc961b541b3e47308889f951e686d6f959c4#diff-5b9c5e06c291aa111d6451d1fe3f2386R93), we can change the ptr to the defaultCamera in https://github.com/AtomicGameEngine/AtomicGameEngine/blob/c15b882d495fd6960478e958dd2c4301d5413027/Source/AtomicEditor/Editors/SceneEditor3D/SceneEditor3D.cpp#L256, this is to avoid save the user camera data(pos, rot)

JoshEngebretson:
@balmacefa It is looking better 👍 ,  we're currently getting the Build 3 binary release prepared (I still have one major issue in the works for it, finishing it up today).  I'd like to get the camera functionality factored in soon after the build ships.

BTW, good suggestion on updating [CONTRIBUTING.md](https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/CONTRIBUTING.md), I  added a link to developer info, including  C++ code convention page: https://github.com/AtomicGameEngine/AtomicGameEngine/wiki/Coding-Conventions-(CPlusPlus)

I noticed that GitHub smartly links the document when making new PR's, issues, etc, the developer info is a start and can be spruced up 🌲 

Thanks!

JimMarlowe:
This is an interesting feature, but it must have some UI feedback. Once the 'L' gets triggered, the camera is frozen. This can be done by pressing the 'L' and when you press 'C' to switch cameras the first time. With no locked indicator, you think the editor is broken.  It could really stand reporting which camera you are using also. Only supporting the 1st camera is hardcoded, is there no value in having more than one camera? If there is none, then this is fine. There is a oddity, when you have a camera selected and then you select the camera in the hierarchy view, you sometimes get a blue square, I've seen a big one and a smaller one, then it can disappear. 
bluebox-1329

PR #1334, opened 2017-01-09
Expose CycleDebugHudMode in UI

mattbenic:
This is just a quick add. I'm cycling these modes in our app, and it seemed like a generally useful one to push upstream.

### DIFF
--- a/Source/Atomic/UI/UI.h
+++ b/Source/Atomic/UI/UI.h
@@ -100,7 +100,7 @@ public:
     void ShowDebugHud(bool value);
     void ToggleDebugHud();
 
-    // Cycle debug HUD between showing primitive stats, current mode, profiler data, all three or none
+    /// Cycle debug HUD between showing primitive stats, current mode, profiler data, all three or none
     void CycleDebugHudMode();
 
     void SetDebugHudProfileMode(DebugHudProfileMode mode);
--- a/Source/Atomic/UI/UI.h
+++ b/Source/Atomic/UI/UI.h
@@ -100,6 +100,8 @@ public:
     void ShowDebugHud(bool value);
     void ToggleDebugHud();
 
+    void CycleDebugHudMode();
+
     void SetDebugHudProfileMode(DebugHudProfileMode mode);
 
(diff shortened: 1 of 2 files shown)


Source/Atomic/UI/UI.h
```diff
@@ -100,6 +100,8 @@ class UI : public Object, private tb::TBWidgetListener
     void ShowDebugHud(bool value);
     void ToggleDebugHud();
 
+    void CycleDebugHudMode();
```
JoshEngebretson: When adding new methods, I have been trying to remember to add a simple one or two line comment

```/// Cycle between showing rendering stats on the debug hud (none, rendering stats, stats and mode, stats and profiler) 

These are used for generated API reference documentation.  We don't have a coding standard document drafted yet.  Though, I want to get to this really soon.  I think requiring comments on public API is a good one.

Comment (mattbenic) [inline Source/Atomic/UI/UI.h]: Ah right, I've been adding comments or not based on surrounding code. I'll fix this one and keep it mind for future PRs :)

File: Source/Atomic/UI/UI.h [hunk 2]
```diff
@@ -100,6 +100,9 @@ class UI : public Object, private tb::TBWidgetListener
     void ShowDebugHud(bool value);
     void ToggleDebugHud();
 
+    // Cycle debug HUD between showing primitive stats, current mode, profiler data, all three or none
+    void CycleDebugHudMode();
+
```
mattbenic: Ah, we avoid /// because it confuses MS's auto commenting. But then that's for in-code comments, no method descriptions

mattbenic: There's actually a mishmash of commenting standards in that header alone. //, ///, and comments starting capitalized and non-capitalized

JoshEngebretson: Thanks :)  BTW, I filed #1335 to add the object metrics mode to the cycle 🐫

JoshEngebretson: Yeah, I think UI.h is one of the more guilty files for method commenting problems :)  

I have drafting a simple easy to follow coding standard (which borrows heavily from the legacy one and hits the major stuff) on my shortlist

JoshEngebretson:
Thanks for the PR 🐪  👍  Landing! 🛩

PR #1342, opened 2017-01-10
Added all keys and scancodes that are supported by SDL but not mapped

### LINKED ISSUE #1337: [Script] Add support for missing keys
There are some keys that are not supported in Atomic.  So far I know of `comma`, `period`, and `slash`.
They are all supported in SDL, but Urho (and thus Atomic) never explicitly support them.

I have the necessary changes in a branch, but am going to go through and find all of the other missing keys before submitting.

rosshadden:
You'll notice the `KEY_UNKNOWN` on line 269 is not in atomic begin/end blocks.  This is because [Urho actually has this mapped](https://github.com/urho3d/Urho3D/blob/master/Source/Urho3D/Input/InputEvents.h#L258) in the latest version, so I didn't see a reason to add it in the blocks.

I will at some point make a PR adding these to Urho itself, and if/when they merge we can replace ours from upstream.

Fixes #1337.


rosshadden:
The `atomicbuilds.moo.com` pages are not working, so I don't know why two of the checks are failing.

JoshEngebretson:
Thanks for the PR (and updating the submodule commit)

```atomicbuild.moo.com```  used to point at our LAN (where the build box currently is) and I have disabled that for security reasons.  We do plan on moving the build box to the "cloud" thing I keep hearing about...

... in the meantime, Windows doesn't like ```const``` here:

``` c:\jenkins\workspace\atomicpr_windows\source\atomic\input\../Input/InputEvents.h(423): error C2059: syntax error: 'constant'```

JoshEngebretson:
Awesome, thanks for the PR, LANDING! ✈️

PR #1347, opened 2017-01-11
emcc snprintf missing

JimMarlowe:
I fixed this only for ATOMIC_PLATFORM_WEB (emcc), the other platform's compilers were able to find these include files.


JoshEngebretson:
This PR is the :bomb:, thanks! LANDING! ✈️

PR #1349, opened 2017-01-11
Fixes CSComponents on reparented nodes not being re-added to the scene for updates

### LINKED ISSUE #290: Add EditorCache to root Cache folder
Cached editor representations will go in here, which will make it possible to have platform specific ones for deployment, and separate for atomic-cli as well, this also remove the need to have the root cache folder registered with the resource system

mattbenic:
CSComponents on reparented nodes not being re-added to the scene for updates

Fixes #290

### DIFF
--- a/Script/AtomicNET/AtomicNET/Scene/Scene.cs
+++ b/Script/AtomicNET/AtomicNET/Scene/Scene.cs
@@ -14,7 +14,16 @@ namespace AtomicEngine
             {
                 //Log.Info($"Node ADDED: {e.Node.Name}");
 
-                // TODO: How to handle adding components on existing node?
+                // The NodeRemoved event is generated when adding a node as a child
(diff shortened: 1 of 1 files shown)


JoshEngebretson:
This looks great, the comment was primarily as I was thinking this was a corner case (ie. removing the node from the scene and then deciding to add it for some reason, wasn't thinking of reparent case).  

I think we may need a specialized managed reparent method, which could save up some significant bookkeeping... maybe throw an exception when attempting to add a removed node with use "Reparent" method instead info in exception, we can wait for more data on adding this though :)

There are individual gotchas with various components being removed and added to a scene, and some that just shouldn't be (like Terrain), though in general this operation (especially when changing parents) should be well supported and quick

Thanks for the PR! LANDING ✈️

PR #1350, opened 2017-01-11
Adds windowed setting to player window in default prefs

mattbenic:
Our artists were looking for windowed functionality and when I showed them how to do it they suggested the key should be there by default, but disabled.

### DIFF
--- a/Script/AtomicEditor/editor/Editor.ts
+++ b/Script/AtomicEditor/editor/Editor.ts
@@ -77,7 +77,7 @@ class Editor extends Atomic.ScriptObject {
                 playerWindow.monitor = data.monitor;
                 playerWindow.maximized = true;
             } else {
-                playerWindow = {x: data.posX, y: data.posY, width: data.width, height: data.height, monitor: data.monitor, maximized: data.maximized};
+                playerWindow = {x: data.posX, y: data.posY, width: data.width, height: data.height, monitor: data.monitor, maximized: data.maximized, windowed: data.windowed};
             }
             Preferences.getInstance().savePlayerWindowData(playerWindow);
         });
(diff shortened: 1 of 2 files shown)


Script/AtomicEditor/editor/Editor.ts
```diff
@@ -77,7 +77,7 @@ class Editor extends Atomic.ScriptObject {
                 playerWindow.monitor = data.monitor;
                 playerWindow.maximized = true;
             } else {
-                playerWindow = {x: data.posX, y: data.posY, width: data.width, height: data.height, monitor: data.monitor, maximized: data.maximized};
+                playerWindow = {x: data.posX, y: data.posY, width: data.width, height: data.height, monitor: data.monitor, maximized: data.maximized, windowed: data.windowed};
```
JoshEngebretson: Does a "windowed" event parameter need to be set here? https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/AtomicApp/Player/IPCPlayerApp.cpp#L270

(BTW, I think "fullscreen" may be a little less confusing and is consistent with the term used in the Graphics API)

Script/AtomicEditor/editor/Preferences.ts
```diff
@@ -373,7 +375,8 @@ class PreferencesFormat {
             width: 0,
             height: 0,
             monitor: 0,
-            maximized: false
+            maximized: false,
+            windowed: false
```
JoshEngebretson: Does this default the player and editor to fullscreen? Where are these being used?

mattbenic:
This is an idiotic PR. My "testing" was clearly insufficient because the effect I was seeing was simply the reset of the config causing the window to move to the middle of the screen so the window frame is exposed. I tried "windowed" as an off chance (I thought I'd seen it previously) and that half-assed test indicated it worked. I am supremely embarrassed and annoyed with myself for this. Sorry :(

The real issue is that When a Height/Width is set, but no x/y, the window displays in such a way that the window title is hidden. But if no Height/Width/x/y is set, the window is centered.

JoshEngebretson:
No problem, I thought something might be up with this one, and it was really quick to look over.  Major context switches, many times a day on massive projects can be rough, am I right?  :)

mattbenic:
Sure.. I'll take that excuse >.>

Issue #940, opened 2016-08-04
[Linux] Pulse audio issues with CPU usage and black screen at startup


jeremyjh:
I followed the build instructions and built the latest master on Ubuntu 14.04 64-bit. After launching the editor I can see it continuously uses 110-120% CPU as reported in top. This is before opening a project, as well as with a project open. Games seem to run fine, its just an issue with the editor.

JimMarlowe:
If top shows SDLAudioDev2 as the process using most of the cpu, it can be changed by adding the environment variable "export SDL_AUDIODRIVER=pulseaudio". 
Note, this will make the game preview from the editor not have sound. 
The SDL 2.0.4 that came in Aug 8, is suppose to have fixed this SDL issue.

jeremyjh:
Top shows AtomicEditor using the CPU. Also, my earlier statement was incorrect - it is an issue playing games as well, the physics platformer burns two cores.

JoshEngebretson:
By default the AtomicPlayer runs with an uncapped framerate, there is a maxFps property on the Engine subsystem which should limit it: 

http://docs.atomicgameengine.com/api/classes/atomic.engine.html#maxfps

In main.js, Atomic.engine.maxFps = 60; // should limit CPU usage

The editor does this, not sure what is going on there:

https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Script/AtomicEditor/editor/Editor.ts#L46

jeremyjh:
Why does it use *two* cores though while playing a game, even if launched from `AtomicTool play`? Is it still drawing the editor  at 60 fps in a hidden window?

JoshEngebretson:
I am not sure profiling should help figure it out, on OSX pressing play from the editor, the editor and player processes are using expected amounts of CPU at 60 fps cap:

![screen shot 2016-08-10 at 9 00 15 pm](https://cloud.githubusercontent.com/assets/376203/17578446/7f6ebdac-5f3d-11e6-98fc-1dea34119250.png)

JoshEngebretson:
BTW, AtomicTool play just launches an editor instance in player mode, which doesn't have any editor stuff running.  The engine is threaded and does use work queues, also could be a SDL thread, where there have been issues with CPU usage and audio as @JimMarlowe mentioned.

EternalXY:
I've seen high CPU usage on Linux as well, in my case it was because of the ALSA sound thread.
[...]

PR #1384, opened 2017-01-16
Fix linux high cpu usage #940

JimMarlowe:
@EternalXY had it right (thanks), it is the ALSA SDL implementation that is coded poorly. SDL had half the solution in a comment! (now it is in their latest 2.0.6?) and the other half they are now aware of apparently. It took a while of trudging thru the SDL repo to see the intent behind these changes, and for the last change, SDL does not know, so it must be because @EternalXY is so smart. It is gated specifically for Linux, because SDL is a cross platform library, and other platforms have support for ALSA sound drivers, which may not incur these issues.


JoshEngebretson:
Fantastic! Wow, this has been a problem, well, forever!

This is sure to be popular with Linux users, especially ones greeted by a black screen at startup, a very :camel: thank you @JimMarlowe !!!

PS: The Windows CI box decided to reboot for updates, will kick it and the PR should go through, will land when green! 👍

PR #1385, opened 2017-01-17
JSONFile GetRoot fixes #1383

### LINKED ISSUE #1383: A minor typo causing assertion fail when saving a project file
In `Source/ToolCore/Project/ProjectFile.cpp`:
```C
void ProjectFile::Save(Project* project)
{
    ...
    SharedPtr<JSONFile> jsonFile(new JSONFile(context_));

    JSONValue root = jsonFile->GetRoot();

    root.Set("version", PROJECTFILE_VERSION);
    ...
}
```
Here `JSONValue root = jsonFile->GetRoot()` should be `JSONValue& root = jsonFile->GetRoot()` to obtain a reference of the JSON root. Otherwise `jsonFile` will not contain a valid JSON value and cause an assertion fail by the JSON writer.

I encountered such assertion fails in my application that integrates the atomic editor and calls this method to save projects. The official editor does not fail on this, it appears that it does not use this method to save projects.

JimMarlowe:
Fixes for JSONFile->GetRoot() returning a JSONValue&. I started out fixing the original #1383 issue, but though, if this mistake was made once, how many more times as it made? And here is your answer. It's too bad cppcheck couldn't find it.

### DIFF
--- a/Source/Atomic/Atomic2D/SpriteSheet2D.cpp
+++ b/Source/Atomic/Atomic2D/SpriteSheet2D.cpp
@@ -282,7 +282,7 @@ bool SpriteSheet2D::BeginLoadFromJSONFile(Deserializer& source)
 
     SetMemoryUse(source.GetSize());
 
-    JSONValue rootElem = loadJSONFile_->GetRoot();
+    JSONValue& rootElem = loadJSONFile_->GetRoot();
     if (rootElem.IsNull())
     {
         ATOMIC_LOGERROR("Invalid sprite sheet");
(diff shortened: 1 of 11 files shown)


JoshEngebretson:
Great catch, thanks for addressing it, hmmmm, I wonder if there is some way to enforce this, maybe by making the copy constructor private?  If this doesn't send a depth charge through the code, would be good to consider.

In the meantime, LANDING! ✈️ 👍  Closes #1383

JimMarlowe:
Doing a pass of the Source tree with cppcheck http://cppcheck.sourceforge.net/ could be helpful. Not too much pain for some gain, though sometimes it's just fixing things to be correct, and does not change use or performance.

Issue #1292, opened 2016-12-21
Remove UI psd assets


JimMarlowe:
There are some psd files among the turbobadger default skin UI images, probably intended to manufacture different widget stylings. They are taking up 9+meg of deploy space.

Type1J:
Can this be done through a .gitignore-like file?  I'm sure that TypeScript files don't need to be deployed either. Could files like .ts, .blender, .xsd, and .psd files (files that are usually converted or compiled into engine consumable files) be not deployed by a ignore.json file, which is also not deployed, but specifies which file to not deploy?

JimMarlowe:
The intent of these files is the question, the user does not have a lot of visibility as to the inclusion or lack of need for these files in their deployable. Some of the psd files are development prototyping, which are not needed at all, the rest of the psd files are (possibly) intended as templates to make new skins sets. We have used a different methodology to make new skins, and do not require these files. If anyone would need them, they are still in the Turbobadger repo.
Having finer control over what gets deployed in a application is a different issue, but a worthy one.

Issue #1267, opened 2016-12-08
[Editor] Add more contrast between button states


JoshEngebretson:
On some monitors it can be difficult to see which languages and platforms are selected on example browser and new project dialogs.  This also makes the case for having the light theme easily accessible.

PR #1386, opened 2017-01-17
Toggle button contrast #1267 and psd files #1292

JimMarlowe:
This pr makes a new skin element specifically for buttons in toggle mode. It adds a pressed image that is higher contrast than the button pressed image. Attempting to change the button pressed to be higher contrast was visible and distracting. This way the down state image can be adjusted separately.
I also removed the psd files in the turbobadger skin directories, as per reasoning in #1292


JoshEngebretson:
This looks great, maybe we should switch toggles in the inspector over to use the new better contrasting skin downstate?  Thanks for the cleanup too, landing 👍  Closes #1292 and #1267

PR #1387, opened 2017-01-17
Incorrect tab selection when resource editor is closed and then a new…

bparvanov:
… one is opened

Steps to reproduce:

1. open a resource in Atomic Editor
2. open another one (after this step we have two tabs)
3. close the first tab without selecting it (just press the X)
4. open another resource

Result: Both tabs are selected and we can see the contents of the tab
created in step 2 instead of the one created in step 4

This happens only when we close tabs that reside on the left hand side
of the currently selected tab.

Reason: TBTabContainer::m_current_page is not updated when tabs are
closed

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -49,6 +49,8 @@
 
 - Ross Hadden (https://github.com/rosshadden)
 
+- Borislav Parvanov (https://github.com/bparvanov)
+
 ### Contribution Copyright and Licensing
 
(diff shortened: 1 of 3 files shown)


JoshEngebretson:
@bparvanov Thanks for the PR, would you mind adding yourself to [AUTHORS.md](https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/AUTHORS.md) and I'll have a look?

bparvanov:
Done :)

JoshEngebretson:
Thanks, I'll test this out and get it in for the imminent next binary release

JoshEngebretson:
Awesome, thanks for the bug fix! 🐛 

LANDING: ✈️

PR #1389, opened 2017-01-18
Fixes #1305 - Convert constants to enums

### LINKED ISSUE #1305: [JavaScript/TypeScript] Add native module enum objects
As a follow up to #1304, the JS and TS API's should have identical enum value resolution.  This requires populating native module enum objects with values that mirror what is currently generated for TypeScript.

On the TypeScript side, ambient declarations for enum values should then be used which reference the JS API enum properties for consistency (including value) 

Important:  The existing JS API enum values should remain intact, though we do need to find a way to deprecate them.

shaddockh:
This PR adds new JS object literals to the Atomic namespace that allows for more proper enums instead of having to use constants.  Now, in both TS and JS, instead of using:
```
Atomic.BT_STATIC
```

you can use
```
Atomic.BodyType2D.BT_STATIC
```

The original constants are still present in the Atomic namespace so existing code doesn't break, but it is flagged to be removed.  I also left the existing constants in the ```d.ts```, but am contemplating removing it.  The repercussions would be that existing TS that uses the constants would have transpilation errors and would have to be converted to use the enums, but the JS would work fine.  Also, it would clean up the auto-generated documentations and not have all those extra constants.  Thoughts?

Each of the existing legacy constants has a comment stating that it is deprecated and points to the corresponding enum.

### DIFF
--- a/Source/ToolCore/JSBind/JSBTypeScript.cpp
+++ b/Source/ToolCore/JSBind/JSBTypeScript.cpp
@@ -367,7 +367,7 @@ void JSBTypeScript::ExportModuleEnums(JSBModule* module)
         while (valsIter != values.End())
         {
             String name = (*valsIter).first_;
-            source_ += "   /** Deprecated - use: [[" + _enum->GetName() + "." + name + "]] */\n";
+            source_ += "   /** Deprecated - use: [[" + _enum->GetName() + "." + name + "]]\n@deprecated */\n";
             source_ += "   export var " + name + ": number;\n";
             valsIter++;
         }
--- a/Source/ToolCore/JSBind/JSBTypeScript.cpp
+++ b/Source/ToolCore/JSBind/JSBTypeScript.cpp
@@ -340,7 +340,7 @@ void JSBTypeScript::ExportModuleEnums(JSBModule* module)
         JSBEnum* _enum = *enumIter;
 
         source_ += "   /** enum " + _enum->GetName() + "*/\n";
-        source_ += "   export const enum " + _enum->GetName() + " {\n";
+        source_ += "   export enum " + _enum->GetName() + " {\n";
 
         HashMap<String, String>& values = _enum->GetValues();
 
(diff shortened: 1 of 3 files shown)


shaddockh:
I searched for some kind of [Obsolete] annotation for TypeScript as that's really handy in C#, but unfortunately there is nothing really available yet.  There has been talk about how to handle it though.  In the mean time, I've just added comments on those elements that will get picked up by TypeDoc.

I like the idea of creating followup tasks to remove deprecated items.  That would be much better than scouring the code for ```TODO``` comments or just trying to make sure you remember that something has to be done.

mattbenic:
Yeah I think we all know how often those actually lead to things being done >.>

rosshadden:
There is some talk about handling deprecations and obsolutions in TypeScript [here](https://github.com/Microsoft/TypeScript/issues/390).  But not currently a thing.

JoshEngebretson:
@shaddockh Thanks for the PR, I'll check it out a bit later, looks good from a quick glance.

Yeah, C# compiler/IDE's have good support for deprecation... on TS, would be nice, though I think we can survive for a bit, for now good to have the comments in place for future TS deprecation language feature.

Source/ToolCore/JSBind/JSBTypeScript.cpp
```diff
@@ -349,28 +349,8 @@ void JSBTypeScript::ExportModuleEnums(JSBModule* module)
...
-            if (_enum->GetName() == "BodyType2D")
-            {
-                if (name == "BT_STATIC")
-                    value = "0";
-                else if (name == "BT_KINEMATIC")
-                    value = "1";
-                else if (name == "BT_DYNAMIC")
-                    value = "2";
-            }
-
```
JoshEngebretson: Nice cleanup :)  This is something that would be nice to be able to do in C#, though we can't due to constant requirements

Source/ToolCore/JSBind/JavaScript/JSModuleWriter.cpp
```diff
@@ -230,13 +230,28 @@ void JSModuleWriter::WriteModulePreInit(String& source)
...
+        // Preferred way - built-in enum
+        itr = values.Begin();
+        source.Append("duk_push_object(ctx);\n");
+        source.Append("duk_dup(ctx, -1);\n");
+        source.AppendWithFormat("duk_put_prop_string(ctx, -3, \"%s\");\n", jenum->GetName().CString());
+        while (itr != values.End())
+        {
+            String name = (*itr).first_;
+            source.AppendWithFormat("duk_push_number(ctx, (double) %s);\n", name.CString());
+            source.AppendWithFormat("duk_put_prop_string(ctx, -2, \"%s\");\n",name.CString());
```
JoshEngebretson: Hm, seems to be a bit odd here, we're using the name to format a double?

``` source.AppendWithFormat("duk_push_number(ctx, (double) %s);\n", name.CString());```

shaddockh: Yeah, I copied that from how it was constructing the constants above.  Those may both need to be fixed.  I'll look at that.

JoshEngebretson:
@shaddockh Hey, some neurons misfired on that string format, which uses the variable name, I have no idea what I was thinking :)  

I'll keep an eye out for any compiler warnings, though I don't think the cast to double was necessary, so in the end a few characters less code, and less code is always better 👍 

Thanks! LANDING for Build 3! ✈️

Issue #1415, opened 2017-01-24
[Windows] Change graphics backend to use Direct3D 11 by default


Nyt0x:
Going forward with the implementation of graphic features, it appears that supporting d3d9 as the default rendering backend is making things more tedious that they should be. 

According to http://store.steampowered.com/hwsurvey the userbase of d3d9 is getting really low nowdays. Sure there's still country / userbase other than steam users that still use d3d9 but we can keep the d3d9 backend for those who needs it, but as an option rather than the default settings.

This means that features integrated after that change would not require to support d3d9, so we can move forward more quickly.

That's why it would be nice to default to d3d11 as the rendering backend.

PR #1416, opened 2017-01-25
[Windows] Change graphics backend to use Direct3D 11 by default

Nyt0x:
[Windows] Change graphics backend to use Direct3D 11 by default  issue
#1415

I think that would be a first step to have d3d11 as the default on windows.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,10 @@ endif()
 
 add_link_libraries_exported (Atomic Box2D Duktape TurboBadger FreeType JO LZ4 PugiXml STB SDL)
 
+if (MSVC)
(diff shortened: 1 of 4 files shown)


JoshEngebretson:
This is great and should be a quick switch, once we're ready to do it.  We do need a bit more feedback on making the switch, so I made a thread: https://discourse.atomicgameengine.com/t/switching-default-windows-renderer-to-direct3d11/125

I am really stoked about the Water HQ improvement too! Let's keep that in a separate PR from the D3D switch, I'll close this, though let's keep it for reference for once we're ready to make the switch.

Thanks! 👍

Nyt0x:
Wait how did the water shader ended up in there? My god I'm bad with git..
No pb, I should not have send a request just for that anyway when I just did test desktop.

PR #1418, opened 2017-01-25
Close wrapped file when disposing a File.Stream

mattbenic:
Like it says on the label, fixes (my own damned) rookie mistake


JoshEngebretson:
Nice catch, thanks, LANDING! ✈️

Issue #1302, opened 2016-12-23
[TypeScript] Add a new native wrapper for ScriptEvents


shaddockh:
All of the native events now have auto-generated wrappers.  Now there needs to be a wrapper to generate custom ScriptEvents.  The wrapper should adhere to the following signature:

```typescript
function ScriptEvent<T extends Atomic.EventMetaData>(eventType: string, callback: Atomic.EventCallback<T>): Atomic.EventMetaData {
    return {
        _callback: callback,
        _eventType: eventType
    };
}
```

Then an example of how to use it needs to be created in the Atomic Examples.

Usage would be:

```typescript
// inside a common module
interface MyscriptEvent extends Atomic.ScriptEvent {
    x: number;
    y: number;
}

function MyscriptEvent(callback: Atomic.EventCallback<MyscriptEvent>): Atomic.EventMetaData {
  return Atomic.ScriptEvent<MyscriptEvent>("MyScriptEvent", callback);
}

// inside various components
window.subscribeToEvent(MyscriptEvent(e => {
    //do stuff
}));
```

shaddockh:
@JoshEngebretson when you get a chance can you look over TSH-ATOMIC-1302?  I think I'm missing something as my callback is not passing through for some reason.  Is there some trick to copy a JS function pointer to another object in duktape?

Test code is:

```
var ev = Atomic.ScriptEvent("foo", function() { console.log('inside callback'); });
console.log('event type: ' + ev._eventType);
console.log('before callback');
ev._callback(); // this executes, but doesn't call console.log like it should
console.log('after callback');
```

JoshEngebretson:
Sorry, just saw this, noted to look soon :)

JoshEngebretson:
I still owe some looking, did file a C# issue for scripted events and think TS/C# designs should be as we can muster: #1381

I am hoping we might be able to reduce the clutter of, somehow, does this need to be SO strictly typed, sorry if this seems like an out there question, I blame context switching :)

```TypeScript
function MyscriptEvent(callback: Atomic.EventCallback<MyscriptEvent>): Atomic.EventMetaData {
  return Atomic.ScriptEvent<MyscriptEvent>("MyScriptEvent", callback);
}
```

shaddockh:
I think it can be a little bit more cleaned up and allow the TS inference to determine the types:
```typescript
function MyscriptEvent(callback: Atomic.EventCallback<MyscriptEvent>) {
[...]

PR #1420, opened 2017-01-26
Fix ScriptEvent for 1302

JoshEngebretson:
This is a fix for the in progress #1302 branch

The parameter count for the native method was incorrect, so the method was never getting called.  

It is possible to debug this with the AtomicEditor by passing the --player option, specifying a project, and placing a break point in the method.  

AtomicEditor --player --project /Users/josh/Documents/AtomicProjects/MyProject

If you want variable arguments to a native function btw, specify DUK_VARARGS (then you need to check the stack top for how many args are passed in).  

I also removed the push of the current function at the top of the method.

The code paste now runs! 👍  😄


shaddockh:
Thanks!  :+1:   It didn't even occur to me that I was passing the incorrect number of params...thanks for the lookover.

PR #1422, opened 2017-01-27
Atomic Script Events

### LINKED ISSUE #1302: [TypeScript] Add a new native wrapper for ScriptEvents

shaddockh:
This PR resolved #1302 allowing for a more standard way of defining script events to align with the Native events.  Thanks for the assistance @JoshEngebretson!

### DIFF
--- a/Script/TypeScript/AtomicWork.d.ts
+++ b/Script/TypeScript/AtomicWork.d.ts
@@ -27,22 +27,19 @@ declare module Atomic {
     // end subsystems
 
     // Base interface for events, contains eventType string and callback
-    interface EventMetaData
-    {
+    interface EventMetaData {
         _eventType: string;
         _callback: any;
     }
 
-    interface NativeEvent extends EventMetaData
-    {
-    }
+    interface NativeEvent extends EventMetaData { }
 
-    interface ScriptEvent extends EventMetaData
-    {
-    }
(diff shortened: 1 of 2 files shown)


JoshEngebretson:
Holy 🐫 !  When it is right, it is right, I believe we've nailed it :)  +28/-6 lines to add ScriptEvent to JS/TS, sure why not!  Hot dammmnnnn! 🔥 

LANDING! ✈️

Issue #1425, opened 2017-01-28
[TS Extension] Set "allowJs" to false when creating the external tsconfig.json


shaddockh:
The setting inside the tsconfig.json for trying to recompile external javascript files should be set to false.  This is an advanced feature and not necessary in most TypeScript projects.

Need to set:
```
"allowJs": false
```
in the generated tsconfig.

PR #1426, opened 2017-01-28
Fix for #1425

shaddockh:
@JoshEngebretson, This fixes the issue with the Atom editor displaying the "js outdated" error described on the forums.


JoshEngebretson:
Great, thanks 👍

PR #1430, opened 2017-02-02
Created a case for UInt types when setting values on CSComponents

raheelx:
Hi @JoshEngebretson 

i added a case for uints when setting values to fields on a CSComponent. 

- Raheel


JoshEngebretson:
Hi @raheelx, awesome, thanks!  LANDING ✈️

PR #1433, opened 2017-02-04
Toggle editor themes

JimMarlowe:
Allows switching of light and dark editor themes, using an entry in the Developer menu, Toggle Theme.
This change happens in real time, and is saved in the preferences to the last setting you've selected, so it will come back in that theme.
It handles 99% of the user interface, there are some minor artifacts when switching between the themes, a couple of borders and hilights are not changed for Turbobadger reasons and these return to normal on the next editor invocation. Since swapping themes is probably going to be a infrequent operation, it is tolerable.
The codeEditor theme is not changed, and needs some work in order to also change it on the fly.


shaddockh:
@JimMarlowe I started looking at what would be required to send the change theme message to the code editor and I have it all working except I ran into some kind of race condition where the webclient pulls a stale version of the settings.  If you submit the value a second time it picks it up though making me think that the settings are getting propagated *after* the menu event completes.  Still looking into it and if I find a solution I'll PR.

shaddockh:
Ok, I've got the plumbing in place in my branch at TSH-ATOMIC-DYNAMIC-WEBVIEW-SETTINGS

Basically, if you pull this in and set the app setting: "CodeEditor","Theme","vs or vs-dark", then it will get sent to any active web views and they will automatically update.

I'll go ahead and PR this so you can pull it in.

JoshEngebretson:
@JimMarlowe Awesome!  This will certainly help get the light theme in circulation :)  I've landed @shaddockh's related PR's including the event signature TypeScript update, would you mind checking the JM-LIVE-THEME branch (might be good to rebase to master) and seeing if anything needs tweaking?  

I am excited to see the light theme switch, also switching out monaco theme, so cool guys, thanks! 🐫 🍰 (Hm, camel cake, maybe camels don't go with *everything* 😉 )

JimMarlowe:
I'm workin it, had to double dip the rebase, and I'll add in the editor theme change too.

JimMarlowe:
ack! this appears to have gone horribly wrong. I think its time to close this and start over.

JoshEngebretson:
No worries, git can really go sideways sometimes :)  I'd like to start squashing PR commits down when possible, definitely helps when rebasing to be working with one changeset

PR #1435, opened 2017-02-04
User and Application settings improvements

shaddockh:
This PR enhances the way settings are handled
- Editor level settings can be retrieved and set from editor extensions
- When settings are updated, any open web views will auto-reconfigure the code editor window with any updates.  This allows for changing the code theme on the fly

### DIFF
--- a/Script/AtomicEditor/editor/Editor.ts
+++ b/Script/AtomicEditor/editor/Editor.ts
@@ -153,7 +153,7 @@ class Editor extends Atomic.ScriptObject {
 
 
     /**
-     * Sets a user preference value in the user settings file
+     * Sets a user preference value in the project user settings file
      * @param  {string} extensionName name of the extension the preference lives under
      * @param  {string} preferenceName name of the preference to set
      * @param  {number | boolean | string} value value to set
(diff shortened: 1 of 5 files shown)


Script/AtomicEditor/editor/Editor.ts
```diff
@@ -170,6 +170,37 @@ class Editor extends Atomic.ScriptObject {
     }
 
     /**
+     * Return a preference value or the provided default from the global user settings file
+     * @param  {string} groupName name of the section the preference lives under
+     * @param  {string} preferenceName name of the preference to retrieve
+     * @param  {number | boolean | string} defaultValue value to return if pref doesn't exist
+     * @return {number|boolean|string}
+     */
```
JoshEngebretson: What is your secret to these doc strings?  It is everything I can muster sometimes to even do a /// one for C++ methods!

Script/AtomicEditor/editor/Editor.ts
```diff
@@ -170,6 +170,37 @@ class Editor extends Atomic.ScriptObject {
...
+     * @param  {string} preferenceName name of the preference to set
+     * @param  {number | boolean | string} value value to set
+     */
+    setApplicationPreference(groupName: string, preferenceName: string, value: number | boolean | string) {
+        Preferences.getInstance().setApplicationPreference(groupName, preferenceName, value);
+        WebView.WebBrowserHost.setGlobalStringProperty("HOST_Preferences", "ApplicationPreferences", JSON.stringify(Preferences.getInstance().cachedApplicationPreferences, null, 2 ));
+        const eventData: EditorEvents.UserPreferencesChangedEvent = {
+            projectPreferences: JSON.stringify(Preferences.getInstance().cachedProjectPreferences),
+            applicationPreferences: JSON.stringify(Preferences.getInstance().cachedApplicationPreferences)
+        };
```
JoshEngebretson: neat stuff on the web host :)

Script/AtomicEditor/editor/Preferences.ts
```diff
@@ -266,9 +265,63 @@ class Preferences {
...
+     * Return a preference value or the provided default from the global user settings file located in the project
+     * @param  {string} settingsGroup name of the group these settings should fall under
+     * @param  {string} preferenceName name of the preference to retrieve
+     * @param  {number | boolean | string} defaultValue value to return if pref doesn't exist
+     * @return {number|boolean|string}
+     */
+    getApplicationPreference(settingsGroup: string, preferenceName: string, defaultValue?: number): number;
+    getApplicationPreference(settingsGroup: string, preferenceName: string, defaultValue?: string): string;
+    getApplicationPreference(settingsGroup: string, preferenceName: string, defaultValue?: boolean): boolean;
+    getApplicationPreference(settingsGroup: string, preferenceName: string, defaultValue?: any): any {
```
JoshEngebretson: Interesting, TS lets you have methods that differ only in return value, I mean since the default value is all optional?  I might be missing something here, though if not, that seems like a TS compiler hole to me, or it could just be my C/C++/C# brain cells being tricked by ```tsc```

shaddockh: These signatures are actually different based upon the ```defaultValue```.  The final signature is actually never seen from the TS side but it's only seen from the JS side which is how overloads work in TS.  Under the covers that return value really could be anything, but the signatures are there to assist with the TS typechecking so it knows that if you pass a boolean default value then the result should be boolean.  I just checked and saw that if you don't pass a default value, it takes the topmost signature and assumes that the result is a number.

JoshEngebretson:
Hey, so master has an issue with the combination of changes from the PR's

Get this after cleaning with ```git clean -dxf```

AtomicEditor/editor/Editor.ts(194,39): error TS2305: Module '"c:/Dev/atomic/AtomicGameEngine/Script/AtomicEditor/editor/EditorEvents"' has no exported member 'UserPreferencesChangedEvent'.
AtomicEditor/editor/Editor.ts(199,37): error TS2339: Property 'UserPreferencesChangedNotification' does not exist on type 'typeof "c:/Dev/atomic/AtomicGameEngine/Script/AtomicEditor/editor/EditorEvents"'.

Any thoughts?

shaddockh:
Yep, just noticed that... fix is incoming.

PR #1438, opened 2017-02-06
Fix event handler to use the native wrapper.  

shaddockh:
The merge pulled out the EditorEvents file that this relied on.


JoshEngebretson:
Easy peasy, will land when green, thanks!

PR #1444, opened 2017-02-10
Add Spin control to Animation Toolbar

JimMarlowe:
This PR adds a control in the Animation Toolbar to spin the previewed model. Though you can use the Inspector to rotate the model, who's got time for that? The control is a Slider, the leftmost position is off, there are 10 steps, the further right the slider is pushed, the faster the spin. Bonus, This feature also works while animations are playing.


JoshEngebretson:
Works great and really handy for viewing model in 360 at measured steps, thanks! LANDING! ✈️

PR #1446, opened 2017-02-12
Add statusbar to AtomicEditor

JimMarlowe:
This PR adds a status bar to the AtomicEditor. It displays text at the bottom of the tool, and uses a timer to remove the text after 10 seconds to avoid having stale text being displayed. The feature can be accessed from inside the editor code with the API  EditorUI.showEditorStatus( "I am a status message" );
or by using an event like : this.sendEvent(Editor.EditorModalEventData({ type: Editor.EDITOR_MODALINFO, title: "", message: "I am a status message" }));


shaddockh:
Very cool!  I didn't even cross my mind that the editor would have access to the Atomic.UpdateEvent.  That opens up some doors to hook in some kind of promises system and handle some situations where one area of the editor is waiting on the results of something in the future.  Nice!

JoshEngebretson:
The status bar is really nice, love it, and another great add in +35 lines, thanks! ❤️ 

screen shot 2017-02-13 at 9 12 19 am

PR #1448, opened 2017-02-14
Add thumbnail tooltip

JimMarlowe:
This PR adds a thumbnail (image) tooltip. It will automagically provide the thumbnail tooltip for .png, .jpg assets in a project's Texture directory.  There is also support for using thumbnails for .prefab, .mdl and .fbx assets.   To add a thumbnail for an asset is "easy". The user can create a new directory in the project root directory (where Resources and Cache are) called Support.   A (png) thumbnail for an asset is named assetname.ext.png, where assetname.ext is what you would see in the Project browser. Then, when you hover over assetname.ext, the thumbnail tooltip will appear. If no thumbnail is created for the asset, then nothing will appear.
Normal text tooltips are unaffected, for a thumbail tooltip, the 1st character in the string is a  "#" character, followed by a pathname to the image to display. The Texture based images in the project will show the actual image files, as defined by the asset. The other type assets are redirected to the Support directory to look for custom thumbnail images.
The reason the Project/Support directory is selected to hold the thumbnails, is so they don’t effect the asset system, they are not needed for deploying the game, but they are relevant to the project.

OH NO 35 lines again!

### DIFF
--- a/Script/AtomicEditor/ui/frames/ProjectFrame.ts
+++ b/Script/AtomicEditor/ui/frames/ProjectFrame.ts
@@ -575,6 +575,21 @@ class ProjectFrame extends ScriptWidget {
         button.skinBg = "TBButton.flat";
         button["asset"] = asset;
         blayout["assetButton"] = button;
+
+        // support for image tooltips
+        var extx = asset.getExtension();
+        if ( extx == ".jpg" || extx == ".png" ) {
+            var tt = "#" + asset.path;
+            button.tooltip = tt;
+        }
+        if ( extx == ".prefab" || extx == ".mdl" || extx == ".fbx" ) {
+            var thumbfolder = ToolCore.toolSystem.project.projectPath + "Support/";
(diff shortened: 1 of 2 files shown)


JoshEngebretson:
Cool, I wonder if it would make sense to put the custom preview images next to the .asset files in the Resources folder instead of in a new Support folder?  So like, ```Camel.png``` would have a ```Camel.preview.png``` or  ```Camel.preview.jpg```, these can be filtered at build time, and will be easy to add to repo as you would want to commit any custom previews I would think?

We need a better asset content UI, which supports previews, at that point probably won't need the preview in the tooptip.  We'll also want automatic generation of texture and fbx/prefab assets (can render the later to texture), custom preview items would override any automated preview. 

Can cross that creek when we get to it... 🛶 😄

JimMarlowe:
Down the spillway it goes.

JoshEngebretson:
Ok, though let's make sure not to break the 35 line streak, 🐫 💨 

😃

PR #1450, opened 2017-02-17
Add screenshot function

JimMarlowe:
This PR add the function to save screenshots of the editor.
There is a preference for the screenshot path.
There is a preference for the save format : png, jpg, bmp, tga, dds
The image is saved with an autogenerated name : Screenshot_dd_mm_yyy_HH_MM_SS.ext
The screenshot is invoked with either the menu entry in Developer->Screenshot, or the key combo, Ctrl+9

Sorry, but this breaks the 35 additions chain... so talk to these ladies...
screenshot_16_02_2017_19_10_19

### DIFF
--- a/Script/AtomicEditor/ui/Shortcuts.ts
+++ b/Script/AtomicEditor/ui/Shortcuts.ts
@@ -174,6 +174,29 @@ class Shortcuts extends Atomic.ScriptObject {
         }
     }
 
+    invokeScreenshot() {
+        var features = Preferences.getInstance().editorFeatures; // get prefs
+        var pic_ext = features.screenshotFormat;
+        var pic_path = features.screenshotPath;
+        var dx = new Date();  // get the date NOW
+        var datestring = ("0" + dx.getDate()).slice( -2 ) + "_" + ("0" + (dx.getMonth() + 1 )).slice( -2 ) + "_" + dx.getFullYear()
+            + "_" + ("0" + dx.getHours()).slice(-2) + "_" + ("0" + dx.getMinutes()).slice(-2) + "_" + ("0" + dx.getSeconds()).slice(-2);
+        pic_path += "/Screenshot_" + datestring + "." + pic_ext;  // form filename
+        var myimage = new Atomic.Image; // make an image to save
+        if (Atomic.graphics.takeScreenShot(myimage)) { // take the screenshot
+            var saved_pic = false;
+            var jpgquality = 92; // very good quality jpeg 
+            if ( pic_ext == "png" ) saved_pic = myimage.savePNG(pic_path);
+            else if ( pic_ext == "jpg" ) saved_pic = myimage.saveJPG(pic_path, jpgquality);
+            else if ( pic_ext == "tga" ) saved_pic = myimage.saveTGA(pic_path);
+            else if ( pic_ext == "bmp" ) saved_pic = myimage.saveBMP(pic_path);
+            else if ( pic_ext == "dds" ) saved_pic = myimage.saveDDS(pic_path);
+            if (saved_pic)  EditorUI.showEditorStatus ( "Saved screenshot " + pic_path );
+            else EditorUI.showEditorStatus ( "Error - could not save screenshot " + pic_path );
+        }
+        else EditorUI.showEditorStatus ( "Error - could not take screenshot.");
+    }
+
     handleKeyDown(ev: Atomic.KeyDownEvent) {
 
         // if the right mouse buttons isn't down
@@ -249,6 +272,9 @@ class Shortcuts extends Atomic.ScriptObject {
                 } else {
                     this.invokePauseOrResumePlayer();
                 }
+            }
+            else if (ev.key == Atomic.KEY_9) {
+                this.invokeScreenshot();
             }
 
         }
--- a/Script/AtomicEditor/ui/Shortcuts.ts
+++ b/Script/AtomicEditor/ui/Shortcuts.ts
@@ -179,7 +179,7 @@ class Shortcuts extends Atomic.ScriptObject {
         var pic_ext = features.screenshotFormat;
         var pic_path = features.screenshotPath;
         var dx = new Date();  // get the date NOW
-        var datestring = ("0" + dx.getDate()).slice( -2 ) + "_" + ("0" + (dx.getMonth() + 1 )).slice( -2 ) + "_" + dx.getFullYear()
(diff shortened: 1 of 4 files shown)


JimMarlowe:
Changed filename format from Screenshot_dd_mm_yyyy_HH_MM_SS.ext to Screenshot_yyyy_mm_dd_HH_MM_SS.ext so they sort better, what was I thinking?

JoshEngebretson:
Thanks for the PR, will look soon, love the pitchfork btw! :)

Script/AtomicEditor/ui/Shortcuts.ts
```diff
@@ -174,6 +174,29 @@ class Shortcuts extends Atomic.ScriptObject {
...
+        var myimage = new Atomic.Image; // make an image to save
+        if (Atomic.graphics.takeScreenShot(myimage)) { // take the screenshot
+            var saved_pic = false;
+            var jpgquality = 92; // very good quality jpeg 
+            if ( pic_ext == "png" ) saved_pic = myimage.savePNG(pic_path);
+            else if ( pic_ext == "jpg" ) saved_pic = myimage.saveJPG(pic_path, jpgquality);
+            else if ( pic_ext == "tga" ) saved_pic = myimage.saveTGA(pic_path);
+            else if ( pic_ext == "bmp" ) saved_pic = myimage.saveBMP(pic_path);
+            else if ( pic_ext == "dds" ) saved_pic = myimage.saveDDS(pic_path);
+            if (saved_pic)  EditorUI.showEditorStatus ( "Saved screenshot " + pic_path );
```
JoshEngebretson: ```showEditorStatus``` FTW! 👍

Script/AtomicEditor/ui/frames/menus/MainFrameMenu.ts
```diff
@@ -40,7 +40,8 @@ class MainFrameMenu extends Atomic.ScriptObject {
         MenuItemSources.createMenuItemSource("menu tools", toolsItems);
         MenuItemSources.createMenuItemSource("menu developer", developerItems);
         MenuItemSources.createMenuItemSource("menu help", helpItems);
-
+        this.goScreenshot = 0;
+        this.subscribeToEvent(Atomic.UpdateEvent((ev) => this.handleScreenshot(ev)));
```
JoshEngebretson: Adding additional event handlers which are called per frame (especially scripted ones) can add up, I wonder if we can bring in the EventLoop stuff @shaddockh did on this, ```setTimeout``` seems like what we want here, which would also remove the need to the bookkeeping ```goScreenshot``` member, a couple examples:

Delayed scene load: https://github.com/AtomicGameEngine/AtomicExamples/blob/1f1bc5c8aa5676e3b084b874a7a5718fe0e3808f/SFXR/JavaScript/Resources/Scripts/main.js#L3

The events example itself: https://github.com/AtomicGameEngine/AtomicExamples/blob/master/EventLoop/JavaScript/Resources/Scripts/main.js

I looked and didn't see an example of the AtomicEventLoop module being brought into the editor's TS, I think the event module needs to be exposed, @shaddockh what do you think of adding the functions to duktape.d.ts, as they are kind of "built-ins"? https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Script/TypeScript/duktape.d.ts

shaddockh: @JoshEngebretson that's definitely doable, but it's only built in if you add the ```require("AtomicEventLoop")``` to expose it.  Is there a way to auto-load that from the native side so that it's always available and doesn't have to be required in?  If that's doable then adding the pieces to ```duktape.d.ts``` can be done and that functionality used w/o needing any additional setup.

JimMarlowe: I can see your point about the Update handlers, though it could take a while to see a performance impact in the Editor. Another option that is available is the TBmessage, which has an ability to post a message after a delay.  This would work for some of the recent Editor features (screenshot, statusbar), but would be Editor + UI specific, which isn't all bad.

JimMarlowe: OK, using the TBmessage subsystem is a **bad** idea. However, after looking at the AtomicEventLoop, this too seems like overkill, and all I may need is `setTimeout ( expression, timeout )` to do the needed job, I'll test it in a little while to see if it works.

JoshEngebretson: @shaddockh I think we can require the event loop methods for the editor in main.ts can't we, then they are accessible in the VM at global scope: https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Script/AtomicEditor/main.ts

The event loop is in CoreData, so available to all Atomic apps, editor and players included :) 

https://github.com/AtomicGameEngine/AtomicGameEngine/tree/master/Resources/CoreData/AtomicModules

@JimMarlowe setTimeout is part of the standard JS API provided by requiring AtomicEventLoop

shaddockh: @JoshEngebretson That would definitely work for the editor.  For the player though, individuals would still need to add the ```require("AtomicEventLoop")``` to their project's ```main.js``` themselves.  One possibility though is to update the new project template to automatically have this in the ```main.js``` that gets generated.  What do you all think of going down that route?

JimMarlowe: I think I'm caught up... Duktape has no native setTimeout function, like you see all over the lying internet. Luckily, svaarala, Jay & TSH had foreseen this deficiency, and provided native functions with the AtomicEventLoop module.

shaddockh: yep.  TSH == me :)  That was my first contribution to Atomic.

JimMarlowe:
I submitted an Issue to get the AtomicEventLoop to AtomicEditor to get access to setTimeout(), in the mean time, I made the eventhandler only get registered when the menu is used to take a screenshot, then unregisters it, until it in needed again.

JoshEngebretson:
@JimMarlowe Good solution on the subscribe/unsubscribe pattern, I bet we could generalize this for single shot events which auto-unsubscribe, delayed events and single shot would be two good adds for event system I am thinking, and would also work in C#/C++

I'll go ahead and land this, if you want to adjust it once setTimeout is available that is fine too, thanks!

PR #1459, opened 2017-03-03
[WINDOWS OPENGL] Fix for WebView on OpenGL on Windows

darrylryan:
Fix for WebView on OpenGL on Windows (also fixes Monaco text editor on Windows WebGL which uses WebView).


Source/AtomicWebView/WebTexture2D.cpp
```diff
@@ -270,67 +270,65 @@ class WebTexture2DPrivate : public CefRenderHandler
...
+			}
+			if (y + h > viewheight)
+			{
+				h -= y + h - viewheight;
+			}
+
+			unsigned char* src = (unsigned char*)buffer;
+			D3DBlit(IntRect(x, y, x + w, y + h), src, width * 4, false);
+
+		}
```
JoshEngebretson: Did tabs get introduced here?

Source/AtomicWebView/WebTexture2D.cpp
```diff
@@ -392,6 +390,14 @@ void WebTexture2D::SetSize(int width, int height)
 
 #endif
 
+#if defined(ATOMIC_OPENGL) && defined(ATOMIC_PLATFORM_WINDOWS) 
+	//Force width and height to multiples of 2 so that we don't squish pixels
+	//This doesn't seem to make the text completely sharp but it makes it usable for now.
+	width = (width | 2) + 1;
+	height = (height | 2) + 1;
+	texture_->SetFilterMode(TextureFilterMode::FILTER_NEAREST);
+#endif
```
JoshEngebretson: Hrm, this could be the UV adjustment in TB, what does it look like using the 0.5f half-pixel adjustment, without the width/height adjustment and not setting the filter mode? 

https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/ThirdParty/TurboBadger/renderers/tb_renderer_batcher.cpp#L241

Source/AtomicWebView/WebTexture2D.cpp
```diff
@@ -392,6 +390,14 @@ void WebTexture2D::SetSize(int width, int height)
 
 #endif
 
+#if defined(ATOMIC_OPENGL) && defined(ATOMIC_PLATFORM_WINDOWS) 
+    //Force width and height to multiples of 2 so that we don't squish pixels
+    //This doesn't seem to make the text completely sharp but it makes it usable for now.
+    width = (width | 2) + 1;
+    height = (height | 2) + 1;
+    texture_->SetFilterMode(TextureFilterMode::FILTER_NEAREST);
+#endif
```
JoshEngebretson: Ok, this needs further looking into as having a pixel discrepancy between requested size and what the size of the texture actually is, only on Windows, and only when under GL, could cause some hard to track down memory buffer access violations.  It also breaks the early out at the top of the method for size == textureSize when requesting odd sized width/height, so will always will recreate

darrylryan: I don't have much clue what is causing this (or why), hence the hack-around.
I might come back to it later but it works OK for my own use at the moment so its not top of my fix list any more... I probably won't go further on this for a couple of months if someone else wants to take a look sooner.

JoshEngebretson:
This turned out to be the D3D9 pixel adjustment I mentioned in the code review being applied in OpenGL.  

Adding 1 to the width/height of the WebView texture compensated for the pixel adjustment, though made the text blurry in the process, also not fixing the underlying issue affecting both TurboBadger controls and the WebView see the "Spinner.js" text in screenshot.  

I rebased to master and applied the pixel adjustment fix, will land from here:  https://github.com/AtomicGameEngine/AtomicGameEngine/pull/1517

Please let me know if it is indeed fixed on your setup, thanks! 

crunchy

PR #1467, opened 2017-03-11
Small improvements in Thread class.

ombre5733:
- Make the helper functions to run the threaded function in the new thread static. This avoids
  exposing the symbol for the helper functions.
- Thread::GetCurrentThreadID() should not use pthread if ATOMIC_THREADING is undefined.


JoshEngebretson:
Great, thanks for the PR, landing! ✈️

PR #1468, opened 2017-03-13
Add error scraper to Player Output window

JimMarlowe:
This PR adds a Javascript error scraper to the Player Output window, which puts the magnifying glass next to an error filename + line number, and when clicked, brings the file up in the editor at the problem line.  The implementation is a little different than the TS version, the JS errors can repeat the same file+line, so only unique errors are captured. The same reference link is used by the magnifying glass widget for shared errors.


JoshEngebretson:
Very cool, thanks! LANDING! ✈️

PR #1474, opened 2017-03-15
Make PackageFile.Scan case-insensitive on Windows

mattbenic:


Source/Atomic/IO/PackageFile.cpp
```diff
@@ -166,12 +166,18 @@ void PackageFile::Scan(Vector<String>& result, const String& pathName, const Str
     if (filterExtension.Contains('*'))
         filterExtension.Clear();
 
+    bool caseSensitive = true;
+#ifdef _WIN32
+    // On Windows ignore case in string comparisons
```
JoshEngebretson: In general like to stick with ATOMIC_PLATFORM_WINDOWS for windows specific code, just noting, don't need adjustment for this PR

Source/Atomic/IO/PackageFile.cpp
```diff
@@ -166,12 +166,18 @@ void PackageFile::Scan(Vector<String>& result, const String& pathName, const Str
...
+
     const StringVector& entryNames = GetEntryNames();
     for (StringVector::ConstIterator i = entryNames.Begin(); i != entryNames.End(); ++i)
     {
         String entryName = GetSanitizedPath(*i);
-        if ((filterExtension.Empty() || entryName.EndsWith(filterExtension)) &&
-            entryName.StartsWith(sanitizedPath))
+        if ((filterExtension.Empty() || entryName.EndsWith(filterExtension, caseSensitive)) &&
+            entryName.StartsWith(sanitizedPath, caseSensitive))
         {
```
JoshEngebretson: This is used when generating file packages for other platforms, from Windows, does this affect those platforms?

mattbenic: Hmm, yes I think it might.
On the comment about the define, I stuck to the one used elsewhere in the same file.

JoshEngebretson: Ok, maybe need a bool parameter on the case insensitive, which Windows builds pass in?  It would be nice if .pak files were universal though, hrm.  Gotcha on the ```_WIN32``` would be good to go through and fix the inconsistencies there, no worries.

JoshEngebretson:
I had a closer look and this appears to only be at runtime, not during build, so should be a-ok! 👍 

Thanks, LANDING!

Issue #710, opened 2016-03-18
[LINUX] Editor crash when creating a new 3D project, or when trying to open some example projects


guzzard:
**Creating a new _3D_ project:**
Info,  T0: Leaving post processing pipeline
[Fri Mar 18 11:21:50 2016] DEBUG: Created directory /home/username/AtomicProjects/MyGame/Resources/Models/Materials
Skipping save of existing material DefaultMaterial
[Fri Mar 18 11:21:50 2016] DEBUG: Loading resource .mdl
[Fri Mar 18 11:21:50 2016] ERROR: Could not open file /home/username/AtomicProjects/MyGame/Cache/
Could not open output file /home/username/AtomicProjects/MyGame/Cache/

**When opening some example projects (for example Basic3D or ToonTown), same error:**
Info,  T0: Leaving post processing pipeline
[Fri Mar 18 11:27:04 2016] DEBUG: Created directory /home/username/temp/AtomicExamples/Basic3D/Resources/Models/Materials
Skipping save of existing material DefaultMaterial
[Fri Mar 18 11:27:04 2016] DEBUG: Loading resource .mdl
[Fri Mar 18 11:27:04 2016] DEBUG: Loading resource Models/Materials/DefaultMaterial.material
[Fri Mar 18 11:27:04 2016] DEBUG: Loading resource Techniques/Diff.xml
[Fri Mar 18 11:27:04 2016] DEBUG: Loading resource Textures/chest.png
[Fri Mar 18 11:27:04 2016] ERROR: Could not open file /home/username/temp/AtomicExamples/Basic3D/Cache/
Could not open output file /home/username/temp/AtomicExamples/Basic3D/Cache/

Opening Basic2D and BunnyMark example projects worked fine, but they have another issue described in #713.

I'm running Arch Linux and got no errors during the build (built with ./Build_AtomicEditor.sh, build with QtCreator fails). I copied the files mentioned in #708 and the editor started fine after that, however it's not usable due to the issues described above.

AlexandrDudnyk:
I have same problem too. Linux Mint 17.3 64 bit
`[Fri Mar 18 17:23:56 2016] DEBUG: Created directory /home/alexandr/AtomicProjects/MyGame3/Resources/Models/Materials
Skipping save of existing material DefaultMaterial

[Fri Mar 18 17:23:56 2016] DEBUG: Loading resource .mdl
[...]

Issue #1469, opened 2017-03-13
XML-loading does not work in linux with German locale


ombre5733:
If a German locale is set in linux, loading from an XML file does not work. The problem is that in the XML files, a dot is used as decimal separator but the code for converting string to double (or string to float) in
`Source/AtomicGameEngine/Source/Atomic/Core/StringUtils.cpp` expects the localized separator (a comma for German). The ultimate problem is that `strtod` is used in `StringUtils.cpp`.

A mitigation could be to the same approach that rapidjson uses.

ombre5733:
Julia had the same issue:

[https://github.com/JuliaLang/julia/issues/5928](https://github.com/JuliaLang/julia/issues/5928)
[https://github.com/JuliaLang/julia/pull/5988](https://github.com/JuliaLang/julia/pull/5988)

ombre5733:
Close via #1479.

Issue #1465, opened 2017-03-08
Black screen when running 2D platformer demo in linux


ombre5733:
I have built the AGE from source on an Ubuntu 16.10 system. Running the 2D platformer demo results in an empty (black) window. Since there is a sound when jumping, I guess that something happens with the display output.

JimMarlowe:
In the AtomicEditor main menu, got to Tools->Logs->Player Log, then open AtomicPlayer.log. Are there any errors listed in the logfile?

ombre5733:
There are no errors in AtomicPlayer.log. If needed, I can upload the whole log. 

The CEF.log in the same directory shows these errors:

```
[0307/224712:ERROR:browser_main_loop.cc(231)] Running without the SUID sandbox! See https://chromium.googlesource.com/chromium/src/+/master/docs/linux_suid_sandbox_development.md for more information on developing with the sandbox on.
[0307/224712:ERROR:socket_posix.cc(124)] bind() returned an error, errno=98: Die Adresse wird bereits verwendet
[0307/224712:ERROR:devtools_http_handler.cc(228)] Cannot start http server for devtools. Stop devtools.
```

But I guess they are not related to the black screen.

JimMarlowe:
Those CEF errors will not impact the display. CEF does run without the SUID set, and the physics 2d platformer is not using the the webview. The bind error is from running a CEF instance in the editor, and starting up a 2nd instance in the player, and wants to use the same port address.
There have been problems in the past with incompatible ALSA drivers causing the SDL initialization to fail, the way you can test this, is by setting "export SDL_AUDIODRIVER=auto", keeping in mind that the player may lose sound, but gain visuals.

ombre5733:
Using "export SDL_AUDIODRIVER=auto" does not solve the black screen but I also lose the sound.

JimMarlowe:
When I play this example, it starts out with a window that has you select the time of day, are you seeing that?

ombre5733:
Yes it does so. I can then select to play daytime or nighttime. After the selection, the screen stays black. However, I can move around and hear sounds from jumping or collecting coins. By now, I wander if the camera is set up correctly.

ombre5733:
The platformer works when I use "export LC_NUMERIC=C" before starting the engine (my default locale is "de_AT.UTF-8"). I only found this solution because it was mentioned in #710.

JoshEngebretson:
Closed via #1466

[...]

PR #1477, opened 2017-03-15
Floating point values should be written with a dot as separator.

ombre5733:
Floating point values should always use a '.' as decimal separator and not the separator
determined by the user's locale. Otherwise, XML files written on one system with an
English locale cannot be parsed back correctly on a system with e.g. a German locale.

Replacements for strtod() and sprintf() have been collected in Core/Io.{h,cpp}. Since
sprintf() has a variadic parameter list, the replacement Atomic_sprintf() is implemented
as a macro for simplicity.

### DIFF
--- a/Source/Atomic/Container/Str.cpp
+++ b/Source/Atomic/Container/Str.cpp
@@ -32,6 +32,13 @@
 #pragma warning(disable:6293)
 #endif
 
+
+// ATOMIC BEGIN
+
+#include "../Core/Io.h"
+
+// ATOMIC END
+
 namespace Atomic
 {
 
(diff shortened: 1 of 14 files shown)


ombre5733:
@JoshEngebretson Please have a look at the OSX build again. Once this PR is merged, it should *really* fix #710 as well as #1465 and #1469.

JoshEngebretson:
I haven't run into this in other projects, instead of modifying C functions, can we set a env variable for the process, or globally set the locale at startup?  This is a problem only on Linux correct?  Android/iOS don't have locale support.

ombre5733:
Using `setlocale()` was actually my first try but I could not get it to work. One problem of `setlocale()` is that the locale is set globally. If one library modifies the global, the whole application will be affected (in the worst case, we run into race conditions, if one thread is `sprintf`-ing while another thread modifies the global).

According to the documentation link above, the locale should start in `C` mode, which is definitely not the case on my system (also `export LC_NUMERIC=C` would not work then).

One mitigation could be to use `uselocale()` instead of `setlocale()`, which modifies the locale only for the current thread...

I'll do some more investigations and return back if I have an answer.

ombre5733:
Forget what I have written above. I was partially wrong (the app starts correctly with the `C` locale as required by the standard). However, CEF initialization changes the locale. I have a much better (and smaller) fix in PR #1479.

PR #1482, opened 2017-03-17
[WIP] Update to Duktape 2.0.1

ombre5733:
The update of AtomicPlugin.h is missing. I haven't figured out yet, which files need to be changed exactly.


JoshEngebretson:
This is a great update, very cool.  Were you able to do much testing with the editor and examples with the update?  I'll do some today before landing as well.  

The AtomicPlugin for JS is something from a fair ways back.  It is really powerful in that you can drive the engine's VM using the duktape API from a shared library, without needing to link to the engine.  Though, I do not believe anyone has used it since it was originally written.

The build error is on Android, where I have had to fix the duk_config.h before: 

https://github.com/AtomicGameEngine/AtomicGameEngine/commit/36ead3d26bc24fa69d5cd12b3ba32b8b7505be3d

```
AtomicPR_MacEditor_iOS_Android/Source/ThirdParty/Duktape/duk_config.h:1778:2: error: #error cannot determine intptr type
 #error cannot determine intptr type
```

JoshEngebretson:
I added the ANDROID check and opened a new PR #1483, will do some testing now, and will land from there is no issues found.

If you want to update the plugin stuff, that's great, though I would only do that if you plan on using it (otherwise, just having it compiling is fine).

ombre5733:
Thanks for fixing the Android build. Also thanks for explaining AtomicPlugin. Now I understand what it is supposed for. 

I do not depend on it and do not want to spend time for fixing it. When it is fixed, I think we should duplicate the duktape headers in the plugin's directory and include them from AtomicPlugin.h rather than copying the content into AtomicPlugin.h. The additional files do not matter for the user because including AtomicPlugin.h will include the duktape headers as well.

However, right now the state of the plugin is somewhat dangerous, because it works most of the time except if you call the wrong function (some functions have a different signature now and some functions have been changed to macros, so the function pointer in `JSVMImports` is null). Wouldn't it be better if we comment the content of `jsplugin_bind_jsvmexports()` so that all function pointers are null? Using the plugin would result in a crash immediately upon the first call which is easier to debug than failing in a rarely used function.

Also I would like to take the opportunity and remove `duk_push_string_file_raw` from AtomicPlugin. This is the only reason why I had to include duktape's V1 compatibility layer. There is not much use of having an additional dependency for a functionality that is not usable (i.e. the plugin).

JoshEngebretson:
If the plugin duktape wrapper ```duk_push_string_file_raw ``` is the only reason we need the V1 compatibility layer, I am all for removing it 🔪 

I also filed an issue to remove the AtomicPlugin functionality entirely: https://github.com/AtomicGameEngine/AtomicGameEngine/issues/1485

ombre5733:
@JoshEngebretson Duktape V1 is gone with #1486.

PR #1486, opened 2017-03-17
Remove Duktape V1 compatibility module

ombre5733:
This module was only needed for the AtomicJS plugin, which is not in a good state anyway.
Rather than pulling in a dependency for a defunct feature, crop the feature a bit more
and remove the dependency.


JoshEngebretson:
Excellent, thanks!

PR #1490, opened 2017-03-21
Add code to allow hooking up vscode debugger

shaddockh:
@JoshEngebretson , @JimMarlowe 
This PR contains code extracted from the more involved branch that I'm working on to enable in-editor debugging.  This just includes the minimal amount of code required to use the recently released version of the VSCode Duktape debugger from: https://marketplace.visualstudio.com/items?itemName=HaroldBrenes.duk-debug

I have been working on a PR for that extension to ensure that it works with the way we are utilizing the duktape 1.5 require system and those changes have just been pulled in and he also released a version that works with Duktape 2.

With this PR, you can now kick off the player and attach to it from vscode.  The extension also properly handles sourcemaps.

- Point vscode to your project root
- Install the VSCode duktape debugger extension in vscode
- Then set up a task in vscode: **tasks.json** (You will need to point command to your editor location)

```json
{
    "version": "0.1.0",
    "tasks": [
        {
            "taskName": "Run Player",
            "command": "${env.HOME}/Development/Atomic/AtomicMaster/AtomicGameEngine/Artifacts/AtomicEditor/AtomicEditor.app/Contents/MacOS/AtomicEditor",
            "args": [
                "--player",
                "--debug",
                "--project",
                "${workspaceRoot}"
            ],
            "isBackground": true
        }
    ]
}
```
Then set up a debugger configuration:
**launch.json**
note: set sourceMaps to true if using TS and make sure you turn on sourcemaps in the tsconfig
```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Attach",
            "type": "duk",
            "request": "attach",
            "address": "localhost",
            "port": 9091,
            "localRoot": "${workspaceRoot}/Resources",
            "sourceMaps": true,
            "outDir": "${workspaceRoot}/Resources",
            "stopOnEntry": false,
            "debugLog": true
        }
    ]
}
```

- Launch the editor task ( ctrl/cmd+shift+p > Tasks: Run Task) then select the "Run Player" task
- go to debug in vscode and attach (the player will be paused waiting until you attach)
- now you can pause, set breakpoints, view variables, etc.

@JoshEngebretson 
The idea behind this is to get this functionality out there while work continues on trying to embed a debugging interface into the editor itself.  The functionality in the editor may not be as advanced, but will be available w/o any third party software needed.  The question I have at this point is would having the debugger built into the Atomic Editor even be necessary anymore?  We could follow the lead of how the C# debugging story is working and just require a 3rd party piece.

Changes:
- Configure duktape 2.x to enable debugging
- Migrate debug socket hooks from @JimMarlowe in IPCPlayerApp.cpp/h
- Updates to jsrequire to set the module filename correctly
- cleanup


JoshEngebretson:
Wow, cool! I'll check out debugging on Windows and Mac via VSCode.  I'll also have a look at pulling the duktape C raw socket API out of the IPCPlayer sources, either placing that in the third party sources or in the JavaScript library.  Looking sweet! 🍬 

I think focusing on external coding is a really good idea.  We can get basic in editor debugging working later and the embeded Monaco editor really isn't intended so much for heavy lifting, as quickly opening up and tweaking scripts... a debugger would be useful there, though a lot more powerful when leveraging VSCode proper with packages, etc

shaddockh:
It looks like the windows build failed.  I'm not currently set up to build on windows, but I suspect it might have something to do with the socket stuff since that is different between MacOS and Windows.  Pulling that all out into it's own files would definitely be an improvement and clean things up. 

I've been really impressed with vscode lately and having this debugger which can allow you to work against raw .js as well as .ts is pretty nice.  One thing that could happen is that the "generate external project" could make things easier and auto populate the various task and launch json files for you so all you need to do is launch vscode and go.

shaddockh:
@JoshEngebretson ok, this looks good.  I added in some logic to shutdown the socket since I noticed it was causing me to have to force quit the player even after closing the window.  Once this lands, I'll start putting together a wiki entry on how to use it. 🐫

JoshEngebretson:
https://github.com/AtomicGameEngine/AtomicGameEngine/pull/1490/commits/bf588c59b7f40ea39c818f97976eaad4ad5c3a66

This doesn't do anything other than set the raw static instance pointer, it doesn't release anything or close any sockets.  If the socket is keeping the process from exiting, the subsystem needs to close it upon shutdown/destruction.

shaddockh:
Ok, so basically what was happening is that when the window was being closed, the socket was automatically disconnecting as it should, but the logic in the JSDebugger subsystem was wired up to automatically start listening for a connection again whenever the socket disconnects.  This was causing the window to never actually close.

What I can do is have that Shutdown method set a flag that tells the socket to not try and reconnect.  The other option I can do is have the subsystem directly listen for a player close event, but wasn't sure if it was appropriate for the subsystem to be listening to player events.

JoshEngebretson:
Ok, yeah in ```do_reconnect()``` checking the instance for reconnect, so ```JSDebugger::Shutdown``` clearing that could "fix" it, though still could probably hit it wrong on exit and freeze thread.  (It would be nice to have a debugging thread, instead of freezing main thread for a future version).

Let's have shutdown clear the auto connect (and close any open sockets?), though not clear the static instance ptr and leave that to the destructor.

JoshEngebretson:
Oh, and we may want to use this outside of player apps, so good thinking on not coupling the Player subsystem.  A note being that it is possible the PlayerApp, which uses the Player subsystem, may want to subscribe to player close and do cleanup, something to consider when in there :)

shaddockh:
:+1:  I'll take a look at this tonight.  :camel:

Source/AtomicJS/Javascript/JSDebugger.h
```diff
@@ -0,0 +1,63 @@
...
+
+    /// Shut down the debugger
+    void Shutdown() const;
+
+    /// Get the JSDebugger subsystem from external code
+    static JSDebugger* GetInstance() { return instance_; }
+
+    /// Automatically listen for a new connection when disconnected
+    bool AutoReconnect;
+
```
JoshEngebretson: ```AutoReconnect``` should be private and the naming convention would be ```bool autoReconnect_;``` for a member variable.  Note, if you want this accessible outside of the class and available to scripting languages, Get/Set methods need to be added.

Source/AtomicJS/Javascript/JSDebugger.cpp
```diff
@@ -0,0 +1,862 @@
...
+namespace Atomic
+{
+
+JSDebugger* JSDebugger::instance_ = NULL;
+
+JSDebugger::JSDebugger(Context* context) : Object(context)
+{
+    instance_ = this;
+    this->AutoReconnect = false;
+}
```
JoshEngebretson: ```this->``` isn't needed here, also example of convention to initialize member variables: https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/AtomicJS/Javascript/JSVM.cpp#L58

Source/AtomicJS/Javascript/JSDebugger.cpp
```diff
@@ -0,0 +1,862 @@
...
+
+JSDebugger::~JSDebugger()
+{
+    instance_ = NULL;
+}
+
+void JSDebugger::Shutdown() const
+{
+    duk_trans_socket_finish();
```
JoshEngebretson: Should ```Shutdown``` ensure that auto reconnect is turned off, that way don't need to depend on calling sites setting it?

Source/AtomicApp/Player/IPCPlayerApp.cpp
```diff
@@ -196,6 +196,13 @@ namespace Atomic
 
         GetSubsystem<Graphics>()->RaiseWindow();
 
+        if (debugPlayer_)
+        {
+            ATOMIC_LOGDEBUG("Starting JSDebugger Subsystem");
+            context_->RegisterSubsystem(new JSDebugger(context_));
+            context_->GetSubsystem<JSDebugger>()->AutoReconnect = true;
```
JoshEngebretson: This should be defaulted to true in the JSDebugger constuctor, unneeded here

Source/AtomicJS/Javascript/JSDebugger.cpp
```diff
@@ -0,0 +1,862 @@
...
+JSDebugger::JSDebugger(Context* context) : Object(context)
+{
+    instance_ = this;
+    this->AutoReconnect = false;
+}
+
+JSDebugger::~JSDebugger()
+{
+    instance_ = NULL;
+}
```
JoshEngebretson: Call ```Shutdown``` upon subsystem destruction?  I am thinking so... that way don't need to be explicit

Source/AtomicApp/Player/IPCPlayerApp.cpp
```diff
@@ -206,6 +213,12 @@ namespace Atomic
 
     void IPCPlayerApp::Stop()
     {
+        if (debugPlayer_)
+        {
+            context_->GetSubsystem<JSDebugger>()->AutoReconnect = false;
+            context_->GetSubsystem<JSDebugger>()->Shutdown();
```
JoshEngebretson: ```Shutdown``` can clear auto reconnect and doesn't need to be set externally, also do we need to be explicitly shutting down subsystem here?  What if the JSDebugger destructor calls ```Shutdown```, see note in destructor.

shaddockh:
Thanks for the feedback, still getting up to speed on best practices under the c++ side of things. :)

Source/AtomicApp/Player/IPCPlayerApp.cpp
shaddockh: The problem here is that once the socket disconnects, it starts listening for a connection again, which blocks the destructor from firing and hangs the window.  Something needs to notify the socket to not reconnect when shutting down.

JoshEngebretson: ok, so shutting down explicitly it is, probably can have ```Shutdown``` turn off the auto connect

shaddockh:
Ok, I believe I've got it cleaned up.  Unfortunately you still need to call shutdown from the IPC Player due to how the socket is automatically reconnecting.  The destructor never fires because the socket immediately spins back up listening for a connection and blocks the app once you try to close the window.

The only other thing I can think of is maybe listening for the ```E_EXITREQUESTED``` event inside the JSDebugger.  If we want to go that route, I will need a little guidance.  I wasn't able to get it to compile when I tried.

JoshEngebretson:
I thought about the event subscription too, that's a little inflexible and couples the code via an event rather than the classes. Needing to startup/shutdown the debugger subsystem is just fine and dandy and the rest looks great, landing, thanks! ✈️

PR #1492, opened 2017-03-22
fix JS error scraper for DukTape v2

JimMarlowe:
DukTape 2.0.1 changed the formatting of error messages! So the old JS error scraper is now crash code, because it leaves in a paren in the error filename, and things downstream crash when an invalid filename is presented to them.  The new error syntax also reduces false positives, I'd seen it put an icon in dates at times.


JoshEngebretson:
Good catch! ⚾️ 

Thanks, LANDING! ✈️

PR #1493, opened 2017-03-22
JSDebugger subsystem

JoshEngebretson:
Here's a refactor to the new JS/TypeScript debugger support which breaks it out to a new ```JSDebugger``` subsystem 🐛, we can also expose some of this to script now if desired.  

Includes some compilation fixes on Windows, I also ran it on Windows and verified it works... it does, and that is awwweeessssoooommmmmeee!!! ⚡️ 

@shaddockh Please have a look, if ok go ahead and land and it will update your PR branch 🐫💨


shaddockh:
Cool!  I'll pull this in and test it out.

PR #1494, opened 2017-03-23
Add the GetResourceDirs function into C# ResourceCache

weinandvv:
Added this function since Vector of Strings isn't implemented for general bindings and thus couldn't be exposed to C# from C++.

### DIFF
--- a/Script/AtomicNET/AtomicNET/Resource/ResourceCache.cs
+++ b/Script/AtomicNET/AtomicNET/Resource/ResourceCache.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.Runtime.InteropServices;
 
 namespace AtomicEngine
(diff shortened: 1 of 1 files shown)


Script/AtomicNET/AtomicNET/Resource/ResourceCache.cs
```diff
@@ -29,6 +30,20 @@ public System.IO.Stream GetFileStream(string name, bool sendEventOnFailure = tru
             return null;
         }
 
+        public List<string> GetResourceDirs()
```
JoshEngebretson: Please add a basic comment summary for the method, see the ```ReleaseAllResources``` method immediately following  for an example.  Choosing the container return type is important,  Array, Vector, and List are all choices here.  Array is the best one as the return value isn't intended to be modified with additions/subtractions (Vector), or randomly inserted into (List)

Script/AtomicNET/AtomicNET/Resource/ResourceCache.cs
```diff
@@ -29,6 +30,20 @@ public System.IO.Stream GetFileStream(string name, bool sendEventOnFailure = tru
...
+        {
+            List<string> resourceDirs = new List<string>();
+
+            ResourceCache cache = AtomicNET.GetSubsystem<ResourceCache>();
+
```
JoshEngebretson: When descended from Object, there is no need to prefix this with ```AtomicNET```, the static ```AtomicNET.GetSubsystem``` is useful when in external code.

JoshEngebretson:
@weinandvv Awesome, thanks for the changes, looks great 👍  

PR #1500, opened 2017-03-28
Expose the webView WebMessageHandler to script and fix an event signature

shaddockh:
Simple bindings update to expose the WebMessageHandler to script


JoshEngebretson:
Nice one, short and sweet! LANDING!

PR #1508, opened 2017-04-05
Database subsystem

rokups:
Database subsystem changes:
* Synced with upstream
* Namespaces/macros converted to atomic names
* Added to build system
* SQLite updated to v3.18.0

I also smuggled in tiny android fix that gets rid of build error and changes log tag from Urho3D to Atomic. I hope it is acceptable.


JoshEngebretson:

Source/Atomic/IO/Log.cpp
```diff
@@ -228,10 +228,10 @@ void Log::WriteRaw(const String& message, bool error)
     if (logInstance->quiet_)
     {
         if (error)
-            __android_log_print(ANDROID_LOG_ERROR, "Urho3D", message.CString());
+            __android_log_print(ANDROID_LOG_ERROR, "Atomic", "%s", message.CString());
     }
```
JoshEngebretson: Maybe pass this fix upstream, otherwise good idea to add an // ATOMIC BEGIN/END here with description of change, helps to not clobber changes when pulling updates

JoshEngebretson:
Weeeeird, the build bot isn't seeing hash 65081f7 and seems to be content testing d091c0b repeatedly

JoshEngebretson:

JoshEngebretson:
Another great PR, noted a possible follow up, thanks! 👍

Source/Atomic/IO/Log.cpp
rokups: Upstream has this [fix](https://github.com/urho3d/Urho3D/blob/master/Source/Urho3D/IO/Log.cpp#L231) already 🐫

PR #1509, opened 2017-04-05
Do not build AtomicPlayer when javascript support is disabled.

rokups:
Overlooked this one somehow. Should be all of it now as my project happily builds with engine in a subdir.

### DIFF
--- a/Source/CMakeLists.txt
+++ b/Source/CMakeLists.txt
@@ -4,9 +4,9 @@ option(ATOMIC_EDITOR "Build Editor" ON)
 
 add_subdirectory(ThirdParty)
 add_subdirectory(Atomic)
-add_subdirectory(AtomicPlayer)
 
 if (ATOMIC_JAVASCRIPT)
+    add_subdirectory(AtomicPlayer)
     add_subdirectory(AtomicJS)
     add_subdirectory(AtomicPlayerJS)
 endif ()


JoshEngebretson:
Great!, maybe we can get this into the AtomicExamples repo as example of how to include Atomic in a external C++ project without JS/C#? https://github.com/rokups/ShaderSketch

Landing ✈️, thanks!

rokups:
I am not sure how to go about this. Whole point of such example would be having an engine as a submodule. Thing is AtomicExamples is a submodule of engine. This would create a weird loop. So maybe better idea would be a separate repo?

JoshEngebretson:
AtomicExamples is checked out as a submodule, solely to easily have it available for development and for packaging examples with editor, it can also be checked out top level.

Would be good to have it in example repo instead of a new one, maybe the CMake for the example could detect if it is in the engine source tree and error, how are you setting up Atomic in your project for add_directory()?

rokups:
> AtomicExamples is checked out as a submodule, solely to easily have it available for development and for packaging examples with editor, it can also be checked out top level.

Would it not create a dependency loop of sorts if both repositories include each other in a submodule? For example if someone thinks of running git submodule checkout with `--recursive`.

> how are you setting up Atomic in your project for add_subdirectory()?

I updated ShaderSketch with latest engine code, take a look at [CMakeLists.txt](https://github.com/rokups/ShaderSketch/blob/master/CMakeLists.txt). Basically it boils down to:

1. Configure engine with `set(ATOMIC_<...> 0 CACHE BOOL "" FORCE)`
2. `add_subdirectory(path/to/AtomicGameEngine)`
3. `target_link_libraries(TargetName Atomic)`

And that is it. Includes, defines and dependencies automatically pulled in thanks to CMake ;)

PR #1513, opened 2017-04-08
Changes to integrate better with VSCode

shaddockh:
- generate vscode tasks file for launching the player in debug mode or non-debug mode
- generate vscode for utilizing the duktape debugger vscode extension
- added a menu item to allow pure JavaScript projects to build out the vscode project for editing under vscode.  Latest vscode actually reads d.ts files for completions, even for JavaScript projects
- modify the project based extension loader to not try to load .map files as extensions [bug]
- modify the project based extension loader to handle loading files that have the source map line on the last line of the file w/o an extra line feed

### DIFF
--- a/Script/AtomicEditor/hostExtensions/coreExtensions/ProjectBasedExtensionLoader.ts
+++ b/Script/AtomicEditor/hostExtensions/coreExtensions/ProjectBasedExtensionLoader.ts
@@ -60,8 +60,8 @@ export default class ProjectBasedExtensionLoader extends Atomic.ScriptObject imp
      * file system and manually pull these in to provide to duktape
      */
     private rewriteModSearch() {
-        Duktape.modSearch = (function(origModSearch) {
-            return function(id: string, require, exports, module) {
+        Duktape.modSearch = (function (origModSearch) {
+            return function (id: string, require, exports, module) {
                 let system = ToolCore.getToolSystem();
                 if (id.indexOf(ProjectBasedExtensionLoader.duktapeRequirePrefix) == 0) {
                     let path = id.substr(ProjectBasedExtensionLoader.duktapeRequirePrefix.length) + ".js";
@@ -75,7 +75,10 @@ export default class ProjectBasedExtensionLoader extends Atomic.ScriptObject imp
                         if (Atomic.fileSystem.fileExists(path)) {
                             let include = new Atomic.File(path, Atomic.FileMode.FILE_READ);
                             try {
-                                return include.readText();
+                                // add a newline to handle situations where sourcemaps are used.  Duktape
+                                // doesn't like not having a trailing newline and the sourcemap process doesn't
+                                // add one.
+                                return include.readText() + "\n";
                             } finally {
                                 include.close();
                             }
@@ -107,7 +110,7 @@ export default class ProjectBasedExtensionLoader extends Atomic.ScriptObject imp
                 filenames.forEach((filename) => {
                     // Filtered search in Atomic doesn't due true wildcarding, only handles extension filters
                     // in the future this may be better handled with some kind of manifest file
-                    if (filename.toLowerCase().lastIndexOf(".plugin.js") >= 0) {
+                    if (filename.search(/\.plugin.js$/i) != -1) {
                         var extensionPath = editorScriptsPath + filename;
                         extensionPath = extensionPath.substring(0, extensionPath.length - 3);
 
--- a/Script/AtomicEditor/hostExtensions/languageExtensions/TypescriptLanguageExtension.ts
+++ b/Script/AtomicEditor/hostExtensions/languageExtensions/TypescriptLanguageExtension.ts
@@ -26,19 +26,22 @@
  * then the one in the project will overwrite these
  * @type {ts.CompilerOptions}
  */
-const defaultCompilerOptions = {
-    noEmitOnError: true,
-    noImplicitAny: false,
-    target: "es5",
-    module: "commonjs",
-    declaration: false,
-    inlineSourceMap: false,
-    removeComments: false,
-    noLib: false,
-    forceConsistentCasingInFileNames: true,
-    allowJs: true,
-    lib: ["es5"]
-};
+function getDefaultCompilerOptions() {
+    return {
+        noEmitOnError: true,
+        noImplicitAny: false,
+        target: "es5",
+        module: "commonjs",
+        declaration: false,
+        inlineSourceMap: false,
+        sourceMap: false,
+        removeComments: false,
+        noLib: false,
+        forceConsistentCasingInFileNames: true,
+        allowJs: true,
+        lib: ["es5"]
+    };
+}
 
 /**
  * Resource extension that supports the web view typescript extension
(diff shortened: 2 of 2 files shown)


JoshEngebretson:
Cool, will look at this soon! We lost power again and it just came back this morning, backlogged :) ⚡️

JoshEngebretson:
Some testing on Windows 10, I was able to create an external project and load it into VSCode, both for JS/TS, very cool!  

It did get me thinking about folder layout, like maybe we want to support code outside of Resources, though that means transpiling/amalgamating so the script can be loaded from packaged resources on device, etc.  Might not be a big deal, and this is kind of a JS norm (ie. your scripts end up processed into an output folder)

In new 3D TypeScript project, I seem unable to set breakpoints in TS, ```this.node.yaw(timeStep * 75 * this.speed);``` TS not hit in the Spinner component, though can set in the transpiled js.

In testing an external JavaScript, I was able to hit the breakpoint.  I am wondering what is the best way to kick off intellisense using the ambient typings in a JS project under VSCode?

Get this warning using Launch/Debug Atomic Player in VSCode console:

```
[0408/140133:ERROR:tcp_socket_win.cc(338)] bind() returned an error: Only one usage of each socket address (protocol/network address/port) is normally permitted. (0x2740)
[0408/140133:ERROR:devtools_http_handler.cc(228)] Cannot start http server for devtools. Stop devtools.
```

Script/AtomicEditor/hostExtensions/languageExtensions/TypescriptLanguageExtension.ts
```diff
@@ -362,57 +372,122 @@ export default class TypescriptLanguageExtension extends Atomic.ScriptObject imp
...
+                        "--player",
+                        "--debug",
+                        "--project",
+                        "${workspaceRoot}"
+                    ],
+                    "isBackground": true
+                },
+                {
+                    "taskName": "Lauch Atomic Player",
+                    "command": `${ToolCore.toolEnvironment.editorBinary}`,
```
JoshEngebretson: Lauch => Launch

Script/AtomicEditor/hostExtensions/coreExtensions/ProjectBasedExtensionLoader.ts
```diff
@@ -75,7 +75,10 @@ export default class ProjectBasedExtensionLoader extends Atomic.ScriptObject imp
                         if (Atomic.fileSystem.fileExists(path)) {
                             let include = new Atomic.File(path, Atomic.FileMode.FILE_READ);
                             try {
-                                return include.readText();
+                                // add a newline to handle situations where sourcemaps are used.  Duktape
+                                // doesn't like not having a trailing newline and the sourcemap process doesn't
+                                // add one.
+                                return include.readText() + "\n";
```
JoshEngebretson: Interesting, maybe this is fixed or should be filed upstream?

Script/AtomicEditor/hostExtensions/coreExtensions/ProjectBasedExtensionLoader.ts
```diff
@@ -107,7 +110,7 @@ export default class ProjectBasedExtensionLoader extends Atomic.ScriptObject imp
                 filenames.forEach((filename) => {
                     // Filtered search in Atomic doesn't due true wildcarding, only handles extension filters
                     // in the future this may be better handled with some kind of manifest file
-                    if (filename.toLowerCase().lastIndexOf(".plugin.js") >= 0) {
+                    if (filename.search(/\.plugin.js$/i) != -1) {
```
JoshEngebretson: Regex is so 🔥 !

Script/AtomicEditor/hostExtensions/languageExtensions/TypescriptLanguageExtension.ts
```diff
@@ -362,57 +372,122 @@ export default class TypescriptLanguageExtension extends Atomic.ScriptObject imp
...
+                        "."
+                    ],
+                    "showOutput": "always",
+                    "problemMatcher": "$tsc",
+                    "isBuildCommand": true
+                },
+                {
+                    "taskName": "Debug Atomic Player",
+                    "command": `${ToolCore.toolEnvironment.editorBinary}`,
+                    "args": [
```
JoshEngebretson: Cool, love the interaction between VSCode and Atomic's ToolCore 👍, looking forward to having a look at VSCode + NET Core + Atomic...

Script/AtomicEditor/hostExtensions/languageExtensions/TypescriptLanguageExtension.ts
shaddockh: bleh...typo :(

Script/AtomicEditor/hostExtensions/coreExtensions/ProjectBasedExtensionLoader.ts
shaddockh: This is something that happens when source maps are used.  The source map index get's appended to the last line of the JS file w/o a newline.  As far as I can tell, this is normal for source maps, however duktape doesn't like it.  This is already being handled in the JSRequire cpp file at:
https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/AtomicJS/Javascript/JSRequire.cpp#L165-L167

so the logic that loads the extensions are doing the same thing.

shaddockh:
> In new 3D TypeScript project, I seem unable to set breakpoints in TS, this.node.yaw(timeStep * 75 * this.speed); TS not hit in the Spinner component, though can set in the transpiled js.

You need to make sure you have source maps turned on in your ```tsconfig.json``` and also source maps turned on in your duktape debugger launch config.  Do you think this should be something turned on by default when generating external TS projects?

>I am wondering what is the best way to kick off intellisense using the ambient typings in a JS project under VSCode?

It looks like there needs to be a ```jsconfig.json``` file in the root for intellisense to start working.  I'll go ahead and add that.

shaddockh:
> Get this warning using Launch/Debug Atomic Player in VSCode console:

[0408/140133:ERROR:tcp_socket_win.cc(338)] bind() returned an error: Only one usage of each socket address (protocol/network address/port) is normally permitted. (0x2740)
[0408/140133:ERROR:devtools_http_handler.cc(228)] Cannot start http server for devtools. Stop

I think this has been happening for a while.  It's the CEF listening on the debug port in both the editor and the player and causing a conflict.  If you launch the player from VSCode and don't have the editor running, this error doesn't appear.  It might be a good idea to make the CEF debug port configurable from the command line.

shaddockh:
I was thinking about it last night, and if you are editing TS in an external project more likely than not you will want to debug the TS and not the generated JS.  Thinking that, I went ahead and put in the sourcemap configuration for TS and the duktape debugger when generating the external project....just seems to make more sense to turn it on from the outset and then turn it off if you need to later.

JoshEngebretson:
Glorious!  I tested it out on and was able to hit TS/JS breakpoints and intellisense is working *great* in JavaScript code, those typings are such a boon :)  I agree on the source maps on by default for TS, seemed like expected behavior in testing it.

Thanks! This seriously levels up the VSCode integration, making for some darned pleasant Atomic scripting! ⛷️:dash:

PR #1514, opened 2017-04-08
enable android source debug

JimMarlowe:
This PR will allow users to insert gdbserver or libgdbserver.so into their APK so it may be debugged with GDB on an external host. Some versions of Android can not copy an executable into the lib directory of an APK, so the work around is to rename it lib*.so, and then execute it anyway, so both options are given.
I fixed the android player to not load anything that has the string "gdbserver" as a shared library, as this borks the resulting application.
And I fixed the push action to the android device to install Atomic-debug.apk instead of Atomic-debug-unaligned.apk, which is an intermediate file.
The Android wiki page needs to be updated to explain the new options.


JoshEngebretson:
Excellent!  Nice tidy delta, looking forward to trying it out, and will really help debug native code on the platform, thanks, LANDING! ✈️

PR #1520, opened 2017-04-15
Jme atomic urho thirdparty update

rokups:
Replaced "urho3d" strings with "atomic"

Fixed crash when calling `UI::GetWidgetAt()` without initializing UI subsystem. Engine makes this call automatically. Since i do not use TB UI it crashes application.

Removed assert that caused crash when using c++11 `std::bind()` for subscribing to events. Fix is on upstream since i added support for c++11 callbacks.

Fixed SDL build, at least on my system it failed to discover `dl` due to `check_c_source_compiles()` missing `dl` link library.

Fixed dependency issue on windows when building without .NET and javascript.


Source/CMakeLists.txt
```diff
@@ -18,15 +22,15 @@ if (ATOMIC_WEBVIEW)
...
-        add_subdirectory(AtomicNET)
-    endif ()
+    add_subdirectory(AtomicNET)
 endif ()
 
 if (ATOMIC_DESKTOP)
-    add_subdirectory(AtomicTool)
+    if (ATOMIC_DOTNET OR ATOMIC_JAVASCRIPT)
+        add_subdirectory(AtomicTool)
+    endif ()
```
JoshEngebretson: Nice one, sure this saves a bit when not using the NET or JavaScript languages

Source/CMakeLists.txt
```diff
@@ -1,5 +1,9 @@
 option(ATOMIC_JAVASCRIPT "Build JS Bindings" ON)
-option(ATOMIC_DOTNET "Build .NET Bindings" ON)
+if (WEB)
+    set(ATOMIC_DOTNET OFF)
```
JoshEngebretson: Good catch ⚾️

Source/ThirdParty/SDL/cmake/sdlchecks.cmake
```diff
@@ -76,6 +78,13 @@ macro(CheckDLOPEN)
...
         endforeach()
       endif()
+      # ATOMIC BEGIN
+      # call check_function_exists() only after trying to locate dlopen in the libs. This provides lib
+      # in _DLLIB which is required for linking when calling check_c_source_compiles().
+      if(NOT DLOPEN_FOUND)
+        check_function_exists(dlopen DLOPEN_FOUND)
+      endif()
+      # ATOMIC END
       if(DLOPEN_FOUND)
```
JoshEngebretson: 👍

PR #1521, opened 2017-04-17
Fix broken headings in Markdown files

bryant1410:
GitHub changed the way Markdown headings are parsed, so this change fixes it.

See [bryant1410/readmesfix](https://github.com/bryant1410/readmesfix) for more information.

Tackles bryant1410/readmesfix#1


JoshEngebretson:
@bryant1410 Thanks, though when sending scripted PR's en masse, probably a good idea to fix the script to not alter file whitespace globally, including markdown files which don't need changes.  

Thanks, closing and will fix in a separate PR

bryant1410:
> @bryant1410 Thanks, though when sending scripted PR's en masse, probably a good idea to fix the script to not alter file whitespace globally, including markdown files which don't need changes.
> 
> Closing and will fix in a separate PR

Yup, sorry. Will fix to avoid changing LF/CRLF. Btw, I tried to consider bugs by running the script incrementally and checking, and already fixed some bugs.

Issue #1209, opened 2016-11-27
[macOS] Startup error when running binary app bundle: Failed to add resource path 'CoreData'


smelukov:
- Download
- Execute
- Got error: `Failed to add resource path 'CoreData', check the documentation on how to set the 'resource prefix path'`

Can't find `how to set the 'resource prefix path'`

MacOS 10.12.1

JoshEngebretson:
Does your app path have a space in it?  This issue had been sporadically seen and thought solved

smelukov:
>Does your app path have a space in it?

No - `/Applications/AtomicEditor.app`

JoshEngebretson:
Thanks for the report, If you run it from your Downloads folder instead of /Applications does the issue persist?

smelukov:
@JoshEngebretson yep

JoshEngebretson:
Just to make sure, this is the version downloaded?

http://docs.atomicgameengine.com/downloads/AtomicEditor_MacOSX_a0505c2173c28314ab7039aba214924e49b6f742.zip

If you wouldn't mind rebooting and seeing if the issue magically goes away, that would be good to know.  The last time this popped up, quoting paths in the binary "fixed it".   Starting to think it maybe something with current directory and resolving the packaged resource folder.

smelukov:
@JoshEngebretson 
> Just to make sure, this is the version downloaded?

yep

> If you wouldn't mind rebooting and seeing if the issue magically goes away, that would be good to know.

The issue is still remaining... even after reboot

JoshEngebretson:
How about opening a terminal and running ./AtomicEditor from inside the app bundle?

smelukov:
@JoshEngebretson It works!

smelukov:
Thanks! But what's wrong with it?

JoshEngebretson:
That is a very good question :)  I haven't seen this personally, though have in the office and when haven't since last fixed.  I have a feeling it is the current folder, need to look into it.  I will assign to Build and look into it soon.

JoshEngebretson:
[...]

PR #1526, opened 2017-04-19
activate webgl builds

JimMarlowe:
This pr fixes some compilation errors that cause the WebGL platform not to compile. 
I am concerned about the root cause of #1209 since then function it was implemented in, FileSystem::GetProgramDir() was totally rewritten and the original fix may not work.
It also adds the nav library to the build, otherwise the webgl player crashes.
On a sunny note, this webgl player actually plays audio, where in the past it had been stubbornly silent.


Source/Atomic/IO/FileSystem.cpp
```diff
@@ -721,25 +721,9 @@ String FileSystem::GetProgramDir() const
...
-    // current working directory instead
-    /// \todo Should not rely on such fixed convention
-    String currentDir = GetCurrentDir();
-    if (!DirExists(programDir_ + "CoreData") && !DirExists(programDir_ + "Data") &&
-        (DirExists(currentDir + "CoreData") || DirExists(currentDir + "Data")))
-        programDir_ = currentDir;
-
-#endif
-// ATOMIC END
     return GetCurrentDir();
```
JoshEngebretson: Yup, we can lose this Atomic block, good call 👍

Source/Atomic/CMakeLists.txt
```diff
@@ -55,10 +55,10 @@ set (SOURCE_FILES ${CONTAINER_SOURCE} ${CORE_SOURCE} ${ENGINE_SOURCE} ${INPUT_SO
                   ${ATOMIC2D_SOURCE} ${ENVIRONMENT_SOURCE}
                   ${SCENE_SOURCE} ${UI_SOURCE} ${SYSTEM_UI_SOURCE}
                   ${WEB_SOURCE} ${SCRIPT_SOURCE} ${METRICS_SOURCE}
-                  ${PLATFORM_SOURCE} ${DATABASE_SOURCE} ${IK_SOURCE})
+                  ${PLATFORM_SOURCE} ${DATABASE_SOURCE} ${IK_SOURCE} ${NAVIGATION_SOURCE})
```
JoshEngebretson: The Navigation subsystem was originally left out of the WebGL build to reduce the size of generated JS, with WASM around the corner, think we can default this on now

PR #1531, opened 2017-04-28
easy_profiler integration

rokups:
This PR integrates [easy_profiler](https://github.com/yse/easy_profiler) replacing old and not really useful on-screen profiler.

Features:

* Profiling data is cached and sent over network (easy mobile device and application startup profiling)
* Profiling data can be saved to file and loaded later
* Profiler recognizes different threads
* Options for changing profiled block color (displayed in profiler tool). Events are orange, resource loading is blue
* Profiler tool is using Qt5. On systems where Qt is not available cmake will display warning but engine will still build
* C# support, requires projects to define `ATOMIC_PROFILING`.
* Profiler is enabled automatically for debug builds
* Command line parameters to set profiler port and enable/disable application listening for profiler connections

In the future easy_profiler maintainers plan support for dx11/12, opengl, vulkan and cuda profiling. Maintainers were very helpful and implemented non-scoped profiler blocks on my request. Tracking arbitrary values (my request as well) is already in the works, this will be useful for tracking c#/js object counts and what not.

I found a way to do reasonably low-overhead c# code block profiling using lambda. There is still a chance to optimize a little by defining custom bindings to `Profiler::BeginBlock()` which takes `const char*` as parameter. It looks like this:
```cs
Profiler.Block("BlockName", () => {
    // Code here
});
```
I did not add js support, however it is trivial. Is there a spot somewhere we could add some extra code for typescript to use much like `Script/AtomicNET/AtomicNET/Core/Profiler.cs`? In js it would look like so:
```js
Profiler.block("BlockName", function() {
    // Code here
});
```
Can we set `engineParameters_` from c#/js applications? If we can - how? We need this in order to make listening for profiler connections opt-in instead of opt-out.

TODO:

- [x] js support
- [x] Add `ATOMIC_PROFILING` to debug configurations of C# projects
- [x] Make profiler not listen by default
- [x] Add manual optimized bindings for `Profiler::BeginBlock()`

Please review.


JoshEngebretson:
Great, thanks 👍,  I'll give this a whirl later today once I can get back to macOS

JoshEngebretson:
Great work!  It is a lot more stable on macOS and the UI seems to be working other than the thread ID's which are still not quite being calculated properly.  I did try and disconnect and then reconnect and the Qt profiler app hangs on macOS when doing this... can you disconnect/reconnect for you?

screen shot 2017-05-16 at 12 06 08 pm

JoshEngebretson:
I was able to profile on Android (awesome! 🚀 ) and the thread id's appear to be the same number:

screen shot 2017-05-16 at 12 35 03 pm

rokups:
@JoshEngebretson easy_profiler maintainers did some work on networking and threading. Our changes were upstreamed as well. I rebased this branch on to master, cleaned it up, updated easy_profiler, fixed crash with event profiling (its a bit dirty, but works great). I squashed all commits as they were more wandering in the woods than meaningful work.. Squashing ate your changes, but rest assured your mark in history of humanity [was presedved](https://github.com/yse/easy_profiler/commit/a01187cf552f31b6bbfb1c165fc3d210f588f0aa) :)

Please test profiler tool on MacOS. I commented out that `socket.listen()` loop for now and i hope it will work without said loop and i can clean it up from commit history altogether.

JoshEngebretson:
@rokups Awesome sauce!  Thanks, and will test and review as soon as possible 👍

cas4ey:
@JoshEngebretson Hello! I'm collaborator of [easy_profiler](https://github.com/yse/easy_profiler). I saw you had a crash in GUI. Does the problem still exist? If yes, could you, please, send me .prof file with which you had the problem?

JoshEngebretson:
@Hello @cas4ey , thanks for your **excellent** work on easy_profiler!  I have not reproduced any crashes in recent testing of the profiler branch, will  let you know if I do 👍 

easyprofiler

rokups:
@JoshEngebretson i updated to latest profiler code and rebased on to engine master. Connection log spam is no longer there and it seems network quirks are fixed. Please test how it fares on MacOS.

Also please point me to some example of manual binding example for c# and to some example how to add extra code for js. I would like to try adding something like [this](https://github.com/rokups/AtomicGameEngine/blob/88fd006085fdafa4a750d81d7fd0d89c75834964/Script/AtomicNET/AtomicNET/Core/Profiler.cs) to javascript.

JoshEngebretson:
Awesome, I'll be able to test this later today.  I'll also provide some guidance for C#/JS 🐫

rokups:
Something weird is going on here and i cant figure it out. Could you please take a look? Search build log for `-- Library type: SHARED` (easy profiler library type). Due to [this](https://github.com/rokups/AtomicGameEngine/blob/feature/easy_profiler/Source/ThirdParty/CMakeLists.txt#L99) it should be STATIC. This variable is created [here](https://github.com/rokups/AtomicGameEngine/blob/feature/easy_profiler/Source/ThirdParty/easy_profiler/easy_profiler_core/CMakeLists.txt#L11), but according to cmake docs it only creates CACHE variable if it is missing. Otherwise value is left intact. I confirmed it testing on my machine. Something odd is going on here. I also had to add [this](https://github.com/rokups/AtomicGameEngine/blob/feature/easy_profiler/Source/ThirdParty/easy_profiler/easy_profiler_core/CMakeLists.txt#L135) workaround for mac build because it didnt work with `CXX_STANDARD 11` even though cmake version is high enough to support it. If you got any ideas whats going on here i am all ears.

JoshEngebretson:
I am not sure on the CMake workaround, 3.1.2 is what is on build box.  If we're modifying the stock easy_profiler CMake, please mark with // ATOMIC BEGIN and // ATOMIC END block with a note about modification, this helps when updating and also removing changes when they are no longer necessary.  

I'll have a look at the build when I get a chance here, won't be too long ;)  Great work!

JoshEngebretson:

rokups:

Source/ThirdParty/easy_profiler/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+
+set(EASY_PROGRAM_VERSION_MAJOR 1)
+set(EASY_PROGRAM_VERSION_MINOR 3)
+set(EASY_PROGRAM_VERSION_PATCH 0)
+set(EASY_PRODUCT_VERSION_STRING "${EASY_PROGRAM_VERSION_MAJOR}.${EASY_PROGRAM_VERSION_MINOR}.${EASY_PROGRAM_VERSION_PATCH}")
+
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_CURRENT_LIST_DIR}/bin)
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_CURRENT_LIST_DIR}/bin)
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_CURRENT_LIST_DIR}/bin)
+
```
JoshEngebretson: These output overrides cause an error when building for multiple platforms as each platform's binaries will be in the same location, static easy_profiler libs for all platforms are written to:

```%ATOMIC_ROOT%/Source/ThirdParty/easy_profiler/easy_profiler_core/Release```

instead of for example:

```%ATOMIC_ROOT%/Artifacts/Build/Mac/Source/ThirdParty/easy_profiler/easy_profiler_core/Release```
```%ATOMIC_ROOT%/Artifacts/Build/IOS/Source/ThirdParty/easy_profiler/easy_profiler_core/Release```
```%ATOMIC_ROOT%/Artifacts/Build/Android/Source/ThirdParty/easy_profiler/easy_profiler_core/Release```

Disabling these output directory overrides, and ```./Build_AtomicEditor.sh --with-android --with-ios``` works.  I think this is a fine solution, though should put // ATOMIC BLOCKS and a description of why disabled.  This isn't standard CMake for a third party library and should probably be addressed upstream.

**Comment by rokups** [comment]
Finally it builds \o/ @JoshEngebretson thank you very much for spotting that easy_profiler cmake quirk. Fresh set of eyeballs are a great thing o/ I will try to upstream that change as well.

**Comment by rokups** [comment]
I am afraid at least on windows profiler deployment is complicated.

`profiler_gui` can not build if engine links to static runtime because prebuilt Qt libraries link to dynamic msvc runtime. `profiler_gui` links to `easy_profiler` lib here, and lib inherits static runtime from the engine.

We can build `profiler_gui` with `-DATOMIC_DYNAMIC_RUNTIME=1`, however then editor and few other targets fail to build because CEF links to static runtime.

We have to do one of the following on windows:

* Build Qt linking to static msvc runtime.
* Build CEF linking to dynamic msvc runtime.
* Generate `-DATOMIC_DYNAMIC_RUNTIME=0` and build engine without providing it access to Qt. Generate build with `-DATOMIC_DYNAMIC_RUNTIME=1` and build `profiler_gui` only. Copy profiler tool with Qt dlls to static build directory for distribution.

File: Source/ThirdParty/easy_profiler/CMakeLists.txt [hunk 2]
```diff
@@ -25,7 +27,13 @@ macro(easy_define_target_option TARGET SOURCE_OPTION TARGET_DEFINITION)
 endmacro()
 
 add_subdirectory(easy_profiler_core)
-add_subdirectory(profiler_gui)
+# ATOMIC BEGIN
+if (MSVC AND "${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
+    message(STATUS "Profiler GUI can only be built in Release builds with MSVC compiler.")
```
JoshEngebretson: Does this cause the easy profiler cmake to look for debug Qt libs?  If so, maybe can be configured to use Release Qt libs regardless in easy profiler scripts?  If not, or another reason, lemme know.  I think this is acceptable, though would be good to have Profiler client available in debug builds, which could also be sorted later.

Source/ThirdParty/easy_profiler/easy_profiler_core/CMakeLists.txt
```diff
@@ -143,6 +143,8 @@ else ()
     endif ()
 endif ()
 
+# ATOMIC BEGIN
+if (0)
```
JoshEngebretson: Good catch on the install stuff

Source/ThirdParty/easy_profiler/easy_profiler_core/CMakeLists.txt
```diff
@@ -207,5 +209,14 @@ install(
     EXPORT "${targets_export_name}"
     DESTINATION "${config_install_dir}"
 )
+endif ()
 
-target_compile_definitions(easy_profiler PUBLIC )
+if (MSVC)
+    add_library(easy_profiler_md ${EASY_OPTION_LIB_TYPE} ${SOURCES} resources.rc)
+    foreach(prop COMPILE_DEFINITIONS COMPILE_OPTIONS CXX_STANDARD CXX_STANDARD_REQUIRED INCLUDE_DIRECTORIES INTERFACE_COMPILE_DEFINITIONS INTERFACE_COMPILE_OPTIONS INTERFACE_INCLUDE_DIRECTORIES INTERFACE_LINK_LIBRARIES)
+        get_target_property(val easy_profiler ${prop})
```
JoshEngebretson: Fancy footwork, the cmake is strong with this one! 🐫

Source/ThirdParty/easy_profiler/profiler_gui/CMakeLists.txt
```diff
@@ -40,7 +40,15 @@ if (Qt5Widgets_FOUND)
         resources.qrc
         resources.rc
     )
-    target_link_libraries(profiler_gui Qt5::Widgets easy_profiler)
+    # ATOMIC BEGIN
+    target_link_libraries(profiler_gui Qt5::Widgets)
+    if (TARGET easy_profiler_md)
+        target_compile_options(profiler_gui PRIVATE /MD)
```
JoshEngebretson: Very nice and clean 👍

Source/ThirdParty/easy_profiler/CMakeLists.txt
rokups: I do not think it is possible to kick certain targets into "Release" build when everything else is building as "Debug". But maybe we could just remove `-DDEBUG=1` (which is implicit) and use `/MD` and have release build within debug build. That that might work, i will try.

rokups:
Again sorry for taking forever. Real life and summer caught up to me.

Sometimes i can be idiot of epic proportions. I completely forgot the fact that Qt ships both release and debug builds of their libraries. Forcing a release (`/MD`) build in otherwise debug build of engine was causing issues and completely unnecessary. Once i realized this it was easy to have it build debug version of profiler that links to dynamic msvc runtime just like rest of Qt. It all works perfectly. I also rebased branch on latest master.

To deploy profiler dlls on windows we are supposed to run this: `C:\Qt\5.9\msvc2017_64\bin\windeployqt.exe --no-angle --no-opengl-sw --no-translations C:\Atomic-easy_profiler\AtomicGameEngine\Artifacts\Build\Profiler\Release\Profiler.exe`. It dumps minimal required Qt libraries into directory where `Profiler.exe` is located.

@JoshEngebretson please give it a spin on MacOS and let me know if something is still missing or if is it complete now.

#1566 makes `Add manual optimized bindings for Profiler::BeginBlock()` goal obsolete so i am crossing it out.

Also please point me to some example how to add extra code for js. I would like to try adding something like [AtomicNET/Core/Profiler.cs
](https://github.com/rokups/AtomicGameEngine/blob/88fd006085fdafa4a750d81d7fd0d89c75834964/Script/AtomicNET/AtomicNET/Core/Profiler.cs) to javascript.

If profiling on MacOS works as expected and when i add this js bit i think this PR will be complete.

JoshEngebretson:
@rokups Absolutely zero worries man, this is a mega-feature! 😎 

I am really swamped here today, will test at first opportunity.  That's good news on the Windows side of things, thanks for R&D'ing it

Also, I haven't forgotten #1566. I'll look into whipping up a quick hand binding example for the profiler if that gets hairy (or for the meantime).

HeadClot:
This is insane. Just saw the Repo for easy profiler. Super happy to see this in AGE :)

Type1J:
@rokups This is great! Good job!

rokups:
Thanks for applauds guys :) 👏 

@JoshEngebretson optimized c# bindings are working. One more piece of the puzzle is that generated c# projects do not have `ATOMIC_PROFILING` defined. It should be defined at least in debug builds and it is ok if it is defined even if engine is built without `ATOMIC_PROFILING`. I can not find where i should change that. Halp 🆘

Script/AtomicNET/AtomicNET/Core/Profiler.cs
```diff
@@ -12,12 +13,16 @@ public partial class Profiler : AObject
         {
 #if ATOMIC_PROFILING
             var profiler = AtomicNET.GetSubsystem<Profiler>();
-            profiler?.BeginBlock(name, file, line, color, (byte)status);
+            if (profiler != null)
```
JoshEngebretson: We don't need AtomicNET here, AObject has GetSubsystem on it, the static AtomicNET flavor is only for getting a subsystem outside of a AObject derived instance.  Also, it might be good to locally cache so doesn't look up per block.

Source/AtomicNET/NETNative/NETCInterop.cpp
```diff
@@ -750,7 +751,15 @@ namespace Atomic
         }
 #endif
 
+#if ATOMIC_PROFILING
+        ATOMIC_EXPORT_API void csi_Atomic_Profiler_BeginBlock(Profiler* profiler, const char* name, const char* file, int line, unsigned int argb, unsigned char status)
+        {
```
JoshEngebretson: Let's make sure the csi_Atomic_Profiler_BeginBlock is available in the shared library, whether profiling is enabled or not... will avoid missing symbols. Can ifdef the profiler usage, also could add a one time log warning (controlled by a local static bool), which warns if called without profiling support in binary

JoshEngebretson:
The C# solution generator has support for adding constants, though not for debug vs release I don't think right now, example:

https://github.com/AtomicGameEngine/AtomicGameEngine/blob/a3cfc78d6a8ca0305c307842b1dd2f30f5877965/Script/AtomicNET/AtomicNETProject.json#L47

Would be a pretty easy add for supporting debug/release specific flags: 

https://github.com/AtomicGameEngine/AtomicGameEngine/blob/a3cfc78d6a8ca0305c307842b1dd2f30f5877965/Source/ToolCore/NETTools/NETProjectGen.cpp#L473

Script/AtomicNET/AtomicNET/Core/Profiler.cs
rokups: > We don't need AtomicNET here, AObject has GetSubsystem on it

👍 

> locally cache

You mean `Profiler.Block()` should not be static so that caller gets `Profiler` and reuses that reference?

Source/AtomicNET/NETNative/NETCInterop.cpp
rokups: Will do

Script/AtomicNET/AtomicNET/Core/Profiler.cs
JoshEngebretson: I mean so don't need to call GetSubsystem per block, which is not a terribly expensive call, though still, if that gets ugly that's fine, can cross bridge if it is an issue 🐫

JoshEngebretson: BTW, will this be the first C# code profiled with easy_profiler @cas4ey, awesome work again @rokups 🥇

rokups: This caching of subsystem pointers seem to comes up rather often. Maybe it would be better idea for next PR to cache engine subsystems as weak pointers in `Context`?

Script/AtomicNET/AtomicNET/Core/Profiler.cs
```diff
@@ -12,7 +12,7 @@ public partial class Profiler : AObject
                                  [CallerLineNumber] int line = 0)
         {
 #if ATOMIC_PROFILING
-            var profiler = GetSubsystem<Profiler>();
+            var profiler = AtomicNET.Context.GetProfiler();
```
JoshEngebretson: When it is right, it is right, very 🐫 !

brucelee

rokups:
@JoshEngebretson i cleaned up commit history, now it is in a state to be merged. Updated easy_profiler to latest code on their `develop` branch too and all custom modifications are marked with atomic comments. The last bit missing is bit for js profiling.

Source/Atomic/Resource/BackgroundLoader.cpp
```diff
@@ -264,20 +264,12 @@ void BackgroundLoader::FinishBackgroundLoading(BackgroundLoadItem& item)
     // If BeginLoad() phase was successful, call EndLoad() and get the final success/failure result
     if (success)
     {
-#ifdef ATOMIC_PROFILING
+#if ATOMIC_PROFILING
```
JoshEngebretson: On these profiling changes, I don't think we need to pepper ATOMIC modification blocks everywhere, we've diverged... next update, I guess see what run into and maybe adjust, though think modification blocks would just add noise in these cases

Source/Atomic/Resource/BackgroundLoader.cpp
```diff
@@ -264,20 +264,12 @@ void BackgroundLoader::FinishBackgroundLoading(BackgroundLoadItem& item)
...
+        ATOMIC_PROFILE_SCOPED(profileBlockName.CString(), PROFILER_COLOR_RESOURCES);
 #endif
         ATOMIC_LOGDEBUG("Finishing background loaded resource " + resource->GetName());
         success = resource->EndLoad();
-
-#ifdef ATOMIC_PROFILING
-        if (profiler)
-            profiler->EndBlock();
-#endif
     }
```
JoshEngebretson: Another sign of code improvement, not needing to explicitly ```EndBlock```, winning! 🏅

Source/CMakeLists.txt
```diff
@@ -8,6 +8,7 @@ option(ATOMIC_EDITOR "Build Editor" ON)
 option(ATOMIC_DATABASE_SQLITE "Enable SQLite database subsystem" OFF)
 option(ATOMIC_DATABASE_ODBC "Enable ODBC database subsystem" OFF)
 option(ATOMIC_IK "Enable inverse kinematics subsystem" OFF)
+option(ATOMIC_PROFILING "Enable profiler" ${ATOMIC_RELEASE_OFF})
```
JoshEngebretson: ```${ATOMIC_RELEASE_OFF}``` is very cool!

Script/AtomicNET/AtomicNETProject.json
```diff
@@ -23,6 +23,7 @@
             "name": "AtomicNET",
             "atomicNET" : true,
             "outputType" : "Library",
+            "defineConstants" : ["ATOMIC_PROFILING"],
```
rokups: People would be profiling a release build anyhow, this flag missing on release builds would certainly cause some confusion.

Source/Atomic/CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ elseif (ATOMIC_DATABASE_ODBC)
 endif ()
 
 if (WIN32)
+    option (ATOMIC_D3D11 "Use DirectX 11" ON)
```
rokups: Not sure how that happened, will fix ;)

Source/Atomic/Core/Object.cpp
```diff
@@ -302,9 +305,35 @@ void Object::SendEvent(StringHash eventType)
 
     SendEvent(eventType, noEventData);
 }
-
+// ATOMIC BEGIN
 void Object::SendEvent(StringHash eventType, VariantMap& eventData)
 {
+#if ATOMIC_PROFILING
+    bool eventProfilingEnabled = false;
+    if (Profiler* profiler = GetSubsystem<Profiler>())
```
rokups: Thanks to magic template specializations it gets profiler object from `context_` anyway, just the same as if new getter was used :p Yes, we get to use new getters for free without changing any existing code. Much 🐫

Source/Atomic/Math/StringHash.cpp
```diff
@@ -43,20 +43,24 @@ const StringHash StringHash::ZERO;
 StringHash::StringHash(const char* str) :
     value_(Calculate(str))
 {
+#if ATOMIC_PROFILING
+    RegisterSignificantString(str, *this);
+#endif
```
rokups: We need this for being able to recognize event names. Events as you know are sent using `StringHash`, but profiler needs actual strings for data to be useful. See [here](https://github.com/rokups/AtomicGameEngine/blob/4b4cc5027f1fc15d8bb9ad4f488c91f687540d10/Source/Atomic/Core/Object.cpp#L326-L329).

Source/Atomic/Resource/BackgroundLoader.cpp
rokups: This was reason why i did not add these comments elsewhere. We know that profiler subsystem was replaced therefore profiler changes should be ignored when merging from upstream. I am adding them to requested spots anyhow ;) Maybe we can find a way for got to help us do merges and make these comments thing of the past. I plan to look into it next ;)

Script/AtomicNET/AtomicNETProject.json
JoshEngebretson: I mean the equivalent of ```${ATOMIC_RELEASE_OFF}```, for release builds the ATOMIC_PROFILING define should default off in C#, no need to worry about it now.

rokups:
@JoshEngebretson requested changes landed. Once you review i will squash `Changes requested in code review` into `easy_profiler integration` in order to keep best appearances 👍

Source/Atomic/Math/StringHash.cpp
JoshEngebretson: Ok, this is definitely an area to keep an eye on, and maybe for future improvement.  It registers every time a string is converted to a StringHash, ouch... I know this is the only way right now, hopefully this doesn't come up on the top of the profiler ;)

It it does, maybe we can introduce a separate ```ATOMIC_PROFILING_EVENTS``` for enabling event profiling...

JoshEngebretson:
Great, cleared for landing, I'll just do some final binary testing once you have the commits arranged, thanks!

JoshEngebretson:
This would also be a good PR to add yourself to [AUTHORS.md](https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/AUTHORS.md) 🌟

rokups:
Already there, Rokas Kupstys ;)

JoshEngebretson:
Haha, awesome, you're a Super Contributor :)  If you want to add yourself to the Atomic Game Engine credits block, I think you should.. and technically, part of the contributing agreement: [CONTRIBUTING.md](https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/CONTRIBUTING.md)

camels4

JoshEngebretson:

Source/ThirdParty/easy_profiler/profiler_gui/CMakeLists.txt
```diff
@@ -0,0 +1,67 @@
+#set(CMAKE_PREFIX_PATH f:/qt/5.5/5.6/msvc2013_64/lib/cmake)
+
+set(CMAKE_AUTOMOC ON)
+set(CMAKE_AUTORCC ON)
+set(CMAKE_INCLUDE_CURRENT_DIR ON)
+
+find_package(Qt5Widgets)
+
+if (NOT Qt5Widgets_FOUND)
```
JoshEngebretson: This is supposed to be ```if (Qt5Widgets_FOUND)```?  I think a crumb may have been introduced when testing the warning message.

JoshEngebretson:
@rokups I made some tweaks and added JS/TS Profiler.BeginBlock support, feel free to look over and cherry pick into PR: https://github.com/AtomicGameEngine/AtomicGameEngine/commit/6a1c747d804a59e6854b49e2eb2ca8067a7ae99d

I also had some problems on CI Mac, so:

JoshEngebretson:
@rokups You absolutely shredded on the new Profiler subsystem 🎸 , very impressive work!  It has also been a lot of fun working on this feature with you, LEVEL UP! 🌟 

A hearty *THANKS* from :camel: everywhere, and special thanks to @cas4ey for such an excellent profiling library and client!!!

LANDING! ✈️

Type1J:
This is pretty awesome @rokups!

cas4ey:
Whoa! Great news! :+1: 
@rokups Thanks for your great work on the integration and for your valuable suggestions for easy_profiler :1st_place_medal: 
@JoshEngebretson It's a pleasure to create useful things which could help people :smiley:

PR #1538, opened 2017-05-16
[C#] Enum improvements and access to CSComponentAssembly type information at runtime

JoshEngebretson:
This PR improved enumeration serialization for script fields to be by enum string instead of numeric value.  It also adds a Cache representation of CSComponentAssembly type information which is available at runtime.  If generating the project Cache, and AtomicNETService is not available to process the binary solution, the cached information in the assemblies ```.asset``` file is used.


mattbenic:
Awesome :)

PR #1542, opened 2017-05-19
Add UISelectList accessors to get list strings

JimMarlowe:
This PR makes the UISelectList a little more useful. There is a lot of data hiding going on between the UISelectList and the UISelectItemSource, as far as I can tell you can only program a UISelectList with a UISelectItemSource. Once the UISelectItemSource is created and then leave the scope, you cant get the UISelectItemSource back out of the UISelectList. But, there is an internal TBSelectItemSource that the UISelectList uses to do the drawing and stuff, and I can find that, and return the strings that you can _clearly_ see... 
The added methods are :
``const String& UISelectList::GetItemString(int index);``
and
``const String& UISelectList::GetSelectedItemString();``
and they pass thru the script bindings without issue. Example code snippet :
```
if ( ev.target.id == "myselectlist") {
   var selectlist = <Atomic.UISelectList>this.mahWindow.getWidget(ev.target.id);
   var selected = selectlist.getValue();
   Atomic.print ( "Atomic.UISelectList selected item = " + selected );
   Atomic.print ( "Atomic.UISelectList selected string = " + selectlist.getSelectedItemString());
   Atomic.print ( "or Atomic.UISelectList selected string = " + selectlist.getItemString(selected));
   }
```

### DIFF
--- a/Source/Atomic/UI/UISelectList.h
+++ b/Source/Atomic/UI/UISelectList.h
@@ -83,6 +83,8 @@ protected:
 
 private:
 
+    String transferStr;
+
 };
 
 
--- a/Source/Atomic/UI/UISelectList.h
+++ b/Source/Atomic/UI/UISelectList.h
@@ -71,9 +71,9 @@ public:
     void SetUIListView(bool value);
 
     /// Returns the string of item at the requested index
-    const String& GetItemString(int index);
+    String GetItemString(int index);
     /// Returns the string of the selected item
-    const String& GetSelectedItemString();
+    String GetSelectedItemString();
 
 protected:
 
@@ -82,8 +82,6 @@ protected:
     virtual bool OnEvent(const tb::TBWidgetEvent &ev);
 
 private:
-
-    String transferStr;
 
 };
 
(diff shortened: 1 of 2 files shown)


JimMarlowe:
hmmm now that im using it for real, its being problematic. wait.

Source/Atomic/UI/UISelectList.h
```diff
@@ -70,6 +70,11 @@ class UISelectList : public UIWidget
 
     void SetUIListView(bool value);
 
+    /// Returns the string of item at the requested index
+    const String& GetItemString(int index);
+    /// Returns the string of the selected item
```
JoshEngebretson: I think we can change these signatures to return a String by value, as the common case is creating a new String, copying that to a member variable, and returning it to satisfy returning a reference... We might as well just return a new String by returning the const char* directly which will be automatically be converted.  This avoids the introduction of a new member variable and the copy,  when there isn't an item string, we can still return String::EMPTY

JoshEngebretson:
Great add, thanks, LANDING! ✈️

PR #1549, opened 2017-06-08
Minor fixes for TMX files

shaddockh:
This PR includes a couple quick fixes
- Latest version of Tiled has a version id of `1.0.0` instead of `1.0` so this recognizes both
- Tiled supports the ability to use individual images in a tileset instead of a spritesheet.  The version of the TMXFile in Urho and Atomic doesn't support this, but also doesn't indicate why loading a tmx with these fails.  This checks for that situation and writes out the appropriate error.

Note: I'll add a new issue to see if it will be possible for the TmxFile loader to support these types of tilesets in the future.

### DIFF
--- a/Source/Atomic/Atomic2D/TmxFile2D.cpp
+++ b/Source/Atomic/Atomic2D/TmxFile2D.cpp
@@ -388,7 +388,10 @@ bool TmxFile2D::BeginLoad(Deserializer& source)
                     return false;
 
                 // ATOMIC BEGIN
-                if (tsxXMLFile->GetRoot("tileset").GetChild("tile"))
+
+                // Look for an image indicating that this is a spritesheet with multiple tiles instead
+                // of a series of individual images which are not supported
+                if (!tsxXMLFile->GetRoot("tileset").GetChild("image"))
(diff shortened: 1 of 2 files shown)


Source/Atomic/Atomic2D/TmxFile2D.cpp
```diff
@@ -387,8 +387,19 @@ bool TmxFile2D::BeginLoad(Deserializer& source)
...
                     return false;
 
-                tsxXMLFiles_[source] = tsxXMLFile;
+                // ATOMIC BEGIN
 
+                // Look for an image indicating that this is a spritesheet with multiple tiles instead
+                // of a series of individual images which are not supported
+                if (!tsxXMLFile->GetRoot("tileset").GetChild("image"))
+                {
+                    ATOMIC_LOGERROR("Load TSX File failed: " + source + ". tsx files with individual images are not supported.");
```
JoshEngebretson: Excellent, these messages will help save some head scratching

JoshEngebretson:
Great, thanks, LANDING! ✈️

PR #1551, opened 2017-06-08
UIPromptWindow and UIFinderWindow Widgets 

JimMarlowe:
This PR does a couple of things... 
   The UISelectList is given access to the items in the list, and has added calls to be able to create, remove and populate the list items within the class (used in the UIFinderWindow).  The UIFieldEdit can now bring up the onscreen keyboard on mobile platforms to edit strings.
   There are two new widgets, UIPromptWindow and UIFinderWindow, both can be used in C++, C# and JS/TS.  UIPromptWindow is an Atomic widget that can get a string from the user, the result is sent to an event handler. UIFinderWindow is an Atomic widget that can get a file or path from the user, the result is sent to an event handler.  The Finder has the ability to create a directory at the current path, and it uses a UIPromptWindow to get the name.  It can also bookmark directories that the user finds interesting, again using a UIPromptWindow to change the name if the preselected name is not wanted. These bookmarks are saved on device (in the AtomicGameEngine preferences directory) and restored when the Finder is used again.

The UIFinderWindow needs testing on Windows because of the slashes.


Source/Atomic/UI/UIEditField.cpp
```diff
@@ -213,6 +215,17 @@ void UIEditField::OnFocusChanged(bool focused)
             if (!w->GetMultiline())
                 styleEdit->selection.SelectAll();
             firstFocusFlag_ = true;
+
+#if defined(ANDROID) || defined( __ANDROID__) || defined(IOS)
+
```
JoshEngebretson: For reference, we  have platform defines which help keep things consistent: https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/Atomic/CMakeLists.txt#L106

```ATOMIC_PLATFORM_ANDROID``` and ```ATOMIC_PLATFORM_IOS```, there is also ```ATOMIC_PLATFORM_LINUX```, ```ATOMIC_PLATFORM_WINDOWS```, etc

File: Source/Atomic/UI/UIEditField.cpp [hunk 2]
```diff
@@ -247,6 +260,22 @@ bool UIEditField::OnEvent(const tb::TBWidgetEvent &ev)
                 styleEdit->selection.SelectAll();
             }
         }
+
+#if defined(ANDROID) || defined( __ANDROID__) || defined(IOS)
+
+        // triple click to get the onscreen keyboard, in case it is auto-focused
+        else if ( ev.count == 3 ) 
+        {
```
JoshEngebretson: I am still not sure on the triple click, I will deploy to device and check this out :)

Source/Atomic/UI/UIFinderWindow.cpp
```diff
@@ -0,0 +1,618 @@
...
+using namespace tb;
+
+namespace Atomic
+{
+
+/// finder window
+UIFinderWindow::UIFinderWindow(Context* context, UIWidget* target, const String& id, bool createWidget) : 
+    UIWindow(context, false),
+    finderMode(0),
+    currentPath(),
```
JoshEngebretson: What's the story with explicitly calling the default constructors here?

Source/Atomic/UI/UIFinderWindow.cpp
```diff
@@ -0,0 +1,618 @@
...
+    {
+        GoFolderUp();
+        return true;
+    }
+
+    if ( ev.type == EVENT_TYPE_POINTER_UP && ev.target && ((uint32)ev.target->GetID()) == 3 ) // add bookmark request
+    {
+        // this check is necessary because you can kill the bookmark window with the "X" and
+        // this kills the UI but not the newBookmarkPtr to it, and it doesnt get cleaned up properly.
+        if ( newBookmarkPtr.NotNull() )
```
JoshEngebretson: Good commenting 👍

Source/Atomic/UI/UIFinderWindow.cpp
```diff
@@ -0,0 +1,618 @@
...
+        String prospect = "";
+        char delim = '/';
+        Vector <String> tokens = currentPath.Split(delim, false);
+        prospect = tokens[ tokens.Size()-1 ]; // get the last folder name as preset
+    
+        newBookmarkPtr->Show("Create New Bookmark", "Enter a name for the new bookmark", prospect );
+        return true;
+    }
+
+    if ( ev.type == EVENT_TYPE_POINTER_UP && ev.target && ((uint32)ev.target->GetID()) == 4 ) // add folder request
```
JoshEngebretson: Hm, magic number 4, can this be a define so more clear?

Source/Atomic/UI/UIFinderWindow.cpp
```diff
@@ -0,0 +1,618 @@
...
+            SetCurrentPath ( bookmarkPaths[selected]);
+            UpdateUiPath();
+            UpdateUiList();
+            UpdateUiResult();
+            bklist->SetValue(-1); // clear the select visuals
+        }
+        return true;
+    }
+    
+    if ( ev.type == EVENT_TYPE_CUSTOM && ev.target && ((uint32)ev.target->GetID()) == 5 ) // bookmarks TB context menu result
```
JoshEngebretson: Magic 5, these would be good to have as defines or constants 🎱

Source/Atomic/UI/UIFinderWindow.cpp
```diff
@@ -0,0 +1,618 @@
...
+        if (eventData[WidgetEvent::P_HANDLED].GetBool())
+           return true;
+    }
+
+    return UIWindow::OnEvent(ev);
+}
+
+void UIFinderWindow::HandleCreateBookmark(StringHash eventType, VariantMap& eventData)
+{
+    String Title = eventData["Title"].GetString();
```
JoshEngebretson: Just a FYI, these can be const String& to avoid a copy from eventData variant map

Source/Atomic/UI/UIFinderWindow.cpp
```diff
@@ -0,0 +1,618 @@
...
+
+// create the list of bookmarks ... can be different per platform
+void UIFinderWindow::CreateBookmarks()
+{
+    FileSystem* filesystem = GetSubsystem<FileSystem>();
+    String basepath = filesystem->GetUserDocumentsDir();
+    UISelectList *bklist = static_cast<UISelectList *>(GetBookmarksWidget());
+    
+#if defined(ATOMIC_PLATFORM_LINUX) 
+    if ( filesystem->DirExists ( basepath )) CreateBookmark ( "Home", basepath );
```
JoshEngebretson: Hey! ATOMIC_PLATFORM_ usage and nice touch adding defaults per platform! 👍

Source/Atomic/UI/UIFinderWindow.cpp
```diff
@@ -0,0 +1,618 @@
...
+    CreateBookmark ( "-", basepath );  // create separator!
+    LoadBookmarks();
+    bklist->SetValue(-1); // fix the selection and scrolling
+
+}
+
+// go up tree 1 folder
+void UIFinderWindow::GoFolderUp()
+{
+    String prospect = "";
```
JoshEngebretson: Would ```GetParentPath``` work here instead?
 ```https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Source/Atomic/IO/FileSystem.h#L155

Source/Atomic/UI/UIFinderWindow.cpp
```diff
@@ -0,0 +1,618 @@
...
+    String bkpath = "";
+    
+#if defined(ANDROID) || defined( __ANDROID__) || defined(IOS)
+    bkpath = filesystem->GetUserDocumentsDir(); // somewhere writable on mobile
+#else
+    bkpath = filesystem->GetAppPreferencesDir("AtomicGameEngine", "Bookmarks"); // desktop systems
+#endif
+    
+    bkpath += "/Bookmarks.txt";
+    if ( filesystem->FileExists ( bkpath ) )
```
JoshEngebretson: Really neat on persistent bookmarks, we do have really nice JSON serialization which is a bit more robust than generic .txt and less prone to parse errors with newlines and stuff.

Source/Atomic/UI/UIFinderWindow.h
```diff
@@ -0,0 +1,86 @@
...
+    UIWidget* GetFilesWidget();
+   
+private:
+
+    void PresetCurrentPath( const String& preset );  /// location where the finder starts
+    void SetCurrentPath( const String& string );  /// set the current path value
+    void ComposePath (const String& string );  /// using the list, jam things together, we'll either get another path or a file.
+    void CreateBookmarks(); /// create the list of bookmarks, per platform
+    void GoFolderUp(); /// go up one folder
+    void UpdateUiPath ();  /// move current path to widget
```
JoshEngebretson: Are the names here ```Ui``` instead of ```UI``` due to script generation, if so we should fix the binding gen for these instead of locking an inconsistent API string case

Source/Atomic/UI/UIFinderWindow.h
```diff
@@ -0,0 +1,86 @@
...
+    void UpdateUiResult (); /// move result path to widget
+    void UpdateUiList(); /// move folder contents into list
+    void CreateFolder( const String& string ); /// utility to add a folder in current path
+    void CreateBookmark ( const String& bkname, const String&  bkpath ); /// utility to add a bookmark from the current path
+    void DeleteBookmark ( int bkindex ); /// removes a bookmark based upon its index in the bookmark list
+    void LoadBookmarks();  /// load user created bookmarks from file
+    void SaveBookmarks();  /// save user created bookmarks to file
+
+    int finderMode;  /// finder mode, 0 = find file, 1 = find folder
```
JoshEngebretson: Member variable convention is to postpend with an ```_```, so ```int finderMode_```, etc

Source/Atomic/UI/UIFinderWindow.h
```diff
@@ -0,0 +1,86 @@
...
+    String currentPath; /// current path for the finder
+    String resultPath;  /// result file for finder
+    Vector <String> bookmarks;  /// array of names of bookmarks
+    Vector <String> bookmarkPaths; /// array of bookmark paths
+    WeakPtr<UIPromptWindow> newBookmarkPtr;  /// pointer for bookmark prompt window
```
JoshEngebretson: Glad ```WeakPtr``` helped out!

Source/Atomic/UI/UISelectList.h
```diff
@@ -74,6 +74,14 @@ class UISelectList : public UIWidget
     String GetItemString(int index);
     /// Returns the string of the selected item
     String GetSelectedItemString();
+    /// Add a new item at the given index
+    bool AddItem(int index, const String& str, const String& id = String::EMPTY );
```
JoshEngebretson: Thanks for the method comments! 👍

Source/ThirdParty/TurboBadger/tb_atomic_widgets.cpp
```diff
@@ -283,4 +285,355 @@ void TBBarGraph::SetAxis(AXIS axis)
...
+    "		TBTextField: id: 1, gravity: left right\n"
+    "			font: size: 14dp\n"
+    "		TBEditField: id: 2, multiline: 0, styling: 0, adapt-to-content: 0, gravity: left right\n"
+     "			font: size: 14dp\n"
+   "	TBSeparator: gravity: left right\n"
+    "	TBLayout: distribution: gravity\n"
+    "		TBButton: text: \"  OK  \", id: \"TBPromptWindow.ok\"\n"
+    "			font: size: 14dp\n"
+    "		TBButton: text: \"Cancel\", id: \"TBPromptWindow.cancel\"\n"
+    "			font: size: 14dp\n";
```
JoshEngebretson: Get your hot baked TB layouts here :)

Source/Atomic/UI/UIEditField.cpp
JimMarlowe: I will use those instead, I found the current set of defines by grepping in the Sources/Atomic directory.

Source/Atomic/UI/UIEditField.cpp
JimMarlowe: Or as I like to call it,  the "angry click"... I have no idea if it is a standard, but I suspect that no one has been entering data on android+atomic+turbobadger, and this makes it at least possible and not too annoying.  There are many more issues with TB on android to clean up.

Source/Atomic/UI/UIFinderWindow.cpp
JimMarlowe: It is an effort to reduce unintended behavior. In a former life, this initialization was required to prevent bugs from uninitialized values, Ive heard on windows, variables are automatically set to 0, but this is not the case on linux and probably osx, android.

JoshEngebretson: The default constructors will automatically be called, it isn't in the C/C++ spec to initialize variables on any compiler.

Source/Atomic/UI/UIFinderWindow.cpp
JimMarlowe: the TB Window has multiple problems, this closing by corner "X" without allowing the application to know being one of them. Others are, there is no concept of topmost window, so you can easily bury windows, and the title area has no provisions to set the size, so on mobile it can be really small.

Source/Atomic/UI/UIFinderWindow.cpp
JimMarlowe: I was using TB as an example, admittedly a poor example. I will make some defines, sadly I cant embed the defines in the tb ui definition string.
And dont get me started on the wondrous TB id system, too late.  A great system, use a name (hashed to a number) or a number, does not have to be unique, and all global. Great for toy programs, sux when you scale up.

Source/Atomic/UI/UIFinderWindow.cpp
JimMarlowe: Yes, that is true, but theses are my "documentation" so you know how to get the values out of the event map :smile_cat:

Source/Atomic/UI/UIFinderWindow.cpp
JimMarlowe: I will try that, and be less exposed in manual path operations.

Source/Atomic/UI/UIFinderWindow.h
JimMarlowe: No, they were born that way. These methods are protected, and not exposed to scripting.

Source/Atomic/UI/UIFinderWindow.h
JimMarlowe: I will fix.

Source/Atomic/UI/UIFinderWindow.h
JimMarlowe: Yes, ty.

Source/Atomic/UI/UISelectList.h
JimMarlowe: SelectList has (had) some real usability issues, as, sadly, do most TB widgets.

Source/ThirdParty/TurboBadger/tb_atomic_widgets.cpp
JimMarlowe: Faced with the options of loading the .ui.txt file or baking it in, baking seemed to be the better choice, less things to go wrong (I had the TBIDs and events to do that).

Source/Atomic/UI/UIFinderWindow.cpp
JimMarlowe: My choices were text, xml, json. In c++ coding the json stuff is just as complex as the xml, so I opted for lower complexity.

JimMarlowe:
After finding GetParentPath(), and applying it, it has a side effect to not stop at "/" and will go to "" as the root directory, so no. It is also assuming "/" as the path separator, so Im no worse off.

and then ...


```
    //
    //C++ snippet for UIPromptWindow
    // note you really need a c++ class to embed this in, and uiwidgetParent needs to exist.
    UIPromptWindow  *promptPtr = new UIPromptWindow(context_, uiwidgetParent, "myprompt", true);
    SubscribeToEvent(promptPtr, E_UIPROMPTCOMPLETE, ATOMIC_HANDLER(myClass, HandlePrompt));
    promptPtr->Show("WindowTitle", "Message in window", "preset value" );
  
    // class event handler for results
    void myClass::HandlePrompt(StringHash eventType, VariantMap& eventData)
    {
        String Title = eventData["Title"].GetString();  // these are the 3 fields in the event
        String Reason = eventData["Reason"].GetString();
        String Selected = eventData["Selected"].GetString();
        ATOMIC_LOGINFO ( "Title = " + Title );
        ATOMIC_LOGINFO ( "Reason = " + Reason );
        ATOMIC_LOGINFO ( "Selected = " + Selected );
        if( Reason == "OK" ) // reasons are strings "OK" and "CANCEL", depending on which was pressed
        {
            // do OK action 
            ATOMIC_LOGINFO ( "the string I got was " + Selected );
        }
    }

    //
    // C# snippet for UIPromptWindow
    uiview = new UIView();  // if you already have one, you can use it.
    UIPromptWindow prompt = new UIPromptWindow(uiview, "myprompt", true );
    SubscribeToEvent<UIPromptCompleteEvent>(prompt, PromptUpdate); 
    prompt.Show( "WindowTitle", "Message in window", "preset value", 0, 0, 0);

    // event handler for results
    void PromptUpdate(UIPromptCompleteEvent ev)
    {
            Log.Info ("Title = " + ev.Title );
            Log.Info ("Reason = " + ev.Reason );
            Log.Info ("Selected = " + ev.Selected );
        if( ev.Reason == "OK" )
        {
             // do OK action 
                Log.Info ("the string I got was " + ev.Selected );
        }
    }

    //
    // Javascript snippet for UIPromptWindow
    // you need a uiview in your program to use this
    var prompt = new Atomic.UIPromptWindow(uiview, "myprompt");
    prompt.show(  "WindowTitle", "Message in window", "preset value", 0, 0, 0);
    prompt.subscribeToEvent( prompt, "UIPromptComplete", function (ev) {
        Atomic.print ( "Title = " + ev.title);
        Atomic.print ( "Reason = " + ev.reason );
        Atomic.print ( "Selected = " + ev.selected );
        if( ev.reason == "OK" )
        {
            // do OK action 
            Atomic.print ( "the string I got was " + ev.selected );
        }
    });


    //
    // C++ snippet for UIFinderWindow
    UIFinderWindow  *finderPtr = new UIPromptWindow(context_, uiwidgetParent, "myprompt", true);
    SubscribeToEvent( finderPtr, E_UIFINDERCOMPLETE, ATOMIC_HANDLER(myClass, HandleFinder));
    
    finderPtr->FindFile("Find a File", "<preset path>" );  // find a filename
*OR*  
    finderPtr->FindPath("Find a Folder", "<preset path>" );  // find a folder
  
    // class event handler for results
    void myClass::HandleFinder(StringHash eventType, VariantMap& eventData)
    {
        String Title = eventData["Title"].GetString();  // these are the 3 fields in the event
        String Reason = eventData["Reason"].GetString();
        String Selected = eventData["Selected"].GetString();
        ATOMIC_LOGINFO ( "Title = " + Title );
        ATOMIC_LOGINFO ( "Reason = " + Reason );
        ATOMIC_LOGINFO ( "Selected = " + Selected );
        if( Reason == "OK" ) // reasons are strings "OK" and "CANCEL", depending on which was pressed
        {
            // do OK action 
            ATOMIC_LOGINFO ( "the file (or folder) I got was " + Selected );
        }
    }

    //
    // C# UIFinderWindow snippet  
    UIFinderWindow finder = new UIFinderWindow(uiview, "myfinder", true );
    SubscribeToEvent<UIFinderCompleteEvent>(finder, FinderUpdate); 
        
    finder.FindFile ("Find a File", "<preset path>", 0, 0, 0); // find a filename
*OR*
    finder.FindPath ("Find a Folder", "<preset path>", 0, 0, 0); // find a folder

    void FinderUpdate(UIFinderCompleteEvent ev)
    {
        Log.Info ("Title = " + ev.Title);
        Log.Info ("Reason = " + ev.Reason);
        Log.Info ("Selected = " + ev.Selected);
        if( ev.Reason == "OK" ) // reasons are strings "OK" and "CANCEL", depending on which was pressed
        {
            // do OK action 
                 Log.Info ("the file (or folder) I got was " + ev.Selected);
        }
    }
 
    //
    // Javascript UIFinderWindow snippet  
    var finder = new Atomic.UIFinderWindow(uiview, "myfinder");
        
    finder.findFile("Find a File", "<preset path>", 0, 0, 0);
*OR*
    finder.findPath("Find a Folder", "<preset path>", 0, 0, 0);
        
    finder.subscribeToEvent(finder, "UIFinderComplete", function (ev) {
        Atomic.print ( "Title = " + ev.title);
        Atomic.print ( "Reason = " + ev.reason );
        Atomic.print ( "Selected= " + ev.selected );
        if( ev.eason == "OK" ) // reasons are strings "OK" and "CANCEL", depending on which was pressed
        {
            // do OK action 
            Atomic.print ( "the file (or folder) I got was " + ev.selected );
        }
    });

```

JoshEngebretson:
Nice usage snippets!  One note on subscribing to events in JS/TS, you don't need to use raw strings like "UIFinderComplete", there should be a ```Atomic.UIFinderCompleteEventType``` you can use instead.

Does the behavior of ```GetParentPath``` need addressing, or is it doing the "right thing"?

On paths, the engine normalizes path internal storage to ```/``` , so all path operations should be working with that assumption, ```GetNativePath``` and ```GetWideNativePath``` are pretty much there for Windows, and only when interfacing with Windows API's that choke on ```/```

I spun up a branch for some quick tweaks and will land from there, thanks!

JoshEngebretson:
Thanks again, really good stuff, I think you are now the resident TurboBadger expert btw ⛷ 

I opened #1557 with a couple tweaks, will land when green! 💚

JimMarlowe:
The GetParentPath() is doing the correct thing. The Finder is providing "/" as the top to give the user a warm-fuzzy about where they are, though admittedly only Linux peeps would be alarmed if they saw nothing as the path.  I tried both types of JS events, both worked, but I was out of snippet space :earth_americas: 
An I'm no TB expert, I'm just annoyed with it enough to kick it, so I can get my job done :baby_chick:

PR #1554, opened 2017-06-10
Building engine as shared lib

rokups:
This PR adds `ATOMIC_DYNAMIC_RUNTIME` which enables building engine as shared library. Shared library builds also link to msvc runtime dynamically. By default flag is not enabled and static build is produced just like now.

Also added few warning cleanups and moved some build flags from global scope into Atomic project so that they are pulled in by downstream projects when linking to Atomic. These flags are `-std=gnu++11` and msvc runtime flags.

Tested and works on Linux and Windows. I didnt touch mac part at all (except for `-std=gnu++11`) since i can not test it nor i know what should be done. Tips would be appreciated. I myself am looking at moving compiler and linker flags from `AtomicMac.cmake` to Atomic project build script so they can be automatically inherited by downstream projects as well, but that is when we know that current changes didnt break anything-mac.


JoshEngebretson:
Excellent, I'll put this in the queue for today and have a look on the mac side too :) 🐫🌴☀️

Source/Atomic/CMakeLists.txt
```diff
@@ -226,3 +235,18 @@ target_include_directories (Atomic PUBLIC
 )
 
 include(AtomicDoc)
+
+if ($ENV{ATOMIC_BUILD_DIST})
+    target_compile_definitions(Atomic PUBLIC -DATOMIC_BUILD_DIST=1)
+endif ()
+
+if (MSVC)
+    target_compile_options(Atomic PUBLIC "$<$<CONFIG:Debug>:${ATOMIC_MSVC_RUNTIME}d>")
```
JoshEngebretson: I was combing through and looking to see we didn't lose linking to debug runtime.  Smiled here: ```${ATOMIC_MSVC_RUNTIME}d```, great attention to detail! 🎸

JoshEngebretson:
**AWESOME** Work!  It is really cool to be able to build an Atomic shared library, I went ahead and got the C++ FeatureExample linking with a shared library.  In order to get a shared Atomic.dll consumed in AtomicNET, the Editor, etc will require some additional work, though in those cases probably just a static link anyway... I also tested the STATIC build on macOS and Windows to ensure it was operational.

I'll be landing once green over in: https://github.com/AtomicGameEngine/AtomicGameEngine/pull/1559 💚

PR #1561, opened 2017-06-11
fix bug setting TB DefaultFont size

JimMarlowe:
This PR fixes a bug that has been there for awhile. There was a hardcoded 12 where the size argument should have been. Where this becomes important, is when you change the operating DPI (I lie to the TB system, since it lies right back) this bug hardcodes 12 (pixels) into all the default TB widgets as their size, and changing your dpi either way, results in peculiar (in a bad way) UI.


JoshEngebretson:
... aaaaand another great catch ⚾️ , thanks! LANDING! ✈️

PR #1567, opened 2017-06-14
imgui-based SystemUI

rokups:
Ready for testing. Example is at: https://github.com/rokups/AtomicExamples/tree/feature/imgui


JoshEngebretson:
This is really awesome :)  I think removing the old SystemUI system in favor of imgui is a good move, as we should have widget parity.  You also did the work of replacing DebugHud and Console, kudos 👍 

One note just from a quick glance, adding imgui's STB libs makes at least 3 copies of some of these headers.  What do you think about using https://github.com/AtomicGameEngine/AtomicGameEngine/tree/master/Source/ThirdParty/STB with imgui, adding any that are missing?

Looking forward to checking out this branch! 🚀 🐫 🚀

JoshEngebretson:

JoshEngebretson:
Thanks! This is a huge level up for tooling! I gave the PR a really close look and have to say: **FANTASTIC FREAKING WORK!** 👍 👍 👍 

The STB updates alone are awesome sauce, thanks for updating those to new versions and consolidating!

I'll land from here which just has a couple minor tweaks ✈️ : https://github.com/AtomicGameEngine/AtomicGameEngine/pull/1568

Issue #1197, opened 2016-11-24
[2D] Rain catch of small 2D issues


JoshEngebretson:
- [ ] Particle emitters have no way to stop them from emitting, disabling removes all particles for emitter from scene
- [ ] 2D physics collision events have a b2contact cast as void* in them and are not accessible in script

Alan-FGR:
I think we can go with a simplified way of exposing only the contact amount and positions of a given fixture, most of the time you never really need to access other contact data and most engines afaik don't expose the information, although in the most rarest case one may need to access it because box2d copies some data like friction to the contact objects, so if you want to modify that while the contacts are persisting you can't do that in the fixture. In only extremely rare cases you'll have to modify the friction of a contact for example.

Alan-FGR:
The default values for rigid bodies and fixtures produce strange unexpected results (fixed rotation objects), I think the defaults should produce an 'usable' physical object.

Alan-FGR:
Is there a way to raycast the scene?

Alan-FGR:
I still have to confirm, but changing joints parameters after the simulations stars doesn't have the expected behavior, and on a side note, apparently setting center of mass of a rigidbody isn't working too

Alan-FGR:
Do you guys plan to have a 2D version of the Ray type? The 2D raycast is more like a 'linecast' in which it takes a start and end position, and not origin, direction and magnitude like the usual representation of a ray. I have a usable code here but it's consistent with the 3D raycasting and inconsistent with the native API (which is consistent with the Box2D API):

Lazily adapted from the 3D physics raycasting code:

C++ interop, e.g.: NETCInterop.cpp (not the automatically generated)
```c++
ATOMIC_EXPORT_API void csi_Atomic_PhysicsWorld2D_RaycastSingle(PhysicsWorld2D* world2D, Vector2* origin, Vector2* direction, float maxDistance, unsigned collisionMask, PhysicsRaycastResult2D* result)
{
    if (!world2D || !result)
        return;
    world2D->RaycastSingle(*result, *origin, *origin+(direction->Normalized()*maxDistance), collisionMask);
}
```

C# interop
```c#
[StructLayout(LayoutKind.Sequential)]
public struct PhysicsRaycastResult2D
{
    public Vector2 Position;
    public Vector2 Normal;
    public float Distance;

    IntPtr bodyPtr;
[...]

PR #1571, opened 2017-06-18
added 2d raycasting

Alan-FGR:


Alan-FGR:
Good Jesus, I don't know how this ended up here but apparently it worked... I was trying to use git request-pull but apparently that doesn't use the same system GitHub uses, it was very confusing :P
https://github.com/AtomicGameEngine/AtomicGameEngine/issues/1197

Alan-FGR:
I need to find a decent git GUI :P

Script/AtomicNET/AtomicNET/Atomic2D/PhysicsWorld2D.cs
```diff
@@ -0,0 +1,27 @@
...
+    [StructLayout(LayoutKind.Sequential)]
+    public struct PhysicsRaycastResult2D
+    {
+        public Vector2 Position;
+        public Vector2 Normal;
+        public float Distance;
+
+        IntPtr bodyPtr;
+        public RigidBody2D Body => bodyPtr == IntPtr.Zero ? null : NativeCore.WrapNative<RigidBody2D>(bodyPtr);
+    }
```
JoshEngebretson: C# is so fancy, this raw access to NativeCore is something to look at some day soon.  It would be good to have a cleanup pass at Atomic NET, the code literally flew together in some places ;)

Source/AtomicNET/NETNative/NETCInterop.cpp
```diff
@@ -27,6 +27,7 @@
 
 #include <Atomic/Input/Controls.h>
 
+#include <Atomic/Atomic2D/PhysicsWorld2D.h>
```
JoshEngebretson: Thanks for being consistent with header includes

Script/AtomicNET/AtomicNET/Atomic2D/PhysicsWorld2D.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+using System.Runtime.InteropServices;
+
+namespace AtomicEngine
+{
+    [StructLayout(LayoutKind.Sequential)]
+    public struct PhysicsRaycastResult2D
+    {
```
JoshEngebretson: Good stuff, and consistent with 2D/3D physics C++ subsystems, which counts for something.  If we're incongruent with Box2D in places, that is ok, we don't expose the API, at least not in scripting.  It is fully available in C++ for customization, etc.  I think your Box2D experience is valuable here, I haven't done a ton with it outside of examples. :+1:

Alan-FGR: Yes, I agree, but as I said in the issue, I really think we should have consistency between the 2D and the 3D physics API, so it's easy for people to go back and forth. I really like how in Unity for example they are consistent, except for some peculiarities and of course one axis :P. That requires some bigger changes though and for the time being I guess this is enough for some use cases.

Issue #1544, opened 2017-05-24
nonet flag isn't applied automatically in all cases - PCL Reference Assemblies not installed


shearer12345:
The linux build instructions say that
> The --nonet flag will automatically be applied if Visual Studio or mono isn't installed on the system

My build (`sh Build_AtomicEditor.sh`) errors with
`/usr/lib/mono/4.5/Microsoft.Common.targets: error : PCL Reference Assemblies not installed.`

 as mono wasn't installed. 

It seems that it isn't detection that mono isn't installed, and is trying to build with C# features.

* Installing mono ([as detailed](https://github.com/AtomicGameEngine/AtomicGameEngine/wiki/Detailed-instructions-for-building-on-Linux)) fixed this
* Adding `--nonet` manually also works (after removing mono)

Happy to look more closely if you can direct to where the detection of mono's installation happens

## System details:
Ubuntu 16.10 64 bit

## Logs
[Atomic.log is attached (as .txt)](https://github.com/AtomicGameEngine/AtomicGameEngine/files/1026615/Atomic.log.txt)

tail of console follows
```
Errors:

/home/shearer/code/AtomicGameEngine/Artifacts/AtomicNET/Build/AtomicNET.sln (default targets) ->
(Build target) ->
/home/shearer/code/AtomicGameEngine/Artifacts/AtomicNET/Build/AtomicNET/AtomicNET.csproj (default targets) ->
/usr/lib/mono/4.5/Microsoft.Common.targets (GetReferenceAssemblyPaths target) ->

	/usr/lib/mono/4.5/Microsoft.Common.targets: error : PCL Reference Assemblies not installed.

	 57 Warning(s)
	 1 Error(s)

Time Elapsed 00:00:01.9028370

Compilation Command: -c mono "/home/shearer/code/AtomicGameEngine/Build/Managed/nuget/nuget.exe" restore "/home/shearer/code/AtomicGameEngine/Artifacts/AtomicNET/Build/AtomicNET.sln" && "xbuild" "/home/shearer/code/AtomicGameEngine/Artifacts/AtomicNET/Build/AtomicNET.sln" /p:Platform="Any CPU" /p:Configuration=Release

NETBuild Error for solution: /home/shearer/code/AtomicGameEngine/Script/AtomicNET/AtomicNETProject.json
/home/shearer/code/AtomicGameEngine/Build/node_modules/jake/lib/api.js:340
        throw errObj;
        ^

[...]

PR #1575, opened 2017-06-20
Modern and legacy Linux mono detection for #1544

JimMarlowe:
This should be sufficient rule to detect the existence of mono on (nearly) any type of linux system.


JoshEngebretson:
Shazam ⚡️ ! Thanks for addressing this issue, closes #1544

PR #1577, opened 2017-06-25
Turbobadger Widget updates

JimMarlowe:
This PR adds UIRadioButton and UIScrollbar to the Atomic widget family  :family_woman_girl_boy: 
Added API for setting a step size to UIInlineSelect, instead of it being hardcoded to 1. It can be changed in the layout file too.
Added API for getting the UISlider min and max values.
Added function in layout file reader to be able to assign "squeezable" attribute on UIButton and UITextFields.


JoshEngebretson:
Thanks for the PR, will have a look-see soon :camel: 👍

JoshEngebretson:
Great work, I went ahead and rebased to master (landed updates to UIView, etc yesterday, so there were conflicts created).  I'll land here when green: https://github.com/AtomicGameEngine/AtomicGameEngine/pull/1584

Thanks! 🍰 🐫

PR #1580, opened 2017-06-26
Common subsystem pointers in Context

rokups:
This PR adds most commonly used subsystem pointers to Context so we can avoid HashMap lookup when we need to get a subsystem. Lookup is cheap, but it all adds up.

In order to get hold of `SystemUI` pointer on engine initialization i had to clean up that part. Now `Engine::Initialize()` creates `SystemUI` subsystem and registers it. `SystemUI` constructor creates and registers `DebugHud` and `Console` subsystems.

### DIFF
--- a/Source/Atomic/Core/Context.h
+++ b/Source/Atomic/Core/Context.h
@@ -228,26 +228,26 @@ public:
     void AddGlobalEventListener(GlobalEventListener* listener) { globalEventListeners_.Push(listener); }
     void RemoveGlobalEventListener(GlobalEventListener* listener) { globalEventListeners_.Erase(globalEventListeners_.Find(listener)); }
 
-    WeakPtr<Engine> engine_;
-    WeakPtr<Time> time_;
-    WeakPtr<WorkQueue> workQueue_;
-    WeakPtr<Profiler> profiler_;
-    WeakPtr<FileSystem> fileSystem_;
-    WeakPtr<Log> log_;
-    WeakPtr<ResourceCache> cache_;
-    WeakPtr<Localization> l18n_;
-    WeakPtr<Network> network_;
-    WeakPtr<Web> web_;
-    WeakPtr<Database> db_;
-    WeakPtr<Input> input_;
-    WeakPtr<Audio> audio_;
-    WeakPtr<UI> ui_;
-    WeakPtr<SystemUI> systemUi_;
-    WeakPtr<Graphics> graphics_;
-    WeakPtr<Renderer> renderer_;
-    WeakPtr<Console> console_;
-    WeakPtr<DebugHud> debugHud_;
-    WeakPtr<Metrics> metrics_;
+    Engine* GetEngine() const { return engine_; }
+    Time* GetTime() const { return time_; }
+    WorkQueue* GetWorkQueue() const { return workQueue_; }
+    Profiler* GetProfiler() const { return profiler_; }
+    FileSystem* GetFileSystem() const { return fileSystem_; }
+    Log* GetLog() const { return log_; }
+    ResourceCache* GetResourceCache() const { return cache_; }
+    Localization* GetLocalization() const { return l18n_; }
+    Network* GetNetwork() const { return network_; }
+    Web* GetWeb() const { return web_; }
+    Database* GetDatabase() const { return db_; }
+    Input* GetInput() const { return input_; }
+    Audio* GetAudio() const { return audio_; }
+    UI* GetUI() const { return ui_; }
+    SystemUI* GetSystemUI() const { return systemUi_; }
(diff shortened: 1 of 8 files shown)


JoshEngebretson:
Excellent, and will automatically be available to C#, JavaScript/TypeScript! 🎁 

I'll have a closer look later today, thanks! 👍

Source/Atomic/Core/Object.h
```diff
@@ -409,4 +432,30 @@ struct ATOMIC_API EventNameRegistrar
...
+template <> Audio* Object::GetSubsystem<Audio>() const;
+template <> UI* Object::GetSubsystem<UI>() const;
+template <> SystemUI* Object::GetSubsystem<SystemUI>() const;
+template <> Graphics* Object::GetSubsystem<Graphics>() const;
+template <> Renderer* Object::GetSubsystem<Renderer>() const;
+template <> Console* Object::GetSubsystem<Console>() const;
+template <> DebugHud* Object::GetSubsystem<DebugHud>() const;
+template <> Metrics* Object::GetSubsystem<Metrics>() const;
+// ATOMIC END
+
```
JoshEngebretson: Excellent, great call on using template specializations for core subsystems! 👍 

lawrenceflame

Source/Atomic/Core/Object.h
```diff
@@ -39,6 +39,29 @@ namespace Atomic
 class Context;
 class EventHandler;
 
+// ATOMIC BEGIN
+class Engine;
```
JoshEngebretson: I cringed a little at first introducing these forward decl in Object.h, though for the greater good, and we'll always know who the major players are... also, as purely passing pointers don't have to bring headers into Object.cpp, which means we're still pretty air tight :dash:

Very good...

Source/Atomic/Core/Context.h
```diff
@@ -228,6 +228,27 @@ class ATOMIC_API Context : public RefCounted
     void AddGlobalEventListener(GlobalEventListener* listener) { globalEventListeners_.Push(listener); }
     void RemoveGlobalEventListener(GlobalEventListener* listener) { globalEventListeners_.Erase(globalEventListeners_.Find(listener)); }
 
+    WeakPtr<Engine> engine_;
```
JoshEngebretson: What do you think about moving these raw members to```private``` and then adding method accessors?  I think would be more consistent, and will make all of these subsystems gettable off Context in C#/JS/TS and also protect them from being set...

rokups: I wouldnt necessarily say getters are more consistent. Bunch of classes expose their members, even `Vector3` and friends, there certainly are some others too. I am torn between kosher (getters) and practical (pointers).

Using pointers would immediately mean less typing, 4 characters per call and we can immediately take advantage of intellisense where with getters in order to get to relevant intellisense suggestion we would need to type at least 4 characters ("GetE", for engine for example).

Using getters would protect clueless people from shooting themselves in the foot, but maybe this software is not for clueless people and they should be in charge of protecting themselves :)

Maybe compromise is possible? C# could have properties and it would be very idiomatic. C++ could stay with pointers and js could... no idea what people do on js.

JoshEngebretson: C#, JS/TS properties are automatically generated for ```Getxxx``` methods so whether the WeakPtr's are private or public the methods should exist, Vector3 is a primitive and you definitely want direct access to ```x_, y_, z_``

My spidey sense is telling me private WeakPtr and public accessors, and not both... 🕷️

rokups: Alrighty, on it

rokups:
@JoshEngebretson need your input. I added seemingly ordinary `Engine* GetEngine() { return engine_; }` but generated bindings for duktape are unhappy:

```cpp
static int jsb_class_Context_GetEngine(duk_context* ctx)
{
duk_push_this(ctx);
Context* native = js_to_class_instance<Context>(ctx, -1, 0);
const Object* object = native->GetEngine();
js_push_class_object_instance(ctx, object);
return 1;
}```

> error: cannot convert ‘Atomic::Engine*’ to ‘const Atomic::Object*’ in initialization
 const Object* object = native->GetEngine();

Adding includes for subsystems to `Context.h` solves the issue. Any objection for that? Is there maybe a better way?

JoshEngebretson:
Add the include to:

```Script/Packages/Atomic/Core.json```

Here's an example:

https://github.com/AtomicGameEngine/AtomicGameEngine/blob/master/Script/Packages/Atomic/Physics.json#L5

JoshEngebretson:
Awesome, level up! 🌟 

... and thanks for the tweaks, LANDING! ✈️

PR #1592, opened 2017-07-06
Automatically deploy Qt dlls to profiler executable directory.

rokups:
Change takes care of putting minimal required Qt dlls to the directory where `Profiler.exe` is built. This is only done for windows. Linux does not need this at all and i am not sure how it should work on MacOS so that one is left out. So as a result after building Atomic now you can go to `Artifacts/Build/Profiler/(Release|Debug)` and run `Profiler.exe`  without needing any extra steps.

### DIFF
--- a/Build/CMake/Modules/AtomicCommon.cmake
+++ b/Build/CMake/Modules/AtomicCommon.cmake
@@ -229,3 +229,28 @@ macro(msvc_set_runtime runtime_flag)
         set(${var} "${${var}} ${runtime_flag}d")
     endforeach()
 endmacro()
+
+# Macro deploys Qt dlls to folder where target executable is located.
+# Macro arguments:
+#  target - name of target which links to Qt.
+macro(deploy_qt_dlls target)
+    if (WIN32)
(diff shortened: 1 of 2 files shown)


Build/CMake/Modules/AtomicCommon.cmake
```diff
@@ -229,3 +229,28 @@ macro(msvc_set_runtime runtime_flag)
...
+                add_custom_command(
+                    TARGET ${target}
+                    POST_BUILD
+                    COMMAND "${windeployqt}" --no-quick-import --no-translations --no-system-d3d-compiler --no-compiler-runtime --no-webkit2 --no-angle --no-opengl-sw $<TARGET_FILE_DIR:${target}>
+                    VERBATIM
+                )
+            endif ()
+        endif ()
+    endif ()
+endmacro()
```
JoshEngebretson: Great using standard ```windeployqt``` here as it does a good job packaging the Qt runtime.  We may want to follow up on the macOS side too using it, in the future ✨

PR #1594, opened 2017-07-07
Linux Control key detection wrong

### LINKED ISSUE #1423: [Linux] Super/Control key detection wrong
There are a couple of instances of detecting when the super key on OSX and the Alt Key on Windows is pressed. When compiling for linux, the OSX methodology is being used, it needs to follow the windows way, since linux uses a standard keyboard.

JimMarlowe:
This PR fixes a couple of instances of detecting when the Control vs Command button is pressed on linux. This resolves #1423


JoshEngebretson:
Good catch, thanks! LANDING! ✈️ 

Closes #1423

PR #1597, opened 2017-07-09
Updates to TypeScript and Monaco Editor

shaddockh:
This PR updates TypeScript to version 2.4.1 (see https://blogs.msdn.microsoft.com/typescript/2017/06/27/announcing-typescript-2-4/ )

It also updates the Monaco editor to 0.9 (see https://github.com/Microsoft/monaco-editor/blob/master/CHANGELOG.md#090-03072017 )

There were also a few changes to ensure that everything still works.  The main difference was that themes are handled differently under Monaco 0.9.  

Monaco 0.9 also adds a new minimap along the side.


JoshEngebretson:
Tested on Windows and all good, including compiling a TypeScript project, btw the minimap is really **awesome**!  Thanks for the update! 🌟 LANDING! ✈️

PS: One thing I want to look into is a submodule for the node_module dependencies, in order to keep the main repo clean.  Do you do anything special with the node_modules or are they pretty much local npm installs?

shaddockh:
Node modules is pretty static, so it could be a submodule.  It's only being used during build and the only time I've really updated it is when I'm bringing in a new version of TypeScript.  I think the TypeDoc process pulls down it's own node modules on the fly during build.

PR #1599, opened 2017-07-10
Changes conflicting enum names, and removes a badly thought out API call.

Type1J:
I've had this locally for a while. I'm trying to reconcile the upstream and my local version of Atomic. This fix may resolve some script naming conflicts, if anybody has them.

### DIFF
--- a/Source/Atomic/Web/WebRequest.cpp
+++ b/Source/Atomic/Web/WebRequest.cpp
@@ -79,8 +79,6 @@ struct WebRequestInternalState
     bool isAddedToMulti;
     /// Error string. Empty if no error.
     char error[CURL_ERROR_SIZE];
-    /// String cache of response
-    String response;
 
     WebRequestInternalState(WebRequest &es_) :
         es(es_)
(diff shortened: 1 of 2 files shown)


JoshEngebretson:
Thanks for the PR, looks great, nice clean ups on the enum :+1: I am curious if you have you used the download BufferQueue object in script yet?  If so, which language, and how was the experience?

We're green here, so LANDING! ✈️

Type1J:
I have in C++ and JavaScript. The BufferQueue exists to prevent possible reallocation when you can't afford it, and you probably already have an allocation that you just want to drop in at the end. It works especially well when you can just use that same allocation that you want to drop in at the end.

I've used something like BufferQueue for years in sound queues. Infact, Atomic's BufferQueue came from it's sound code. I forget where exactly, but somewhere down in the "here's the next sound buffer" code.

Type1J:
The downside is that it's a Deserializer subclass, but it doesn't seek. When any full allocation that it owns is read, it just DecRef()s it, so it will be deallocated if it's the only remaining owner. That's causing me a bit of trouble with something that I've been working on as a higher level interface to WebRequest, actually. I've modified the Resource class a bit to allow it a DOWNLOADING state, and it seems to work well, but some Resource loaders try to seek on the Deserializer that they're given, so that made me have to copy the BufferQueue to a VectorBuffer before using it as the Deserializer to hand off to BeginLoad(). I'm doing the copy in the loading thread, so it doesn't affect main thread performance, but I was wondering if I needed to refactor the Resource loaders that exist and put a "no seek" restriction on them. I think that's probably too limiting, so I'll stick with the quick copy in the loading thread for now. ResourceCache isn't being using currently. It seems that most of the ResourceCache API wants the resource to be immediately available, which is not practical for a Resource that's being downloaded from some remote system. I'm trying to integrate it anyway because it would fit in better, but I may have to keep them separate.  ... Sorry... rambling...

PR #1603, opened 2017-07-12
Fix #1602

### LINKED ISSUE #1602: [Windows OpenGL] Windows OpenGL regression (DX11 default)
Setting DX11 to default has broken OpenGL build switch - you get DX11 whether you like it or not.

rokups:
Fix #1602


JoshEngebretson:
Great, that's what I get for only testing D3D11 and D3D9, thanks for the fix (and improvements), landing! ✈️

rokups:
Noooo that still had one bug... ! 💥

JoshEngebretson:
We can fix it in post! ✨ 

fixpost

PR #1606, opened 2017-07-13
Turbobadger TabContainer Docking

JimMarlowe:
This PR adds undocking and redocking for the Turbobadger UITabContainer Widget. For undocking a tab, the page contents can be reparented into a Window (for real!). For redocking, the content is reparented again back into the UITabContainer. A new widget, the UIDockWindow, was created, that can redock with an embedded button in the titlebar.
Added NumChildren to TBWidget, a missing method.
Added PrettyPrint to TBWidget, this has been used to check is the un/redocking was doing what it was suppose to.


JoshEngebretson:
Wow, widget docking in 500 lines, impressive!    Can I test the docking behavior?  Maybe with https://github.com/JimMarlowe/AtomicExamples/tree/JM-PERIODIC/UIPeriodicTable?

I will savor this fine PR in the AM, thanks!

tyrion-wine-gif

JimMarlowe:
Yes, Ive updated the UIPeriodicTable program to demonstrate the docking stuff.

Source/ThirdParty/TurboBadger/tb_atomic_widgets.cpp
```diff
@@ -759,5 +760,144 @@ void TBPulldownMenu::OnInflate(const INFLATE_INFO &info)
...
+    {
+        if (ev.type == EVENT_TYPE_CLICK)
+            Redock();
+        return true;
+    }
+    return TBWidget::OnEvent(ev);
+}
+
+void TBDockWindow::OnResized(int old_w, int old_h)
+{
```
JoshEngebretson: I really like this method too, very solid.

Source/ThirdParty/TurboBadger/tb_widgets.cpp
```diff
@@ -773,6 +775,47 @@ TBWidget *TBWidget::GetWidgetAt(int x, int y, bool include_children) const
...
+    TBLinkListOf<TBWidget>::Iterator mx = GetIteratorForward();
+    while (TBWidget *mxw = mx.GetAndStep())
+    {
+        nn++;
+    }
+    return nn;
+}
+
+/** print out the widget tree */
+void TBWidget::PrintPretty( TBStr indent, bool last)
```
JoshEngebretson: Nice add! 👍

Source/Atomic/UI/UIDockWindow.h
```diff
@@ -0,0 +1,56 @@
...
+
+#include<TurboBadger/tb_atomic_widgets.h>
+
+#include "UIWidget.h"
+#include "UIWindow.h"
+
+namespace Atomic
+{
+
+class ATOMIC_API UIDockWindow : public UIWindow
```
JimMarlowe: I will add comment

JimMarlowe:
The ability to undock/redock is actually not limited to tabs, it can be anything, but the TabContainer is deterministic enough to demonstrate a completed system without having to change it in the future.  A general TBLayout decoder ring might take awhile...  Due to the ID system being somewhat relative, if you currently have code that looks up an undocked content ID, you might not find it when undocked.  The solution to that is, search from the uiview widget, and then it will find the ID in the dockwindow.  I have some follow on work, which will make this a general function so you don't have to remember what to do, and create some widget tree query functions to find all widgets of a class, an ID, and a GetText.  Then you can create a menu of DockWindows.

Script/Packages/Atomic/UI.json
```diff
@@ -23,6 +23,7 @@
 		],
 		"UIWidget": [
 		    "getWidget<T extends UIWidget>(id: string): T;",
+		    "findWidget<T extends UIWidget>(id: string): T;",
```
JoshEngebretson: Nice bonus add! 👍

Source/Atomic/UI/UIWidget.h
```diff
@@ -203,6 +204,17 @@ class ATOMIC_API UIWidget : public Object, public tb::TBWidgetDelegate
 
     void DeleteAllChildren();
 
+    /// searches for specified widget ID from the top of the widget tree, returns the 1st one found.
+    virtual UIWidget *FindWidget ( const String& searchid );
```
JoshEngebretson: Where has this been all my life? 💋

Source/Atomic/UI/UIWidget.h
```diff
@@ -203,6 +204,17 @@ class ATOMIC_API UIWidget : public Object, public tb::TBWidgetDelegate
...
+    /// return all of the widgets of the specified classname and id that is not 0 from the current widget
+    virtual void SearchWidgetClass ( const String& className, UISelectItemSource *results );  
+    ///  return all of the widgets of the specified id and id that is not 0 from the current widget
+    virtual void SearchWidgetId ( const String& searchid, UISelectItemSource *results );
+    /// return all of the widgets with the specified text and id that is not 0 from the current widget
+    virtual void SearchWidgetText ( const String& searchText, UISelectItemSource *results );
+    /// print out the widget tree to stdout from the current widget
```
JoshEngebretson: Hm, I wonder if we should be using UISelectItemSource here as a general container?  It would probably be better to use a ```Vector<SharedPtr<UIWidget>>```, this will automatically be available to C#.  Automated JavaScript vector support is really not far off since it was added to C#.  In the meantime, on the JS side, it would take adding something like we do for getting scene children by name: https://github.com/AtomicGameEngine/AtomicGameEngine/blob/ab0157e4900011a0f70216e16f5de48fe9c43abe/Source/AtomicJS/Javascript/JSScene.cpp#L141 

What do ya think?

JimMarlowe: My rational was I needed a dynamic container that already existed. I though (briefly) about using std::vector as the container, but didn't want to add that dependency to TB, I tried the TB LinkedList, but that was a disaster. So that left UISelectItemSource, which works ok.  Even though it will ignore anonymous id widgets in the scripting languages, you're going to have to _grok_ Turbobadger pretty good to know what you're missing, and you still have the option of putting an id on a widget if you want to talk to it.

JoshEngebretson: On the Atomic side of things, we should be returning a Vector of UIWidgets here.  It should be possible to fill one of these pretty easily from the tb side of the search.  

I'd rather not have UISelectItemSource being used as a general container, or add a UISelectItemSource dependency to UIWidget when it is ultimately a Vector of Widgets that we are interested in.  

I can look into this later in the week if needed.  Cool?

JimMarlowe: I'll look a little more at this, I think I can get away from using the UI/TBSelectItemSource, and then tackle the js array binding.

JoshEngebretson:
Awesome, thanks for the API improvements!  👍 

I added a quick fix to ensure return results were clean and will land in #1611 when green ✅

JimMarlowe:
Good deal. That JS/TS manual binding was an "adventure" :wine_glass:

PR #1613, opened 2017-07-26
Add UIPeriodicTable example

JimMarlowe:
This adds the UIPeriodicTable example to the AtomicEditor example browser. 
My Examples submodule isn't updating for the new UIPeriodicTable directory, is there something else I need to do to make that happen?


JoshEngebretson:
Yeah, the submodule needs to be updated to commit af394b23609d45071b490d8d85c8c5187ecd05db, you should be able to achieve this by checking out that commit in the Submodules/AtomicExamples and then committing that in the main repo.  Let me know if you're having any problems with it.

JoshEngebretson:
Awesome, thanks again :)

flash2

PR #1615, opened 2017-07-27
Allow Editor Plugins to define shortcuts and some linting

JimMarlowe:
This PR allows AtomicEditor Plugins to define Shortcut definitions and process the shortcut actions.  The Editor has an elaborate Shortcut system, but it doesn’t allow for shortcuts to be dynamically defined. There is also some mechanism that also defines unqualified F1-F12 as "special", which, sadly, won't allow these keys to be programmed from plugins, since they won't emit an event. Using qualified (shift, control, alt) emit events though. It will not check for duplicate shortcut definitions.

Using is page as an example https://github.com/AtomicGameEngine/AtomicGameEngine/wiki/Editor-Extensions

To define the shortcut : 

    this.serviceLocator.uiServices.createPluginMenuItemSource("My Menu", { "Open" : [`${this.name}.open.myaction`, "Ctrl+F5" ] });

To enable the shortcut action, add this code to the plugin projectLoaded() function :

```
        var myplug = this;
        this.subscribeToEvent( "UIShortcut", function (ev) {
            if ( ev.key == Atomic.KEY_F5  && ev.qualifiers == Atomic.QUAL_CTRL ) 
                myplug.menuItemClicked( "MyMenuPlugin.open.myaction" );
        });

```

There are also some misc TSLINT reject fixes that have crept in, while I was checking the file I changed.

### DIFF
--- a/Script/AtomicEditor/editor/Preferences.ts
+++ b/Script/AtomicEditor/editor/Preferences.ts
@@ -491,7 +491,7 @@ class PreferencesFormat {
 
         this.developmentUI = {
             projectFrameWidthScalar: 1
-        }
+        };
 
     }
 
@@ -562,7 +562,7 @@ class PreferencesFormat {
             prefs.developmentUI = this.developmentUI;
             updatedMissingDefaults = true;
         }
-        
+
         return updatedMissingDefaults;
     }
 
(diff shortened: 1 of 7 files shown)


JoshEngebretson:
Very nice, thanks!  LANDING ✈️

PR #1617, opened 2017-07-31
Add Atomic UIButtonGrid widget

JimMarlowe:
This PR adds an Atomic UIButtonGrid widget. This "widget" is special. It creates a regular grid of buttons, these buttons will remain in a grid organization, even if the container is resized. It will ellipse "..." the text in the buttons if there is not enough room to fully display the text.  This will make it easy to create game menus, and sudoku, tic-tac-toe boards :smile_cat:  There are two methodologies to enter the button text, either by specifying a row and column address, or the "floodfill" add text, which tiles in the text sequentially. The embedded UIButtons can be retrieved by both indexing methodologies, so you can assign event handlers to process the button clicks, or by IDs assigned at creation, using the spreadsheet like scheme, having A1 as the first id, and row incrementing letters, and columns incrementing numbers.
I had to do a little plumbing in Turbobadger to get access to the resize information, otherwise, I'd have to make a TB widget and wrap it.

This is what it can do for the feature examples ...
```
#if ATOMIC_DESKTOP || ATOMIC_MOBILE  
            var buttonGrid = new UIButtonGrid( 6, 4, 6 );
            buttonGrid.SetLayoutHeight( rootLayout.Rect.Height - 32 );
            buttonGrid.SetLayoutWidth( rootLayout.Rect.Width - 32);
            rootLayout.AddChild( buttonGrid );
            var sampleTypes = typeof(Sample).Assembly.GetTypes().Where(t => t.IsSubclassOf(typeof(Sample)) && t != typeof(Sample) ).ToArray();
            foreach (var sample in sampleTypes)
            {
                int got = buttonGrid.AddGridText ( sample.Name );
                var button = buttonGrid.AtGridWidget(got);
                button.SubscribeToEvent<WidgetEvent>( button,  e => 
                {
                    if (e.Type != UI_EVENT_TYPE.UI_EVENT_TYPE_CLICK)  return; // We're only interested in clicks
```
featuredcs

### DIFF
--- a/Source/Atomic/UI/UIButtonGrid.h
+++ b/Source/Atomic/UI/UIButtonGrid.h
@@ -36,23 +36,37 @@ class ATOMIC_API UIButtonGrid : public UIWidget
 
 public:
 
-    UIButtonGrid(Context* context, int numRows, int numCols, int margin=0, bool createWidget = true );
+    UIButtonGrid(Context* context, int numRows, int numColumns, int margin=0, bool createWidget = true );
     virtual ~UIButtonGrid();
-
-    virtual String GetGridId( int row, int col ); /// returns id text at r,c, ie, "A1" for 0,0
-    virtual void SetGridText ( int row, int col, String str );  /// set button text at r,c location
-    virtual String GetGridText( int row, int col ); /// returns button text at r,c
-    virtual UIWidget* GetGridWidget( int row, int col ); /// returns widget at r,c
-    virtual int AddGridText ( String str ); /// add strings starting at 0,0 and filling out cols, then next row... , returns count
-    virtual void ResetAddCount() { m_count = 0; } /// resets the counter for AddGridText()
-    virtual UIWidget* AtGridWidget( int count ); /// returns widget at count
-    virtual String AtGridText( int count ); /// returns text at count
-    virtual void DisableEmptyButtons(); /// convience function to disable buttons with no text, and enable those with text.
-    virtual int GetNumRows() const;  /// returns number of rows that were programmed
-    virtual int GetNumCols() const;  /// returns number of cols that were programmed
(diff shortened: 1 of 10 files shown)


Source/Atomic/UI/UIButtonGrid.h
```diff
@@ -0,0 +1,77 @@
...
+class ATOMIC_API UIButtonGrid : public UIWidget
+{
+    ATOMIC_OBJECT(UIButtonGrid, UIWidget)
+
+public:
+
+    UIButtonGrid(Context* context, int numRows, int numCols, int margin=0, bool createWidget = true );
+    virtual ~UIButtonGrid();
+
+    virtual String GetGridId( int row, int col ); /// returns id text at r,c, ie, "A1" for 0,0
```
JoshEngebretson: If comments are above the method, they will be included in documentation.  The comments should also be written how you would want them to appear in documentation.  Do these methods all need to be virtual?

Source/Atomic/UI/UIButtonGrid.h
```diff
@@ -0,0 +1,77 @@
...
+   virtual bool OnEvent(const tb::TBWidgetEvent &ev);
+   virtual void OnResized(int old_w, int old_h);
+
+private:
+
+    void GenerateGrid();
+    void ResizeGrid();
+    virtual void AddGridRow( int rownum );
+
+    int m_rows;
```
JoshEngebretson: Member naming convention would be ```rows_``` here, I'd also go with ```columns_``` as it is more clear and matches the full word usage or ```rows_```.  We live in the age of autocomplete :)

JimMarlowe:
Points taken, resolved.

JoshEngebretson:
Awesome, looks great, LANDING! ✈️

PR #1622, opened 2017-08-11
Move imgui ui assembly and rendering to END_RENDERING event. 

rokups:
Required for interoperability with Urho3D UI. It allows me to render Urho3D UI just before imgui in order to have imgui be rendered on top of everything. It does not affect anything related to TurboBadger, imgui still renders on top of it like before.


JoshEngebretson:
Ok, great, thanks for the PR! ✈️

PR #1623, opened 2017-08-12
Misc small improvements

rokups:
This is a small collection of recent fixes and few additions. Lets go through them:

CMake version requirement bumped to 3.3. Why you ask? Because noone noticed, but we already depend on that version since we started using `$<$<COMPILE_LANGUAGE:CXX>:-std=gnu++11>`. `COMPILE_LANGUAGE:CXX` thingy was introduced in 3.3. Older distros may get recent prebuilt CMake binaries from CMake website and they run. Tested with CMake 3.9 on ubuntu 14.04 on travis.ci.

Several build fixes for android NDK 13b. Namely must not link to `pthread` because it does not/no longer exists on android. Also wrapped `-std=gnu++11` into generator expression because otherwise compiler fails building c files due to unknown flag. Also excluded profiler tool from build on android in a way that does not cause any warnings.

`ImGui::Text(messageText_.CString());` -> `ImGui::TextUnformatted(messageText_.CString());` because again on android it fails to build due to printing user-supplied text as format string being insecure. Clang is really nice, isnt it? ;)

Added `IntVector2` constructor taking in float array. No explanation needed here.

Added `SetBlockEvents(bool)` and `GetBlockEvents()` to `Object`. This allows us to block certain objects from sending and receiving events. I found it very convenient for implementing game states, but i am sure there are other useful applications. Blocking events is easier than unsubscribing and resubscribing after all.

Added `SendEvent(StringHash, const VariantMap&)` overload which allows us to use initializer lists for sending events. For example `SendEvent("foo", { {Foo::P_BAR, "bar"}, {Foo::P_FOO, 1} })`. While using this method variant causes extra copy i think there should be a convenience method for cases where speed does not matter.

This is a mix of various things, but i think they should be acceptable. If not, and if you think i should split PR up or get rid of some changes please let me know.

### DIFF
--- a/Source/Atomic/CMakeLists.txt
+++ b/Source/Atomic/CMakeLists.txt
@@ -241,5 +241,5 @@ if (MSVC)
 endif ()
 
 if (UNIX OR MINGW)
-    target_compile_options(Atomic PUBLIC -std=gnu++11)
+    target_compile_options(Atomic PUBLIC $<$<COMPILE_LANGUAGE:CXX>:-std=gnu++11>)
 endif ()
--- a/Source/Atomic/Core/Object.cpp
+++ b/Source/Atomic/Core/Object.cpp
@@ -75,7 +75,10 @@ bool TypeInfo::IsTypeOf(const TypeInfo* typeInfo) const
 }
 
 Object::Object(Context* context) :
-    context_(context)
+    context_(context),
+    // ATOMIC BEGIN
+    blockEvents_(false)
+    // ATOMIC END
 {
     assert(context_);
 }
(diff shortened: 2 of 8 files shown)


rokups:
@JoshEngebretson builds are failing on CI. Maybe it is a time to discuss official bump of required cmake version?

JoshEngebretson:
I will be out of the office this week.  I'll be able to look into this when I return.  Let's keep the PR open until then, thanks!

rokups:
Crowd in the chat cant wait for master to come home ;)

JoshEngebretson:

JoshEngebretson:
@rokups Thanks for the PR, I updated the CI boxes and now green ✔️ 

LANDING!

rokups:
Awesome! One thing though - for testing i have reverted back to `cmake_minimum_required(VERSION 2.8.12.1)`, it should be changed to `cmake_minimum_required(VERSION 3.3)`

Issue #1620, opened 2017-08-08
[UI] Turbobadger wishlist (in no particular order)


darrylryan:
- UIScrollContainer doesn't seem to work properly from code (@JimMarlowe gave me an example in TB layout language, I tried to do the same in code, it broke. Possibly due to different defaults and differences in the layout enums?)

- Currently can't set a cursor image

- Window resizing currently only works in BOTH dimensions at once if enabled. Sometimes we want to be able to resize a window in X or Y only.

- Turning off the window header just leaves a big useless space at the top - sometimes we want windows without any header at all.

- Combined with the above, sometimes we want to have a drag handle on the side of windows (e.g. skill bars where you can stack them on top of eachother) and we don't want to have any padding at the top.

- It would be good to be able to set windows to snap to a grid, or to edges of the screen or other windows.

- A very useful custom widget would be a grid or table where you can define X rows and Y columns and then assign items to them by X/Y. This would be useful in tons of games for inventories and containers, item shops.

JimMarlowe:
To "condition" UILayouts to match the layout file implementations,  use this, right after constructing the UILayout: 
 ```mylayout.SetLayoutConfig("XPCPC");  // set defaults same as for layout file.```

PR #1625, opened 2017-08-14
Add titlebar axis control for UIWindow

JimMarlowe:
This Pr adds API control of the titlebar axis/location on the UIWindow. The default will be the AXIS_Y, which places the titlebar at the top of the window. The new option is AXIS_X, which moves the titlebar to the left side, the closer 'X' moves to the top of the titlebar, and what's left of the title string is at the bottom, though since Turbobadger can't display rotated text, it quickly clips the string. The resizer still occupies the lower right corner. The DockWindow was also updated with this capability.
This will satisfy a couple of items in #1620  [UI] Turbobadger wishlist (in no particular order)

window_xy

### DIFF
--- a/Source/Atomic/UI/UIDockWindow.cpp
+++ b/Source/Atomic/UI/UIDockWindow.cpp
@@ -117,6 +117,13 @@ bool UIDockWindow::OnEvent(const tb::TBWidgetEvent &ev)
     return UIWidget::OnEvent(ev);
 }
 
+void UIDockWindow::SetAxis(UI_AXIS axis)
+{
+    if (!widget_)
+        return;
+
+    widget_->SetAxis((AXIS) axis);
+}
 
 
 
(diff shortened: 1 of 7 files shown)


JimMarlowe:
Im adding a change to the UIScrollContainer, which makes AddChild() do the expected action, and the action taken by the layout file. This will fix #1626

JimMarlowe:
This code makes for easier programming of undocked windows. When a window is undocked,
the event flow from the contained widgets now go into the parent which has been changed to the dockwindow. Any past event subscriptions are still in effect, except they don't work as expected (at all). What this change does, is if the origin widget has been specified (used for integrated redocking) it will also reroute event emits back to this widget, and the original event handlers will again receive the events. magic.  If no origin widget is given, one will have to unsubscribe and resubscribe event handler for widgets that emit events.

PR #1628, opened 2017-08-18
Exposed SetColor for UIBargraph so that it can be used from script

darrylryan:


rokups:
Some tabs have sneaked in ;)

JoshEngebretson:
@darrylryan I concur, tab-elicious! 🍗 :)

JoshEngebretson:
Great, thanks for pruning the tabbage 👍

PR #1629, opened 2017-08-22
Merge with upstream

rokups:
Merge with upstream. Basically Urho3D 1.7 + few fixes.

Most notable since last merge:

* IK updates
* WebP support
* TMX file format updates
* Some Text3D updates (oversampling, etc)


rokups:
Need advice on these c# errors, what should be done with `IntVector2` here?

JoshEngebretson:
Thanks for another mighty fine PR!

I will have a look at the IntVector2 issue tomorrow and see if I can tap it into place. 🔨

Alan-FGR:
@rokups any ideas on what tiled version is recommended now? Docs still mention 0.92.

rokups:
@Alan-FGR i have no idea what `tiled` is.

Alan-FGR:
> @Alan-FGR i have no idea what tiled is.

And you call yourself a game developer? :trollface:
You proclaim that TMX support was updated, I thought you'd know what it's about :P

rokups:
> @rokups and you call yourself a game developer?

I dont. I am random noob from the depths of internetz :trollface: 

> You proclaim that TMX support was updated, I thought you'd know what it's about :P

Not like its my work ;)

JoshEngebretson:
@rokups I fixed the IntVector2 issue and will land in https://github.com/AtomicGameEngine/AtomicGameEngine/pull/1631, thanks!

BTW, are you still using the same git merge method?

rokups:
Yep, did this using git grafting again. Only TMX parser class was more of an issue. We should probably consider upstreaming our changes if possible.

Alan-FGR:
I've tried the lastest tiled stable release (1.0.2) and TMXs fail to load the images for some reason :(... the supported version definitely should be documented, so I don't know if the problem is somewhere else, or if it's the importer since the debug info is quite limited, so I don't even know if I should try downgrading tile until it works or something :(... maybe I should start testing a file from 0.92 to make sure the problem isn't elsewhere

Alan-FGR:
After further investigation, it has something to do with the filenames apparently...

Alan-FGR:
I'm still going to try some more complex maps, but saving the tmx file crashed the editor, can somebody else confirm?

Alan-FGR:
OK, TMX does need some work... objects layers zfight with tile layers and some object types won't show too :(

rokups:
@Alan-FGR could you test with urho3d? Maybe i messed up solving conflicts. That file was a bit problematic.

Alan-FGR:
@rokups I will... I will also test the TMX files in the examples to see if they didn't break. I'm starting to think though that the features having problem they never worked properly to begin with. Object layers are apparently just partially supported and upon a quick check at the source apparently the layers don't have anything to prevent zfighting with the tile layers. The editor crashes even with the simplest map though, when you save if in tiled and the editor is open it always crash.

PR #1630, opened 2017-08-22
Add Multicolumn List Widget

JimMarlowe:
UIMultiList
These are 2 classes that enable the UISelectList to have multiple columns.  There is API to set the column type, column width, data for the column and fixing the column height, if necessary. Once set into the UISelectList, it operates the same way, to get which row is selected.

There are a couple of column types that can be used
  TEXT = which is a TextField, with left justified text, the column data is the text that will be displayed.
  TEXTC = a TextField, with center justified text
  TEXTR = a TextField with right justified text
  IMAGE = an ImageWidget, the column data is the resource name (or pathname) to the file to display
  COLOR = a ColorWidget, the column data is the hex color string, like "#228800", to display.
  ICON = a SkinImage, the column data is the skin name for the icon to display.
  
If the column height is left as 0, the content will determine the height of the row. If you need it to be a fixed size, the column height field can be used.
The columm widths must be programmed, and are needed to keep alignment for the widget, the content will not exceed the programmed value.

Example Javascript code
```
    var mis = new Atomic.UIMultiItemSource();  // new itemsource

    var columnHeight = 0;   // 0 = let content determine column height
    var  = "item1";   // RefId string returned when list entry is selected
	var aumi1 = new Atomic.UIMultiItem( columnRefId, "ICON", "LogoAtomic", 32, columnHeight );
	aumi1.addColumn ( "TEXT", "a new kind of list...", 210 );
	aumi1.addColumn ( "TEXTC", "RW",  20 );
	aumi1.addColumn ( "COLOR", "#33FF33", 20 );
	aumi1.addColumn ( "TEXT", "moar", 66 );
	aumi1.addColumn ( "TEXT", "moar text is fine, in multiple columns.", 266 );
    mis.addItem( aumi1 );
    
    mylist.setSource(mis);

```
multilist

### DIFF
--- a/Script/Packages/Atomic/UI.json
+++ b/Script/Packages/Atomic/UI.json
@@ -11,7 +11,7 @@
 								"UIContainer", "UISection", "UIInlineSelect", "UITextureWidget", "UIColorWidget", "UIColorWheel",
 								"UIScrollContainer", "UISeparator", "UIDimmer", "UISelectDropdown", "UISlider", "UIBargraph",
 								"UIPromptWindow", "UIFinderWindow", "UIPulldownMenu", "UIRadioButton", "UIScrollBar", "UIDockWindow",
-								"UIButtonGrid"],
+								"UIButtonGrid", "UIMultiItem", "UIMultiItemSource" ],
 	"overloads" : {
 	},
 	"typescript_decl" : {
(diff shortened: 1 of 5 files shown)


JimMarlowe:
Going thru and testing the TS, C++, C# I found I had picked who I derived the "Multi" objects from poorly. Now they are derived from UISelectItem/Source and no longer give errors in use.

JoshEngebretson:
Excellent!  Thanks for the PR 👍 

Bonus points for moors filled with lavender :)
moor

LANDING! ✈️

PR #1637, opened 2017-08-25
Optional submodules

rokups:
By popular demand (from @Alan-FGR) this change implements pulling submodules only if features that require these submodules are enabled. So disabling `ATOMIC_CPLUSPLUS_EXAMPLES` will not pull examples and disabling `ATOMIC_WEBVIEW` will not pull CEF.


Alan-FGR:
@rokups while I solved my problem just by cloning recursive when GitHub got fast for a moment (it often caps speed @ ridiculous levels), I think this is extremely important since the only way to get a new version of Atomic these days is by compiling it yourself (bin distro's old as heck), and these submodules are extremely huge; some people just don't have gigabit internet, and I personally would honestly give up if I was on a slow connection and had to wait hours just to clone it in order to compile and give it a try. Even if you have gigabit internet GitHub caps speed way too often, we shouldn't expect our potential users to go through all that trouble. Ideally we should provide binaries though.

Build/CMake/Modules/AtomicGit.cmake
```diff
@@ -1,18 +1,26 @@
...
-            WORKING_DIRECTORY ${ATOMIC_SOURCE_DIR}
-            RESULT_VARIABLE ATOMIC_GIT_STATUS)
-        # check return status
-        if (NOT ATOMIC_GIT_STATUS EQUAL 0)
-            message(FATAL_ERROR "BUILD ERROR:\n\nError initializing submodules: git submodule update --init returned ${ATOMIC_GIT_STATUS}\n")
-        endif ()
-    else ()
-        # Git not found
-        message(FATAL_ERROR "BUILD ERROR:\n\Submodules not initialized, please run: git submodule update --init\n")
+macro (initialize_submodule target_dir result_variable)
```
JoshEngebretson: Nice refactor! 👍

JoshEngebretson:
Thanks for the PR, landing! ✈️

PR #1652, opened 2017-09-12
Add F# and other .NET languages

ShalokShalom:


mattbenic:
In theory this should be true, but without resources to explicitly test and deal with any possible properties, we'll stick to only specifically stating we have C# support.

ShalokShalom:
Yeah, just ignore dozens of languages, including one of the simplest functional programming languages, just because it could be that there are issues. As C# compiles down to CIL, it makes no difference. 

If it makes a difference, it could make the same difference in C# as well. 

I suggest to simply merge this patch and act if any feedback comes in.

ShalokShalom:
No, I prefer to use an Engine that will support it so as Vulkan: http://www.banshee3d.com/

Thanks

mattbenic:
Fair enough. Good luck with your projects.

ShalokShalom:
Thanks a lot.

Issue #357, opened 2015-10-21
Add right-click import on resources folders


LaraEngebretson:
To quickly import sounds, textures, scenes, etc.

Originally requested here: #159

PR #1656, opened 2017-10-08
Add right-click import on resources folders #357

JimMarlowe:
This PR add the ability to import an asset file by searching for it with a file finder, instead of using OS drag and drop. 
There is a new entry in the context menu in either Project window when a folder is selected, this entry is called "Import Asset...".  Selecting "Import Asset..." brings up the file finder to search for an asset file that will be saved into the Project folder which was originally selected.  Due to limitations in the file finder, only one asset may be imported at at time.  There is a message in the editor's status area which reports the success or failure of the import operation.
This will be for all (AtomicEditor) platforms.


JimMarlowe:
I added a asset database scan to a successful file import, since that's what a drop of a asset does.

mattbenic:
Verified on Windows. Just waiting for a linux verify. In the meantime, please rebase this to latest master so it'll include the CI hooks and trigger a build.

JimMarlowe:
Rebased, as requested. I have tested it on Mint Linux, Which is Ubuntu 14.x based, with the now required CMake upgrade.

PR #1663, opened 2017-10-27
Update README.md

iamblocksberg:
add License add bottom


mattbenic:
Making the in-use license clear is a valid edit, but doesn't need 3 commits. Please squash them and resubmit.

PR #1691, opened 2018-03-11
Unhardcode icon file type in GNU/Linux launcher

palob:
Since you're going to install the icon to a location compliant with [freedesktop.org standards](https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html) you don't need to specify the icon file extension in the `.desktop` launcher. It will be found anyway.

This facilitates the use of non-PNG (e.g. SVG) app icon themes.
Cf. this [example `.desktop` launcher](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#example).


mattbenic:
Your PR failed travis build, please correct before we can accept your PR. Are you able to access those build logs from the link above? If not let me know and I'll drop them in a gist for you.

